PR #26, opened 2015-06-11
sponsor adapter added

mananwason:


mananwason:
@creativepsyco One class was missing from the last commit. I am not sure why. So, I am merging this one.

PR #47, opened 2015-06-30
set statusbar and navbar color on Lollipop and above

championswimmer:
###OLD
old   

###NEW
new

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.activities;
 
+import android.os.Build;
 import android.os.Bundle;
 import android.support.design.widget.NavigationView;
 import android.support.v4.app.FragmentManager;
@@ -40,6 +41,11 @@ public class MainActivity extends AppCompatActivity {
         NavigationView navigationView = (NavigationView) findViewById(R.id.nav_view);
         if (navigationView != null) {
             setupDrawerContent(navigationView);
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            getWindow().setStatusBarColor(getResources().getColor(R.color.color_primary));
+            getWindow().setNavigationBarColor(getResources().getColor(R.color.color_primary));
         }
 
 //        FragmentManager fragmentManager = getSupportFragmentManager();


championswimmer:
@creativepsyco @mananwason review and merge :)

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -42,6 +43,11 @@ protected void onCreate(Bundle savedInstanceState) {
             setupDrawerContent(navigationView);
         }
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            getWindow().setStatusBarColor(getResources().getColor(R.color.color_primary));
```
creativepsyco: I believe setting this in the theme palette as colorPrimary etc. Results in less code and more separation.

```xml
<style name="Theme.MyTheme" parent="Theme.AppCompat.Light">
    <!-- colorPrimary is used for the default action bar background -->
    <item name="colorPrimary">@color/my_awesome_color</item>

    <!-- colorPrimaryDark is used for the status bar -->
    <item name="colorPrimaryDark">@color/my_awesome_darker_color</item>

    <!-- colorAccent is used as the default value for colorControlActivated
         which is used to tint widgets -->
    <item name="colorAccent">@color/accent</item>

    <!-- You can also set colorControlNormal, colorControlActivated
         colorControlHighlight & colorSwitchThumbNormal. -->
</style>
```
Please refer to https://chris.banes.me/2014/10/17/appcompat-v21/

championswimmer: Yes it totally is right. I just used the color from where Manan has placed
it right now and using everywhere else. I am assuming he has a theming
scheme in mind.

creativepsyco: Would depend on the event I guess. Or manually customisable by changing a themes.xml for different events. Or we a script to write it altogether.

championswimmer: Can be done from the server.  Set a primary color field in the event
endpoint.
Fetch it, and setColor programmatically wherever required?

creativepsyco: That's a pretty broad issue in itself, don't think setting programmatically
here makes a difference because the rest of the palette is immutable once
declared in XML, there isn't a way to change the theme, the textView
cursors etc all derive from the palette colours defined in XML. Even if we
are able to do this, we still then need to manually change accent colours
for ALL the widgets because they will look out of place. I would recommend
placing the theme colours inside XML at this stage.

But feel free to work this out in a separate issue. Status bar colors are
only 1 spec of the material theme, there is widget tinting, highlightColors
and a whole bunch of other properties that need addressing. It would look
odd if the theme colors are red, the status bar is blue and the cursor or
the switch, or the checkbox is blue in color.

The XML already allows a declarative style semantic, so I would in my
opinion just stick with it.

On Wed, Jul 1, 2015 at 1:17 AM, Arnav Gupta <notifications@github.com>
wrote:

> In app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
> <https://github.com/fossasia/open-event-android/pull/47#discussion_r33598346>
> :
>
> > @@ -42,6 +43,11 @@ protected void onCreate(Bundle savedInstanceState) {
> >              setupDrawerContent(navigationView);
> >          }
> >
> > +        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
> > +            getWindow().setStatusBarColor(getResources().getColor(R.color.color_primary));
>
> Can be done from the server. Set a primary color field in the event
> endpoint. Fetch it, and setColor programmatically wherever required?
> … <#14e45797cda6dfd3_>
> On Jun 30, 2015 10:43 PM, "Mohit Kanwal" <notifications@github.com>
> wrote: In
> app/src/main/java/org/fossasia/openevent/activities/MainActivity.java <
> https://github.com/fossasia/open-event-android/pull/47#discussion_r33597794>
> : > @@ -42,6 +43,11 @@ protected void onCreate(Bundle savedInstanceState) {
> > setupDrawerContent(navigationView); > } > > + if (Build.VERSION.SDK_INT
> >= Build.VERSION_CODES.LOLLIPOP) { > +
> getWindow().setStatusBarColor(getResources().getColor(R.color.color_primary));
> Would depend on the event I guess. Or manually customisable by changing a
> themes.xml for different events. Or we a script to write it altogether. —
> Reply to this email directly or view it on GitHub <
> https://github.com/fossasia/open-event-android/pull/47/files#r33597794>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fossasia/open-event-android/pull/47/files#r33598346>.
>



-- 
Mohit Singh KANWAL (Mr)

[image: Facebook] <http://facebook.com/mohitkanwal>  [image: Twitter]
<http://twitter.com/mohitkanwal>

mananwason: Yeah I was thinking about this. We would not be able to change the colour of every component by obtaining colors from api. Maybe yeah the primary colors but I don't know. @championswimmer you said you have worked on something like this before. Can you guide here?

creativepsyco: I have seen reddit flow app do it. Essentially it follows this approach:

http://stackoverflow.com/questions/4663752/switching-application-wide-theme-programmatically

But a restart of any activity is required. Or put it in a BaseActivity
before calling setContentView(). But in any case you can still only declare
styles once in XML, so still needs to be part of the app package. So it is
dynamic in a sense, but only dynamic within the declared styles.


On Wed, Jul 1, 2015 at 1:48 AM, Manan Wason <notifications@github.com>
wrote:

> In app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
> <https://github.com/fossasia/open-event-android/pull/47#discussion_r33602029>
> :
>
> > @@ -42,6 +43,11 @@ protected void onCreate(Bundle savedInstanceState) {
> >              setupDrawerContent(navigationView);
> >          }
> >
> > +        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
> > +            getWindow().setStatusBarColor(getResources().getColor(R.color.color_primary));
>
> Yeah I was thinking about this. We would not be able to change the colour
> of every component by obtaining colors from api. Maybe yeah the primary
> colors but I don't know. @championswimmer
> <https://github.com/championswimmer> you said you have worked on
> something like this before. Can you guide here?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fossasia/open-event-android/pull/47/files#r33602029>.
>



-- 
Mohit Singh KANWAL (Mr)

[image: Facebook] <http://facebook.com/mohitkanwal>  [image: Twitter]
<http://twitter.com/mohitkanwal>

championswimmer: Well in the earlier effort I made (individually, before GSOC), we were
generating a styles.xml when the apk was being generated using the web
based android app generator as per the organizers chosen color scheme.

creativepsyco: Yep that I think is a good way way to do it. No point tinkering with
dynamic colours if the framework doesn't allow it transparently.

mananwason: Agreed. Seems like a better way.

PR #103, opened 2015-12-18
Added Traditional Chinese

seadog007:


mananwason:
Here too!! Squash commits.

Issue #127, opened 2015-12-28
Error Handling FIXED


shivenmian:
A system is required to handle errors / give prompts whenever an error occurs in the server / HTTP / elsewhere, instead of a mere snackbar.

- [x] Show network errors through dialog/snackbar
- [ ] Create a logger to assimilate log messages for the app

creativepsyco:
:+1:

creativepsyco:
Log also maybe.

shivenmian:
I have logs in the code. Please check my PR.

vjs3:
@shivenmian can't we use ACRA (www.acra.ch) for that purpose?or your error handling system is doing someting else as well?

shivenmian:
I don't think we should use any external bug reporting feature like this. Directly using libs is not something we have done in any part  of the code.

creativepsyco:
Not yet but later, for now it is okay.

shivenmian:
@creativepsyco Do you want me to create that logger as well?

creativepsyco:
The system should behave in such a way that once you enable it, there should be a log collection happening to a file. That file can then be exported or sent via email. then all the log.d or log.v messages will have to be passed to that file + printed in logcat.

Check out https://github.com/JakeWharton/timber for log aggregation.

shivenmian:
Okay @creativepsyco . I'll definitely work on this.

mananwason:
@creativepsyco I don't think there's any point in sending all the logs to the developer. We should send the mail if there's an error that requires the attention of the developer. So there can be a check in the download logs that if there is conversion error or HTTPError which could mean that the response format is bad or the server is down, we can write in a file and send it to the developer for immediate action.

shivenmian:
Exactly what I was thinking.

PR #128, opened 2015-12-28
Fixed Issue #127

shivenmian:
I've come up with a system to handle errors as in Issue #127. It will show dialogs and give the respective error that has occurred.

For HTTP errors, it gives a dialog with the Error code too. For others, it gives a dialog with the type of error and the cause.

When network is unavailable, it gives a dialog to turn on network in Settings.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -49,6 +49,7 @@
             android:enabled="true" />
 
         <receiver android:name=".Receivers.NotificationAlarmReceiver" />
+
     </application>
 
 </manifest>
--- a/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
+++ b/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
@@ -1,11 +1,14 @@
 package org.fossasia.openevent;
 
 import android.app.Application;
+import android.content.Context;
+import android.content.IntentFilter;
 import android.os.Handler;
 import android.os.Looper;
 
 import com.squareup.otto.Bus;
 
+import org.fossasia.openevent.Receivers.NetworkConnectivityChangeReceiver;
 import org.fossasia.openevent.dbutils.DbSingleton;
 import org.fossasia.openevent.modules.MapModuleFactory;
 
@@ -39,6 +42,7 @@ public class OpenEventApp extends Application {
         super.onCreate();
         DbSingleton.init(this);
         mapModuleFactory = new MapModuleFactory();
+        registerReceiver(new NetworkConnectivityChangeReceiver(), new IntentFilter(Context.CONNECTIVITY_SERVICE));
     }
 
     public MapModuleFactory getMapModuleFactory() {
--- a/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -1,8 +1,13 @@
 package org.fossasia.openevent.activities;
 
+import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.graphics.Color;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
 import android.os.Bundle;
+import android.provider.Settings;
 import android.support.design.widget.NavigationView;
 import android.support.design.widget.Snackbar;
 import android.support.v4.app.FragmentManager;
@@ -10,6 +15,7 @@ import android.support.v4.app.FragmentTransaction;
 import android.support.v4.view.GravityCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBarDrawerToggle;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
 import android.util.Log;
@@ -28,13 +34,15 @@ import org.fossasia.openevent.OpenEventApp;
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.dbutils.DataDownload;
 import org.fossasia.openevent.dbutils.DbSingleton;
-import org.fossasia.openevent.events.CantDownloadEvent;
+import org.fossasia.openevent.events.ConnectionCheckEvent;
 import org.fossasia.openevent.events.CounterEvent;
+import org.fossasia.openevent.events.DataDownloadEvent;
 import org.fossasia.openevent.events.EventDownloadEvent;
 import org.fossasia.openevent.events.MicrolocationDownloadEvent;
 import org.fossasia.openevent.events.NoInternetEvent;
 import org.fossasia.openevent.events.RefreshUiEvent;
 import org.fossasia.openevent.events.SessionDownloadEvent;
+import org.fossasia.openevent.events.ShowNetworkDialogEvent;
 import org.fossasia.openevent.events.SpeakerDownloadEvent;
 import org.fossasia.openevent.events.SponsorDownloadEvent;
 import org.fossasia.openevent.events.TracksDownloadEvent;
@@ -44,6 +52,7 @@ import org.fossasia.openevent.fragments.SpeakerFragment;
 import org.fossasia.openevent.fragments.SponsorsFragment;
 import org.fossasia.openevent.fragments.TracksFragment;
 
+import retrofit.RetrofitError;
 
 public class MainActivity extends AppCompatActivity {
 
@@ -57,6 +66,10 @@ public class MainActivity extends AppCompatActivity {
     private int eventsDone;
     private int currentMenuItemId;
     private final static String STATE_FRAGMENT = "stateFragment";
+    public String errorType;
+    public String errorDesc;
+    public static final String TYPE = "RetrofitError Type";
+    public static final String ERROR_CODE = "Error Code";
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -71,11 +84,8 @@ public class MainActivity extends AppCompatActivity {
         downloadProgress = (ProgressBar) findViewById(R.id.progress);
         downloadProgress.setVisibility(View.VISIBLE);
         downloadProgress.setIndeterminate(true);
-        DataDownload download = new DataDownload();
-        download.downloadVersions();
-
         this.findViewById(android.R.id.content).setBackgroundColor(Color.LTGRAY);
-
+        OpenEventApp.postEventOnUIThread(new DataDownloadEvent());
         if(savedInstanceState == null){
             currentMenuItemId = R.id.nav_tracks;
         } else {
@@ -182,6 +192,7 @@ public class MainActivity extends AppCompatActivity {
         }
 
         Snackbar.make(mainFrame, getString(R.string.download_complete), Snackbar.LENGTH_SHORT).show();
+        Log.d("DownNotif", "Download done");
     }
 
     private void downloadFailed() {
@@ -254,6 +265,16 @@ public class MainActivity extends AppCompatActivity {
         mDrawerLayout.closeDrawers();
     }
 
+    public void showErrorDialog(String errorType, String errorDesc){
+        downloadProgress.setVisibility(View.GONE);
+        AlertDialog.Builder builder = new AlertDialog.Builder(this);
+        builder.setTitle(R.string.error)
+                .setMessage(errorType + ": " + errorDesc)
+                .setNeutralButton(R.string.ok, null)
+                .create();
+        builder.show();
+    }
+
     //Subscribe Events
     @Subscribe
     public void onCounterReceiver(CounterEvent event) {
@@ -324,11 +345,6 @@ public class MainActivity extends AppCompatActivity {
         downloadFailed();
     }
 
-    @Subscribe
-    public void cantDownload(CantDownloadEvent event) {
-        downloadProgress.setVisibility(View.GONE);
-        Snackbar.make(mainFrame, getString(R.string.cantDownload), Snackbar.LENGTH_LONG).show();
-    }
 
     @Subscribe
     public void onEventsDownloadDone(EventDownloadEvent event) {
@@ -359,4 +375,91 @@ public class MainActivity extends AppCompatActivity {
 
     }
 
+
+    @Subscribe
+    public void showNetworkDialog(ShowNetworkDialogEvent event){
+        downloadProgress.setVisibility(View.GONE);
+        AlertDialog.Builder builder = new AlertDialog.Builder(MainActivity.this);
+        builder.setTitle(getString(R.string.net_unavailable))
+                .setMessage(getString(R.string.turn_on))
+                .setPositiveButton(R.string.action_settings, new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
+                        Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                        startActivity(setNetworkIntent);
+                    }
+                })
+                .setNeutralButton(R.string.cancel, null)
+                .create();
+        builder.show();
+    }
+
+    @Subscribe
+    public void downloadData(DataDownloadEvent event){
+        DataDownload download = new DataDownload();
+        download.downloadVersions();
+        downloadProgress.setVisibility(View.VISIBLE);
+        Log.d("DataNotif", "Download has started");
+    }
+
+    @Subscribe
+    public void ErrorHandlerEvent(RetrofitError cause) {
+        ConnectivityManager connMgr = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo netinfo = connMgr.getActiveNetworkInfo();
+        if(!(netinfo!=null && netinfo.isConnected())){
+            OpenEventApp.postEventOnUIThread(new ShowNetworkDialogEvent());
+        }
+        else
+        {
+            switch(cause.getKind()){
+                case CONVERSION: {
+                    Log.d(TYPE, "ConversionError");
+                    errorType="Conversion Error";
+                    errorDesc=String.valueOf(cause.getCause());
+                    break;
+                }
+                case HTTP: {
+                    Log.d(TYPE, "HTTPError");
+                    errorType="HTTP Error";
+                    errorDesc=String.valueOf(cause.getResponse().getStatus());
+                    Log.d(ERROR_CODE, String.valueOf(cause.getResponse().getStatus()));
+                    break;
+                }
+                case UNEXPECTED: {
+                    Log.d(TYPE, "UnexpectedError");
+                    errorType="Unexpected Error";
+                    errorDesc=String.valueOf(cause.getCause());
+                    break;
+                }
+                case NETWORK: {
+                    Log.d(TYPE, "NetworkError");
+                    errorType="Network Error";
+                    errorDesc=String.valueOf(cause.getCause());
+                    break;
+                }
+                default: {
+                    Log.d(TYPE, "Other Error");
+                    errorType="Other Error";
+                    errorDesc=String.valueOf(cause.getCause());
+                }
+            }
+            showErrorDialog(errorType, errorDesc);
+        }
+    }
+
+    @Subscribe
+
+    public void onConnectionChangeReact(ConnectionCheckEvent event)
+    {
+        if(event.connState())
+        {
+            OpenEventApp.postEventOnUIThread(new DataDownloadEvent());
+            Log.d("NetNotif", "Connected to Internet");
+        }
+        else
+        {
+            Log.d("NetNotif", "Not connected to Internet");
+            OpenEventApp.postEventOnUIThread(new ShowNetworkDialogEvent());
+        }
+    }
 }
--- a/app/src/main/java/org/fossasia/openevent/api/APIClient.java
+++ b/app/src/main/java/org/fossasia/openevent/api/APIClient.java
@@ -6,7 +6,6 @@ import com.squareup.okhttp.OkHttpClient;
 import org.fossasia.openevent.api.network.OpenEventAPI;
 
 import java.util.concurrent.TimeUnit;
-
 import retrofit.RestAdapter;
 import retrofit.converter.GsonConverter;
 
--- a/app/src/main/java/org/fossasia/openevent/api/Urls.java
+++ b/app/src/main/java/org/fossasia/openevent/api/Urls.java
@@ -28,5 +28,4 @@ public abstract class Urls {
 
     public static final String BASE_GET_URL_ALT = "https://raw.githubusercontent.com/fossasia/open-event/master/testapi";
 
-
 }
--- a/app/src/main/java/org/fossasia/openevent/api/processor/VersionApiProcessor.java
+++ b/app/src/main/java/org/fossasia/openevent/api/processor/VersionApiProcessor.java
@@ -7,7 +7,6 @@ import org.fossasia.openevent.api.protocol.VersionResponseList;
 import org.fossasia.openevent.data.Version;
 import org.fossasia.openevent.dbutils.DataDownload;
 import org.fossasia.openevent.dbutils.DbSingleton;
-import org.fossasia.openevent.events.CantDownloadEvent;
 import org.fossasia.openevent.events.CounterEvent;
 import org.fossasia.openevent.utils.CommonTaskLoop;
 
@@ -104,7 +103,9 @@ public class VersionApiProcessor implements Callback<VersionResponseList> {
 
     @Override
     public void failure(RetrofitError error) {
-        OpenEventApp.getEventBus().post(new CantDownloadEvent());
+        Log.d("Ret", "RetrofitError");
+        OpenEventApp.postEventOnUIThread(error);
+        Log.d("ErrorDescription", String.valueOf(error.getCause()));
     }
 
 }
--- a/app/src/main/java/org/fossasia/openevent/events/ShowNetworkDialogEvent.java
+++ b/app/src/main/java/org/fossasia/openevent/events/ShowNetworkDialogEvent.java
@@ -1,0 +1,18 @@
+package org.fossasia.openevent.events;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.provider.Settings;
+import android.support.v7.app.AlertDialog;
+import android.view.View;
+
+import org.fossasia.openevent.R;
+
+/**
+ * Created by shivenmian on 01/01/16.
+ */
+public class ShowNetworkDialogEvent {
+
+}
--- a/app/src/main/java/org/fossasia/openevent/Receivers/NetworkConnectivityChangeReceiver.java
+++ b/app/src/main/java/org/fossasia/openevent/Receivers/NetworkConnectivityChangeReceiver.java
@@ -1,0 +1,44 @@
+package org.fossasia.openevent.Receivers;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+import android.util.Log;
+
+import com.squareup.otto.Bus;
+
+import org.fossasia.openevent.OpenEventApp;
+import org.fossasia.openevent.activities.MainActivity;
+import org.fossasia.openevent.dbutils.DataDownload;
+import org.fossasia.openevent.events.ConnectionCheckEvent;
+import org.fossasia.openevent.events.DataDownloadEvent;
+import org.fossasia.openevent.events.ShowNetworkDialogEvent;
+
+/**
+ * Created by shivenmian on 02/01/16.
+ */
+public class NetworkConnectivityChangeReceiver extends BroadcastReceiver {
+
+    public boolean isConnected = false;
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.d("NetNotif", "Received network information");
+        ConnectionCheckEvent event = new ConnectionCheckEvent(isConnected);
+        if (isNetworkAvailable(context)) {
+                if (!event.connState()) {
+                    event.isConnected = true;
+                }
+        } else {
+            event.isConnected=false;
+        }
+        OpenEventApp.postEventOnUIThread(event);
+    }
+
+    private boolean isNetworkAvailable(Context context) {
+        ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+        return (networkInfo!=null && networkInfo.isConnected());
+    }
+}
--- a/app/src/main/java/org/fossasia/openevent/events/ConnectionCheckEvent.java
+++ b/app/src/main/java/org/fossasia/openevent/events/ConnectionCheckEvent.java
@@ -1,0 +1,17 @@
+package org.fossasia.openevent.events;
+
+/**
+ * Created by shivenmian on 04/01/16.
+ */
+public class ConnectionCheckEvent {
+
+    public boolean isConnected;
+    public ConnectionCheckEvent(boolean isConnected)
+    {
+        this.isConnected=isConnected;
+    }
+    public boolean connState()
+    {
+        return isConnected;
+    }
+}
--- a/app/src/main/java/org/fossasia/openevent/events/DataDownloadEvent.java
+++ b/app/src/main/java/org/fossasia/openevent/events/DataDownloadEvent.java
@@ -1,0 +1,7 @@
+package org.fossasia.openevent.events;
+
+/**
+ * Created by shivenmian on 03/01/16.
+ */
+public class DataDownloadEvent {
+}
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -58,9 +58,16 @@
     <!--Search Titles-->
     <string name="search_hint_tracks"> Search Tracks</string>
     <string name="refresh_failed">Couldn\'t Refresh</string>
-    <string name="cantDownload">Can\'t Download right now</string>
-    <string name="download_failed">Download Failed, Check Internet Connection</string>
+    <string name="download_failed">Download Failed</string>
     <string name="download_complete">Download Done</string>
     <string name="share_links">Sharing URL</string>
     <string name="time_not_specified">Timings not specified</string>
+
+    <!--Error Titles-->
+    <string name="net_unavailable">Network not available</string>
+    <string name="turn_on">Turn on WiFi or Mobile data in Settings</string>
+    <string name="cancel">Cancel</string>
+    <string name="error">Error</string>
+    <string name="ok">OK</string>
+
 </resources>


shivenmian:
@mananwason @creativepsyco Please check this and tell me if something more is required.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        download.downloadVersions();
+        ConnectivityManager connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+        if(networkInfo!=null && networkInfo.isConnected()){
+            download.downloadVersions();
+        }
+        else {
+            downloadProgress.setVisibility(View.GONE);
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
+            builder.setTitle("Network not available")
```
creativepsyco: Strings should go inside `Strings.xml`

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
         downloadProgress.setIndeterminate(true);
         DataDownload download = new DataDownload();
-        download.downloadVersions();
+        ConnectivityManager connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+        if(networkInfo!=null && networkInfo.isConnected()){
+            download.downloadVersions();
+        }
+        else {
+            downloadProgress.setVisibility(View.GONE);
```
creativepsyco: what happens if a dialog is already present.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,11 +44,61 @@ public APIClient() {
...
+                                Log.d("Type", "ConversionError");
+                                ErrorDialogEvent errEvent = new ErrorDialogEvent("Conversion Error", String.valueOf(cause.getCause()));
+                                OpenEventApp.postEventOnUIThread(errEvent);
+                                Log.d("ErrorType", String.valueOf(cause.getResponse().getStatus()));
+                            } catch (NullPointerException e) {
+                                Log.d("Exception", "NPE");
+                            }
+                        } else if (cause.getKind().equals(RetrofitError.Kind.HTTP)) {
+                            try {
+                                Log.d("Type", "HTTPError");
```
mananwason: Add tags to all the logs

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                        public void onClick(DialogInterface dialog, int which) {
+                            Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                            startActivity(setNetworkIntent);
+                            MainActivity.this.finish();
+                        }
+                    })
+                    .setNeutralButton("Cancel", new DialogInterface.OnClickListener() {
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+
```
creativepsyco: You should probably move this to a different class and pass the current activity as an argument instead of accessing it via `MainActivity.this.finish()`

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
```diff
@@ -0,0 +1,14 @@
+package org.fossasia.openevent.events;
+
+/**
+ * Created by shivenmian on 28/12/15.
+ */
+public class ErrorDialogEvent {
+
+    public String s1;
+    public String s2;
+    public ErrorDialogEvent(String s1, String s2){
```
mananwason: Write descriptive names for both the strings

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                    .setMessage("Turn on WiFi or Mobile data in Settings")
+                    .setPositiveButton("Settings", new DialogInterface.OnClickListener() {
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+                            Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                            startActivity(setNetworkIntent);
+                            MainActivity.this.finish();
+                        }
+                    })
+                    .setNeutralButton("Cancel", new DialogInterface.OnClickListener() {
```
creativepsyco: There is a dialog without a neutral button as well, again captions etc. go inside `strings.xml`

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
         downloadProgress.setVisibility(View.VISIBLE);
         downloadProgress.setIndeterminate(true);
         DataDownload download = new DataDownload();
-        download.downloadVersions();
+        ConnectivityManager connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+        if(networkInfo!=null && networkInfo.isConnected()){
+            download.downloadVersions();
+        }
+        else {
```
creativepsyco: Indentation, brace position.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -58,6 +65,7 @@
     private int currentMenuItemId;
     private final static String STATE_FRAGMENT = "stateFragment";
 
+
```
creativepsyco: empty line?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -72,7 +80,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                            Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                            startActivity(setNetworkIntent);
+                            MainActivity.this.finish();
+                        }
+                    })
+                    .setNeutralButton("Cancel", new DialogInterface.OnClickListener() {
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+
+                        }
```
creativepsyco: This method does nothing? Then why is this here.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -327,7 +360,8 @@ public void noInternet(NoInternetEvent event) {
     @Subscribe
     public void cantDownload(CantDownloadEvent event) {
         downloadProgress.setVisibility(View.GONE);
-        Snackbar.make(mainFrame, getString(R.string.cantDownload), Snackbar.LENGTH_LONG).show();
+
```
mananwason: Remove extra lines

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -359,4 +393,21 @@ public void onMicrolocationsDownloadDone(MicrolocationDownloadEvent event) {
 
     }
 
+    @Subscribe
+    public void ErrorDialog(ErrorDialogEvent event){
+        downloadProgress.setVisibility(View.GONE);
+        AlertDialog.Builder builder = new AlertDialog.Builder(this);
+        builder.setTitle("Error")
+                .setMessage(event.s1 + ": " + event.s2)
+                .setNeutralButton("OK", new DialogInterface.OnClickListener() {
```
creativepsyco: You should pass `null` as the listener if nothing is being done, again strings go into a separate file.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,11 +44,61 @@ public APIClient() {
...
+                            } catch (NullPointerException e) {
+                                Log.d("Exception", "NPE");
+                            }
+                        } else if (cause.getKind().equals(RetrofitError.Kind.NETWORK)) {
+                            try {
+                                Log.d("Type", "NetworkError");
+                                ErrorDialogEvent errEvent = new ErrorDialogEvent("Network Error", String.valueOf(cause.getCause()));
+                                OpenEventApp.postEventOnUIThread(errEvent);
+                                Log.d("ErrorType", String.valueOf(cause.getResponse().getStatus()));
+                            } catch (NullPointerException e) {
```
creativepsyco: Why are you catching null pointer exception?

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,11 +44,61 @@ public APIClient() {
...
+                                Log.d("Exception", "NPE");
+                            }
+                        } else {
+                            Log.d("Type", "Other Error");
+                            Log.d("ErrorType", String.valueOf(cause.getResponse().getStatus()));
+                            ErrorDialogEvent errEvent = new ErrorDialogEvent("Other Error", String.valueOf(cause.getCause()));
+                            OpenEventApp.postEventOnUIThread(errEvent);
```
creativepsyco: Too many repeated statements here, you can use a switch case and fix this up. use chaining (or builder pattern) to fix up generating an event based on the arguments involved like `event.setType().setCause().setCode()` etc.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,11 +44,61 @@ public APIClient() {
...
+                        }
+                        return new Exception(cause);
+                    }
+                })
                 .build();
         openEventAPI = adapter.create(OpenEventAPI.class);
+
```
creativepsyco: empty line

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,11 +44,61 @@ public APIClient() {
...
     }
 
     public OpenEventAPI getOpenEventAPI() {
         return openEventAPI;
+
```
creativepsyco: empty line.

app/src/main/java/org/fossasia/openevent/api/processor/EventListResponseProcessor.java
```diff
@@ -46,7 +46,6 @@ public void run() {
 
     }
 
-
```
creativepsyco: unexpected deletion?

app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
```diff
@@ -57,6 +57,6 @@ public void run() {
     @Override
     public void failure(RetrofitError error) {
         // Do something with failure, raise an event etc.
-        OpenEventApp.getEventBus().post(new SponsorDownloadEvent(false));
+        OpenEventApp.getEventBus().post(new SpeakerDownloadEvent(false));
```
creativepsyco: This apparently is a different fix, not related to showing up dialogs, why is this here, part of this commit?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: It does. This will be called when network is not present. The app won't start if network is not there.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
```diff
@@ -0,0 +1,14 @@
+package org.fossasia.openevent.events;
+
+/**
+ * Created by shivenmian on 28/12/15.
+ */
+public class ErrorDialogEvent {
+
+    public String s1;
```
creativepsyco: Better naming please.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -327,7 +360,8 @@ public void noInternet(NoInternetEvent event) {
     @Subscribe
     public void cantDownload(CantDownloadEvent event) {
         downloadProgress.setVisibility(View.GONE);
-        Snackbar.make(mainFrame, getString(R.string.cantDownload), Snackbar.LENGTH_LONG).show();
+
+
     }
```
creativepsyco: extra lines?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -187,7 +221,6 @@ private void syncComplete() {
     private void downloadFailed() {
         downloadProgress.setVisibility(View.GONE);
         Snackbar.make(mainFrame, getString(R.string.download_failed), Snackbar.LENGTH_LONG).show();
-
```
creativepsyco: unexpected deletion?

creativepsyco:
Requires heavy cleanup.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: I had this as an error once so added it. I'll remove it, it's not needed now.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Oops, did it accidentaly.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: I am talking about the `onClick` listener for the neutral button.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: I'd added something in this function and deleted it later. I'll remove them.

app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
shivenmian: Yeah this isn't related to showing up of dialogs, this is possibly a small typo @mananwason did, so I fixed it too.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -327,7 +360,8 @@ public void noInternet(NoInternetEvent event) {
     @Subscribe
     public void cantDownload(CantDownloadEvent event) {
```
creativepsyco: Why are there 2 events: `CantDownloadEvent` & `ErrorDialogEvent` I don't there should be 2 events for 1 point of failure, unless you have 2 different UI components or some background processing, you should consider using only one and remove the other.

creativepsyco:
Also commit message is missing, can use a better description than `Updated`

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Oh. Yeah that does nothing, you just go back to the activity. I'll remove it.

app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
creativepsyco: In either case, should not be part of your commit.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: cantDownload was already there, didn't touch this. I'll remove it, it isn't needed.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: @shivenmian please look at my comment above, I am talking about the `onClick` empty function, it is a dialog button click listener and has nothing to do with network conditions.

shivenmian: Yeah, I'll put null for it instead of onClick. I'd kept the Cancel button should the user not go to Settings and stay offline.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: It doesn't matter anyway, there wasn't anything there, just extra lines.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: I don't think a switch case is beneficial because I've used .equals() instead of ==, and the former is more preferred than the latter. 

I've chained stuff though, dropping a PR soon :)

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Why?

The purpose of MainActivity.this.finish() is to close the app as soon as I go to Settings, otherwise download won't happen even after switching WiFi/Data on.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: yes that would be better.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: You should implement a NetworkConnectivityChange receiver to monitor connectivity. And don't do it in the AndroidManifest, do it in the application onCreate. See http://stackoverflow.com/questions/12274997/why-broadcastreceiver-works-even-when-app-is-in-background and https://gist.github.com/aniXification/5896231

app/src/main/java/org/fossasia/openevent/api/APIClient.java
creativepsyco: I believe that switch-case is pretty efficient in terms of enums, which is the case here, nevertheless I feel switch case is more readable and easier to maintain than an if-then-else.

shivenmian:
@creativepsyco @mananwason I've fixed stuff, if there's anything more let me know.

For the ErrorDialogs, I created another function in the ErrorDialog.java and used it to shorten code.

app/src/main/java/org/fossasia/openevent/api/processor/VersionApiProcessor.java
```diff
@@ -104,7 +107,7 @@ public void run() {
 
     @Override
     public void failure(RetrofitError error) {
-        OpenEventApp.getEventBus().post(new CantDownloadEvent());
```
mananwason: Shouldn't you fire a ErrorDialogEvent here?

app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
```diff
@@ -47,7 +47,7 @@ public void run() {
                 dbSingleton.clearDatabase(DbContract.Speakers.TABLE_NAME);
                 dbSingleton.insertQueries(queries);
 
-                OpenEventApp.postEventOnUIThread(new SpeakerDownloadEvent(true));
+                OpenEventApp.postEventOnUIThread(new SponsorDownloadEvent(true));
             }
```
mananwason: I believe this should be SpeakerDownloadEvent Only. Please fix it.

shivenmian: You asked me to change it to Sponsor :$

mananwason: @creativepsyco asked you to change it in the failure callback.

shivenmian: Yeah I realised.

app/src/main/java/org/fossasia/openevent/api/processor/VersionApiProcessor.java
shivenmian: It gives two dialogs. I tried it.

shivenmian:
@mananwason @creativepsyco Fixed stuff. If there's anything more, please let me know.

To handle errors and tidy up code, I created a new function in ErrorDialog.java and used it throughout.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -86,6 +99,23 @@ protected void onCreate(Bundle savedInstanceState) {
 
     }
 
+    public void WiFiDialog(final Activity activity){
```
creativepsyco: starting method name with capitals? Also should use a verb for function name, like `showXXXX` instead of a noun.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -86,6 +99,23 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        downloadProgress.setVisibility(View.GONE);
+        AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+        builder.setTitle(getString(R.string.net_unavailable))
+                .setMessage(getString(R.string.turn_on))
+                .setPositiveButton(R.string.action_settings, new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
+                        Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                        startActivity(setNetworkIntent);
+                        activity.finish();
```
creativepsyco: I don't think you need to call activity.finish(), I know you are doing this to force download the information the next time the person comes back to the app, but I think that should be handled from a connectivity change receiver because network can change anytime and you need that.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -359,4 +382,15 @@ public void onMicrolocationsDownloadDone(MicrolocationDownloadEvent event) {
 
     }
 
+    @Subscribe
+    public void ErrorDialog(ErrorDialogEvent event){
```
creativepsyco: method names are improper, should be verbs and follow camel case notation. Please see the names of the functions within the file to get an idea.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
```diff
@@ -0,0 +1,59 @@
...
+import org.fossasia.openevent.api.APIClient;
+
+import retrofit.RetrofitError;
+
+/**
+ * Created by shivenmian on 28/12/15.
+ */
+public class ErrorDialogEvent {
+
+    public String ErrorType;
```
creativepsyco: class attributes starting with caps?

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
```diff
@@ -0,0 +1,59 @@
...
+    public String ErrorDesc;
+    public static final String TYPE = "RetrofitError Type";
+    public static final String ERROR_CODE = "Error Code";
+
+    public ErrorDialogEvent(String s1, String s2){
```
creativepsyco: descriptive string please, not `s1` `s2`

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,6 +48,15 @@ public APIClient() {
                 .setConverter(new GsonConverter(gson))
                 .setEndpoint(Urls.BASE_GET_URL_ALT)
                 .setLogLevel(RestAdapter.LogLevel.FULL)
+                .setErrorHandler(new ErrorHandler() {
+                    @Override
+                    public Throwable handleError(RetrofitError cause) {
+                        ErrorDialogEvent errEvent = new ErrorDialogEvent(null, null);
```
creativepsyco: Hmm. Why are you creating an object and then passing it to your handler?

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,6 +48,15 @@ public APIClient() {
...
+                        errEvent.ErrorHandlerFunction(cause, errEvent);
```
creativepsyco: Here you are calling a member function and passing the member as an argument. This is definitely not right.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,6 +48,15 @@ public APIClient() {
...
+                        OpenEventApp.postEventOnUIThread(new ErrorDialogEvent(errEvent.ErrorType, errEvent.ErrorDesc));
```
creativepsyco: You already created an object in the lines above. I am not sure what you are doing here.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -37,6 +48,15 @@ public APIClient() {
...
+                        return new Exception(cause);
```
creativepsyco: `RetrofitError` extends from `RuntimeException` I don't believe you need to create a new object here.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -28,7 +34,12 @@
 
     private final OpenEventAPI openEventAPI;
 
+    public static final String TYPE = "RetrofitError Type";
+
+    public static final String ERROR_CODE = "Error Code";
```
creativepsyco: unused strings

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -28,7 +34,12 @@
 
     private final OpenEventAPI openEventAPI;
 
+    public static final String TYPE = "RetrofitError Type";
```
creativepsyco: unused strings.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: I tried keeping my network on and then running the app and switching it off while the download is in progress. It gives a Network Error (Socket time-out exception) everytime.

Instead of adding a BroadcastReceiver, why not just add this case ^ to my existent method?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Done.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
```diff
@@ -0,0 +1,59 @@
...
+        this.ErrorType = s1;
+        this.ErrorDesc = s2;
+    }
+
+    public void ErrorHandlerFunction(RetrofitError cause, ErrorDialogEvent event){
+
+        if (cause.getKind().equals(RetrofitError.Kind.CONVERSION)) {
```
creativepsyco: please use switch case, and make this return an `ErrorDialogEvent` object.

creativepsyco:
@shivenmian Still a long way to go, lots of issues. Please fix. Thanks.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: We want to refresh the information whenever we have an active network connection and broadcast receivers allow us to do that.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: That's because if you look at ErrorHandlerFunction, I'm changing errorType and errorDesc based on the type of error (i.e cause). That's why I first call ErrorDialogEvent and pass null null, then change the member's parameters using the second line.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Oh okay, didn't realise you wanted this. Sure, will do.

app/src/main/java/org/fossasia/openevent/api/processor/EventListResponseProcessor.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: The reason I created the object was because all the error dialogs differed only in the type of error and the description, and that can be changed using one function and if-else / switch. Once you get that, you can simply post it.

app/src/main/java/org/fossasia/openevent/api/processor/VersionApiProcessor.java
creativepsyco: I think you should not depend on APIClient for your implementation. The code should be as decoupled as possible. You should continue firing the error or at least log it, but whether the MainActivity chooses to subscribe to it or whether it has shown a network error dialog, should be upto it.

Also I think that instead of setting an Error Handler on the Retrofit adapter you should piggy back on the event that the processor generates and transform the event into something that you can show (the dialog in your case). The reason I am saying this is because the error generated from the Processor is local, has more contextual information, you know exactly where the error is coming from, while for the Retrofit adapter it will cause the handler to fire for all global error issues. I don't think that should be the approach.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: Will do.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
creativepsyco: You can just create an `ErrorDialogEvent` from the function itself right?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: I mean whether you implement it as part of a new GCI task or you don't implement it is up to you. But at least the finish statement makes it look it the problem is fixed. But its not :)

shivenmian: I'll fix it. 

PS: I'm not doing GCI. I'm in college. GSoC :stuck_out_tongue:

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: I haven't looked at all the other issues you pointed, will do in a bit. But for this one, I've done this, look at these three images: [This](http://i.imgur.com/Bw41D12.png), [this](http://i.imgur.com/8AgNo3h.png) and [this](http://i.imgur.com/lebbW8J.png).

The first one is of the handleError function in APIClient, the next two are of HandleErrorDialogEvent.java (I changed the name).

Is this okay?

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: Fixed.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: I've made it return an ErrorDialogEvent object.

I tried Switch case with cause.getKind() being compared, but just as I thought, it didn't work. 

When I took cases to be compared as RetrofitError.Kind.CONVERSION etc, it gave an error saying "Constant Expression required". And it's apt. RetrofitError.Kind.CONVERSION is not an object so the error came.

Then I took the cases as merely CONVERSION, NETWORK, HTTP etc. The code worked without errors. But then I took the base url as a malformed JSON ([this](http://www.mocky.io/v2/5682128c120000993893a2c1)) and ran. Instead of conversion error, it gave HTTP 200 error.

The if-else had worked perfectly with the malformed JSON and it did give me conversion error. The problem with switch is that it uses '==' and this compares objects as a whole, while .equals() also compares parts. That's why switch won't work in this case as I expected and saw. Check [this](http://stackoverflow.com/questions/1643067/whats-the-difference-between-equals-and).

app/src/main/java/org/fossasia/openevent/api/APIClient.java
creativepsyco: I think we should drop the errorHandler in retrofit client, use the one from the process see [above](https://github.com/fossasia/open-event-android/pull/128#discussion_r48522034)

app/src/main/java/org/fossasia/openevent/api/APIClient.java
creativepsyco: I don't think we should do that because the error will be posted for all API calls, and this API Client is shared everywhere. The if-else or switch will create only 1 object so I don't really understand what's your point here:

> The reason I created the object errEvent was because all the error dialogs differed only in the type of error and the description

shivenmian: Yeah, the error dialogs only differ in the type of error and the description. Check out HandleErrorDialogEvent. The constructor takes in two strings.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
creativepsyco: My IDE seems to be okay with this:
image

>Then I took the cases as merely CONVERSION, NETWORK, HTTP etc. The code worked without errors. But then I took the base url as a malformed JSON (this) and ran. Instead of conversion error, it gave HTTP 200 error.

200 is success NOT error.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
shivenmian: Hmm. I'll give it a try.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: Yeah, I know. I just mean to say that it gives the wrong dialog. It's supposed to give an error.

http://www.mocky.io/v2/5682128c120000993893a2c1

This is a malformed JSON. This should give a Conversion Error instead of an HTTP 200.

creativepsyco: And I get 

image 
when I plug in http://www.mocky.io/v2/5682128c120000993893a2c1

I think it works, do you have more counterexamples?

creativepsyco:
@shivenmian I believe switch case works fine. Please give it a shot! Also as far as switch case vs comparison branches (I don't know from where you believe that switch uses `==` and not `equals`) is concerned please take a look at the [JVM spec][1], I don't want to argue but if you are interested clearly from spec we can see that N no of comparisons (if-else-if) are better a single table lookup (switch-case). And internally byte code maps switch case labels as integers, and enums are perfect for this because they are sparse and uniformly distributed ordinals.


[1]: https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-3.html

shivenmian:
Switch-case optimises, no doubt, but in my case it gave an error, for which I tried to come up with a plausible explanation.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
shivenmian: Err, was giving an error in my case. I'll try it once more. I had the exact same code, mind you.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
creativepsyco: if the constructor is available you can just create the object right? Why do you need to initialise it with `null` and then pass it. The thing I don't get is that whats the point of having a member function when you are using it as a static helper method. Plus your datastructure returns nothing and you create another copy of the object from it, which is completely baffling. Objects are *NOT* cheap.

app/src/main/java/org/fossasia/openevent/events/ErrorDialogEvent.java
creativepsyco: I just pulled your branch in, don't think there are any differerences. Here's my diff, just some formatting issues:

https://gist.github.com/creativepsyco/ea33fd373d79f52d47a7

shivenmian: I didn't really push since morning. I made a lot of changes. Do I push right now?

creativepsyco: You can push anytime do take a look at the issues we discussed here, fix them and hopefully you learn something from the process.

shivenmian: Yeah I've fixed most of what you've asked me to, myself. :)

I'm fixing that WiFi BroadcastReceiver thing and removing ErrorHandler from APIClient. Rest done. I'll push now, you can see.

shivenmian:
@mananwason @creativepsyco Fixed stuff since morning. Just pushed. Apart from the TODOs I have written, if there are any other issues please let me know.

creativepsyco:
Apart from the above issues we discussed I don't think much is required. Just need to fix up your function names, variable names to align with Java Style Guide. Remove if-else in favour of switch, add a broadcast receiver to detect connectivity changes to trigger download (you can do this in a separate PR) instead of finishing the activity. Use Processor#failure() events instead of handling globally through Retrofit. Then I think this is good to go.

shivenmian:
You did check my latest [commit](https://github.com/shivenmian/open-event-android/commit/f5f6bc28342b0ede5fbaf02dc63ee9da7d0ceddf), right?

creativepsyco:
Yes I did. Still has the same issues. https://github.com/shivenmian/open-event-android/commit/f5f6bc28342b0ede5fbaf02dc63ee9da7d0ceddf

shivenmian:
Just the TODOs you mean. Okay. I'll fix them.

creativepsyco:
No not just the TODOs please see the issues we have gone over in this discussion thread and check your diff again after pushing. You can see all the changes here: https://github.com/fossasia/open-event-android/pull/128/files

shivenmian:
Fixed the lines stuff and the function names I believe. The comments on https://github.com/fossasia/open-event-android/pull/128/files show only the things I marked as TODO. I think that's the only thing left now @creativepsyco

shivenmian:
I have that NetworkBroadcastReceiver fix left. I have done the rest. @mananwason @creativepsyco 

Please tell me if any other fix left, apart from removing the extra URLs in the URL file. Those are tests. I'll do that in the end.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -3,11 +3,15 @@
...
 
+import org.fossasia.openevent.OpenEventApp;
 import org.fossasia.openevent.api.network.OpenEventAPI;
+import org.fossasia.openevent.events.HandleErrorDialogEvent;
 
 import java.util.concurrent.TimeUnit;
 
+import retrofit.ErrorHandler;
 import retrofit.RestAdapter;
+import retrofit.RetrofitError;
```
creativepsyco: remove the imports as well.

app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -29,6 +33,7 @@
     private final OpenEventAPI openEventAPI;
 
     public APIClient() {
+
```
creativepsyco: remove the extra line.

creativepsyco:
ok only the TODOs left, I have added minor cleanup comments.

shivenmian:
Only one TODO is left. I converted to Switch-case, removed the error-handling from the RestAdapter. Only the NetworkBroadcastReceiver TODO is left.

@creativepsyco is the rest ok?

creativepsyco:
yes looks fine.

shivenmian:
@mananwason @creativepsyco Sorry for the long overdue. Was out the whole day. Just came back.

I've added the BroadcastReceiver, I think it works although I'm not fully sure. If there are any issues, please tell me.

app/src/main/AndroidManifest.xml
```diff
@@ -49,6 +49,7 @@
             android:enabled="true" />
 
         <receiver android:name=".Receivers.NotificationAlarmReceiver" />
+
```
creativepsyco: Empty line

shivenmian:
Apart from the lines, is the rest fine?

I wanted to register the broadcastreceiver on the Manifest but then decided against it.

Is the rest working?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -71,8 +80,9 @@ protected void onCreate(Bundle savedInstanceState) {
         downloadProgress = (ProgressBar) findViewById(R.id.progress);
         downloadProgress.setVisibility(View.VISIBLE);
         downloadProgress.setIndeterminate(true);
-        DataDownload download = new DataDownload();
-        download.downloadVersions();
+        registerReceiver(new NetworkDialogHandler(), new IntentFilter(Context.CONNECTIVITY_SERVICE));
```
creativepsyco: Receiver should be registered in Application#onCreate() because Activity can be created multiple times.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -71,8 +80,9 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        Intent intent = new Intent(Context.CONNECTIVITY_SERVICE);
+        sendBroadcast(intent);
```
creativepsyco: why are you sending a broadcast here?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
...
 
+    public class NetworkDialogHandler extends BroadcastReceiver{
+        private boolean isConnected = false;
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            Log.d("NetNotif", "Received notif about net status");
+            isNetworkAvailable(context);
+        }
+
+        private boolean isNetworkAvailable(Context context){
```
creativepsyco: this function is returning a `boolean` but you don't seem to using it anywhere

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
         mDrawerLayout.closeDrawers();
     }
 
+    public class NetworkDialogHandler extends BroadcastReceiver{
+        private boolean isConnected = false;
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            Log.d("NetNotif", "Received notif about net status");
```
creativepsyco: Probably a better log message indicating what are the contents of the intent.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
...
+        }
+
+        private boolean isNetworkAvailable(Context context){
+            ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+            NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+            if(networkInfo!=null){
+                if(networkInfo.getState().equals(NetworkInfo.State.CONNECTED)){
+                    if(!isConnected){
+                        isConnected=true;
+                        DataDownload download = new DataDownload();
```
creativepsyco: You should probably check if a data download is already in progress or not.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
...
+                        download.downloadVersions();
+                        Log.d("NetNotif", "Connected to Internet");
+
+                    }
+                    return true;
+                }
+            }
+            Log.d("NetNotif", "Not connected to Internet");
+            showNetworkDialog(MainActivity.this);
```
creativepsyco: you cannot call view or activity related code from a BroadcastReceiver. This is wrong usage,

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
...
+                    if(!isConnected){
+                        isConnected=true;
+                        DataDownload download = new DataDownload();
+                        download.downloadVersions();
+                        Log.d("NetNotif", "Connected to Internet");
+
+                    }
+                    return true;
+                }
+            }
```
creativepsyco: where is the `else` branch.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -254,6 +281,36 @@ private void doMenuAction(int menuItemId){
...
+            Log.d("NetNotif", "Not connected to Internet");
+            showNetworkDialog(MainActivity.this);
+            isConnected=false;
+            return false;
+        }
```
creativepsyco: Code is not formatted properly.

app/src/main/java/org/fossasia/openevent/api/Urls.java
```diff
@@ -27,6 +27,6 @@
     public static final String BASE_GET_URL = BASE_URL + "/api/" + API_VERSION;
 
     public static final String BASE_GET_URL_ALT = "https://raw.githubusercontent.com/fossasia/open-event/master/testapi";
-
+
```
creativepsyco: unclean.

creativepsyco:
@shivenmian formatting issues + Broadcast receiver cannot communicate to any view or activity you should use event bus events for that. also please do a check so that downloading info is not called multiple times.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
shivenmian: registerReceiver is already in Mainactivity#onCreate(). Please check.

Or do you mean I should keep it in OpenEventApp#oncreate()? OpenEventApp extends Application.

creativepsyco: Application not Activity

shivenmian:
@creativepsyco @mananwason Sorry for being a bit late. Was stuck up a bit lately.

I've implemented the broadcastreceiver and all the features you asked, it's working fine for me.

If there are any other issues, please tell me.

Oh, and Happy New Year! :smiley:

app/src/main/AndroidManifest.xml
```diff
@@ -49,6 +49,12 @@
             android:enabled="true" />
 
         <receiver android:name=".Receivers.NotificationAlarmReceiver" />
+        <receiver android:name=".Receivers.NetworkDialogHandler">
```
creativepsyco: No need for the receiver to be declared here if you are registering it in the `onCreate` of the Application.

app/src/main/java/org/fossasia/openevent/OpenEventApp.java
```diff
@@ -1,11 +1,16 @@
...
+import android.content.Context;
+import android.content.IntentFilter;
 import android.os.Handler;
 import android.os.Looper;
+import android.support.v7.app.AppCompatActivity;
 
 import com.squareup.otto.Bus;
 
+import org.fossasia.openevent.Receivers.NetworkDialogHandler;
+import org.fossasia.openevent.activities.MainActivity;
```
creativepsyco: Why these imports are here?

app/src/main/AndroidManifest.xml
shivenmian: No, I didn't register in the onCreate. I registered it statically.

shivenmian:
Fixed the lines and the redundant imports. Are there any other issues? Please tell me if so. @creativepsyco

app/src/main/AndroidManifest.xml
creativepsyco: When you register it in the Manifest it gets activated everytime the network changes, even if your app isn't working or killed by the system, by registering in the manifest you are infact making the phone do more work even when the app is not active. Do you know why wasn't it working? Its a broadcast receiver only getting triggered when the network changes. So your testing steps should be to switch off network, startup the app and then switch the network back on again.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -71,10 +78,15 @@ protected void onCreate(Bundle savedInstanceState) {
         downloadProgress = (ProgressBar) findViewById(R.id.progress);
         downloadProgress.setVisibility(View.VISIBLE);
         downloadProgress.setIndeterminate(true);
-        DataDownload download = new DataDownload();
-        download.downloadVersions();
-
         this.findViewById(android.R.id.content).setBackgroundColor(Color.LTGRAY);
+        ConnectivityManager connMgr = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
```
creativepsyco: Why is this here? Shouldn't the processor be firing the `ShowNetworkDialogEvent`? Maybe the connection is weaker, maybe it takes forever to download the info. Or maybe there is a network timeout. You are just handling one situation here.

app/src/main/java/org/fossasia/openevent/events/ShowNetworkDialogEvent.java
```diff
@@ -0,0 +1,23 @@
...
+import android.view.View;
+
+import org.fossasia.openevent.R;
+
+/**
+ * Created by shivenmian on 01/01/16.
+ */
+public class ShowNetworkDialogEvent {
+
+    public Context context;
```
creativepsyco: never cache context anywhere, it is a memory leak.

app/src/main/AndroidManifest.xml
shivenmian: I'd done exactly that but I guess dynamic registration should be avoided or something, no? Most people do it in the manifest or so.

creativepsyco:
Hi @shivenmian I am unable to understand what's your approach. You should probably discuss why are you doing certain things over Gitter or IRC. For the me the ideal way to inform the user about network errors would be the activity just starts the Data download action, it doesn't care about whether the network is on or off. The processor that gets triggered later will throw an error if data download could not be successful, you can then check whether it is the network that is the problem or whether server is down and so on and fire the relevant events.

The network receiver will just fire an event telling whether network came back offline/online. The activity can listen to this event and restart the download if previously it didn't. You can track whether there is a download request in progress by using a boolean member variable in the activity.

Is that clear info for you? Please discuss this strategy, I feel commenting on the PR and you repeating your mistakes is a waste of my time and yours as well.

shivenmian:
My approach was to first check about Network presence and then call DataDownload, where the other errors are sorted out. That's because when Network isn't present, I intended on leading the user to the Settings page, while for other errors, the user just gets notified about them. Once network presence is confirmed, the other connection errors will be fired if they are present, but I intended to do a pre-check for Network presence because of the Settings thing.

shivenmian:
@creativepsyco Please check Gitter, I've explained what I've done in the code. There are a lot of comments as I didn't want to delete anything before being sure. I'll cleanup later.

Please tell me about any other issues.

app/src/main/AndroidManifest.xml
creativepsyco: we need to do dynamic registration. A lot of people don't understand the difference, please make sure that you do.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -359,4 +386,86 @@ public void onMicrolocationsDownloadDone(MicrolocationDownloadEvent event) {
...
+            }
+            case UNEXPECTED: {
+                Log.d(TYPE, "UnexpectedError");
+                errorType="Unexpected Error";
+                errorDesc=String.valueOf(cause.getCause());
+                break;
+            }
+            case NETWORK: {
+                if(!(netinfo!=null && netinfo.isConnected())){ //Network not connected error
+                    OpenEventApp.postEventOnUIThread(new ShowNetworkDialogEvent(MainActivity.this));
```
creativepsyco: Yes this is what was in my mind. But you can move the switch-case to a static function. like 
```java
Retrofit error = /* Obtained from somewhere */
errorType = NetworkUtils.getNetworkError(error)
if (errorType == null && !connected()) {
    // if network error dialog is not showing
    // show network connection dialog
} else {
    // display normal error dialog.
}
```

shivenmian:
@creativepsyco Please check this PR out and tell me if there are any other issues.

I'm a beginner at minimalising stuff so if there's something silly I've missed out, please tell me about it.

app/src/main/AndroidManifest.xml
```diff
@@ -49,6 +49,12 @@
             android:enabled="true" />
 
         <receiver android:name=".Receivers.NotificationAlarmReceiver" />
+        <receiver android:name=".Receivers.NetworkConnectivityChangeReceiver">
+            <intent-filter>
+                <action android:name="android.net.conn.CONNECTIVITY_CHANGE" />
+            </intent-filter>
+        </receiver>
+
```
creativepsyco: why is this still here?

shivenmian: Oops. Removing it.

shivenmian:
Fixed @creativepsyco . Any other issues?

creativepsyco:
yes rest is fine.

creativepsyco:
Will merge thanks :+1:

creativepsyco:
Although there are a few more formatting issues (I will fix those) rest is fine, You got what I wanted to convey to you in the end which is great :) hopefully you continue to contribute to this project in the future. Cheers!

shivenmian:
Yeah I'm definitely contributing. Cheers!

creativepsyco:
Before committing you should reformat each file that you are changing to fix up formatting according to Android Studio Default style.

See option below, it will take care of lining up braces etc.
image

shivenmian:
Hmm. Thanks so much! :smile: :smile: :smile:

PR #133, opened 2016-01-03
App crashing on long scrolling of Speaker's list FIXED

### LINKED ISSUE #121: App crashing on scrolling for long time on the speaker's list
Application get crashed few times when we scroll for long time on the speaker's list.Its been fixed, I have make a pull request #119

vjs3:
Application was crashing on long or infinite scrolling of speaker's list has been fixed #121


vjs3:
@mananwason as told by you I have make a new pull request for my issue, please consider this and ignore previous PR on issue #121

app/src/main/java/org/fossasia/openevent/utils/CircleTransform.java
```diff
@@ -24,7 +24,7 @@ public Bitmap transform(Bitmap source) {
             source.recycle();
         }
 
-        Bitmap bitmap = Bitmap.createBitmap(size, size, source.getConfig());
+        Bitmap bitmap = Bitmap.createBitmap(size, size, Bitmap.Config.ARGB_8888);
```
creativepsyco: @vjs3 Do you have a stack trace for the crash? I am not quite sure if this is right.

shivenmian:
Even I found out this error but @mananwason tells me it's an API issue, not the program.

vjs3:
I found out that this error is found many times in cases when we have long
scrolling list. This can easily be fixed by making some small changes in
code. It's not an Api error as much as I read on net. I will provide log
crash also.

app/src/main/java/org/fossasia/openevent/utils/CircleTransform.java
vjs3: I have to produce it again by using the older code I think, if there is any other way than please suggest me I will show you the stack trace.

mananwason:
@vjs3 The issue was due to using Circular Transform for Picasso. This was the Stack trace I believe 
`java.lang.NullPointerException: Transformation circle returned null after 0 previous transformation(s).`

app/src/main/java/org/fossasia/openevent/utils/CircleTransform.java
creativepsyco: I just need to see the stacktrace. I am sure there is one right?

vjs3:
@creativepsyco stack trace is same as written by @mananwason (java.lang.NullPointerException: Transformation circle returned null after 0 previous transformation(s)). The code change which I have done remove this problem.

app/src/main/java/org/fossasia/openevent/utils/CircleTransform.java
vjs3: @creativepsyco as commented by me and @mananwason the stack trace is as stated below.Let me know if you want more detailed one but the null pointer was due to the below-stated reason only.

creativepsyco:
I am not able to reproduce it.

vjs3:
The way to reproduce it is to scroll speaker list fastly for some time then sudenly app get crashed.I have used API level 19 (Android v 4.4).

creativepsyco:
The problem in your code is that you are assuming the Bitmap decode config to be always `ARGB_8888` which is absolutely plain wrong. Your transformation could be used anywhere you cannot assume that it is used for only 1 kind of configs

Well technically you are recycling the bitmap so does the config gets recycled/destroyed too? We should always check for `getConfig()` to potentially return a `null` value. We should not assume that `ARGB_8888` is always the config that is desired. What if the original image had a different config, then you have a mismatch.

See [GitHub issue][issue] for details on this.

[issue]: https://github.com/square/picasso/issues/997

vjs3:
@creativepsyco I get your point and I think if we do changes as suggested in that issue whenever we are getting null pointer that means its gif image we can use ARGB_8888 and if not than we can keep using getConfig().Let me know if this looks feasible to you.

creativepsyco:
you should not force defaulting to `ARGB_8888 ` you should check and cache config before recycling. Only if it is null then you should go ahead with the default config choice. Please change the code to reflect this.

vjs3:
Sure I will do that.

*Vishwajeet Srivastava*
B.Tech, Pre-Final (Computer Science and Engineering)
The LNM Institute of Information Technology
Mob:+91-9461787316

vjs3:
I apologise due to some android studio issues some wrong changes has been pushed here.I will make a new PR with changes as suggested by @creativepsyco.

creativepsyco:
You can force push to the same branch. It will appear here.

vjs3:
@creativepsyco i have tried to push the same branch but it is aksing me to make a new PR when I am making a PR.

creativepsyco:
You should force push to the same branch from which you created the
original PR from.

vjs3:
I have deleted the previous fork and I have again taken the fork does this
not make any difference?

*Vishwajeet Srivastava*
B.Tech, Pre-Final (Computer Science and Engineering)
The LNM Institute of Information Technology
Mob:+91-9461787316

creativepsyco:
Its okay just create a new PR and ping me.

vjs3:
I have created a a new PR just check it out #144

PR #142, opened 2016-01-08
Tests for SplashActivity

PoprostuRonin:


mananwason:
@PoprostuRonin Looks good. Please squash commits.

PoprostuRonin:
Done.

creativepsyco:
This test is just testing for wrap_content and alignment of the view objects I don't think that is something that should be tested. Also I would appreciate if the tests were making use of Espresso Testing framework and testing something more concrete. Tests should be adding value to the code base.

PR #144, opened 2016-01-11
Application crashing on long scroll of speaker's list FIXED

vjs3:
@creativepsyco here is the new PR with updated code.Please review it. #121


shivenmian:
I think you should remove the comments and then squash and commit again.

vjs3:
@shivenmian I have added comment on purpose, for the increased readability of code.

PR #145, opened 2016-01-13
Added Temporary Logger 

shivenmian:
For task 2 in Issue #127 , I've created a logger.

In this, for every *error* (I've covered those errors I mentioned in #128), there is an option to mail an error report (I've used my own email ID for receiving those reports: shiven.exun97@gmail.com). This error report will also be downloaded to the Downloads folder of the phone.

**Issues:**

1. Instead of my own Email ID, it has to be replaced with FOSSASIA's Email ID, or whoever the developer is. Or maybe something automated.

2. The file is being downloaded to the Downloads folder. I don't think the user should be able to see / have access to this file. Even though for every error report, this file gets refreshed, but the user shouldn't know it's present. I tried to save the file in the Internal Storage but you can't send attachments from there. Gotta fix this.

3. This error sending opens up the Email app which actually ruins the purpose of automation. I've looked it up and many links use the GMail API and SMTP, but they authenticate the Email account at the start, which also defeats the purpose.

**Enhancements I thought of:**

Issue 3 as noted above is the main one. What we require is this:

 1. Log sending should be done with an automated ID. And the receiver should be a server. I think we could use the orga server for both sending and receiving these logs.

 2. All of this should run in the background with no user interference.

I spoke to @mananwason and he said the orga server is not set up as of now for performing this task. We could do this.

For the current system I've adopted, I came across [this](https://github.com/yesidlazaro/GmailBackground) which sends emails in the background. It could work for now, but I don't know if I'm allowed to import this. Should I?

@mananwason @creativepsyco Please check my Issues and the enhancements I suggested. And please tell me about issues in my current code, how to resolve the Issues I pointed out, and suggest some other enhancements which could be done. Thanks :smiley:


mananwason:
@shivenmian Please save the logs in Internal storage. It should be doable using content providers.

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -269,12 +339,39 @@ public void showErrorDialog(String errorType, String errorDesc){
...
+                    }
+                })
                 .create();
         builder.show();
     }
 
+    public void sendEmail() {
+        Intent emailIntent = new Intent(Intent.ACTION_SENDTO);
+        emailIntent.setType("jpeg/image");
+        emailIntent.setData(Uri.parse("mailto:shiven.exun97@gmail.com"));
```
mananwason: Make this a constant that can be changed from one class.

app/src/main/java/org/fossasia/openevent/api/Urls.java
```diff
@@ -26,6 +26,8 @@
 
     public static final String BASE_GET_URL = BASE_URL + "/api/" + API_VERSION;
 
-    public static final String BASE_GET_URL_ALT = "https://raw.githubusercontent.com/fossasia/open-event/master/testapi";
+    //public static final String BASE_GET_URL_ALT = "https://raw.githubusercontent.com/fossasia/open-event/master/testapi";
+
+    public static final String BASE_GET_URL_ALT = "http://google.com/404";
```
mananwason: Remove this Url

shivenmian: Yeah, that was just a test. I wrote in the PS.

shivenmian:
Yeah I'll try the content providers thing. Thanks.

mananwason:
@rafalkowalski I think we should have this functionality in the serve, where we can send logs and the server can send the mails to the developers. Views @rafalkowalski @creativepsyco ?

creativepsyco:
Server should probably set up something like sentry but setting it up will
be sort of another responsibility. I am not sure if too many  components
are a distraction. We can table this for later. Let's get the server and
client ready for 2016 fossasia

mananwason:
Agreed. Let's first get the app and the server ready for fossasia 2016.

shivenmian:
@mananwason So do I do that content provider thing? And @creativepsyco did you see the github repo I wrote about which sends background emails? 

Looks like you're putting this for later. So should I work on this?

mananwason:
Yes @shivenmian do the content provider thing. Implementing the new feature
on the server is not a priority right now.

creativepsyco:
Hi @shivenmian @mananwason,

I have not heard anyone use Content Providers for Logs. I don't think its a brilliant idea, you can go ahead and prove me wrong :+1: 

Although at this point of time we don't have the server to get the log files. We certainly can still do the local log storage and send via email part (or perhaps even a built-in log viewer to view locally on the device).

I am summarizing these here, but really these are part of standard Android code practices.
* You are assuming there is only 1 component `MainActivity` which does the initialization of the app. Please think for a moment if that assumption is true. You can have more than 1 component in the app, `Services`, `BroadcastReceivers` just to name a few. Plus there very well can be more than one activity, so you should do your initialization part in `application#onCreate`
* Never expose your or anyone's email in source code, this is a public code site. We have no interest in knowing your email address. You should provide a button which provides a dialog, accept email address and sends it (with additional comments from the person using it.)
* Your function names are so weird, I told you previously **every function is a verb NOT a noun**. Still your name `errorLogHandler` represents a function name, which clearly isn't a Verb. Using verb as function helps enforce the convention that the function **DOES** something instead of saying **it is** something. Please read [Google Style Guide for Java][1] carefully multiple times if you are unsure how your code names work and if it is counter-intuitive to how you code.
* Also you are assuming there is an external storage directory and the whole log thing fails if there isn't one. What do you do then? Should you display error somewhere. Or should you check if you store the log somewhere else. Like in the application's cache directory. From Marshmallow onwards, file permissions & write access is pretty fine grained. I can guarantee you that this code will crash because you never asked for permission.
* Your log messages are so weird. `Cannot write logfile`. I am pretty sure the one looking at the log message will need to check what the hell really happened. Don't you think you have a responsibility to provide the best possible snapshot of the situation going on? I am sure the person will ask the question, as to why this is happening. So good log messages are important.
* Also when you initialize, you probably want to note down, app version, android version etc in the log. The log for different versions can vary. I suppose when debugging with code, you probably want to know what's the version for the thing that you are debugging.
* Limits on Log sizes - I am sure you cannot fill up the user's directory with the logfile. Have a sensible constant to limit the log to a certain size. If it grows beyond it, just clean up the whole file.

[1]: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

shivenmian:
@creativepsyco 

1. As of now, I'm clearing the log file every time the application opens, so I don't think the log file size can exceed anything. 

2. Okay, I'll note the app version etc down, and put in in OnCreate of the application.

3. I used those log messages for my own debugging, I didn't intend that this logger would be merged right now, that's why it's a "Temporary Logger". As I wrote in my PS, I promise I'll definitely sort out the function name and the messages in the end, once I get the functionality working, which is obviously more important.

4. Yeah, Marshmallow has a ``requestpermission()`` thing for dangerous permissions, which needs to be there even though you specified the permission in the Manifest. I was testing with non API 23 so I forgot to add it. I'll surely add it as ``WRITE_EXTERNAL_STORAGE`` is indeed dangerous. And I'll put a check if storage is present too.

5. @mananwason and I wanted to store the file in the Internal Storage instead of External, so that the viewer can't see the file. I'd tried to do it using ``getFilesDir()`` but didn't work, so he suggested Content Provider. If Content Provider is a bad idea, what do you think should be done if the log file should be kept in the Internal memory instead of external? Or should I go with external as of now?

mananwason:
@creativepsyco, @shivenmian told me that he wasn't able to attach files stored in the Internal storage and I told him that there's definitely a way and One way to do so is given [here] (http://stackoverflow.com/questions/13482310/how-to-share-internal-storage-file-with-gmail-client). My main point to him was that you can't rely on storing files to External storage. I am not sure if this would be a legible way or not, because I haven't used this functionality as of yet. @creativepsyco If you have a better way, do share it !! :D

mananwason:
okay I found another link [here](http://stephendnicholas.com/archives/974). Same thing is done here, creating and attaching cache files, using Content Providers.

creativepsyco:
@shivenmian 
- I don't think functionality is important for a Pull-Request. GitHub PR are for tracking code-reviews. Code readability is important. Everyone is busy, you can't waste their time by telling them that you are doing it in a certain way because you can't figure it out. IRC, Slack, Gitter, GitHub issues are all communication tools for sharing ideas and/or discussing a feature. Don't pollute the sanctity of code reviews by using it for asking comments from reviewers when you know it's not a final piece of code that you wish be merged with the master branch. PRs should be used for final-final versions of **mergeable** code.
- You are assuming that you still need to use external storage to store your log file. You should probably first ask yourself why? Because if you take this path, then you have to ask for permission, add extra permission etc. and its gonna be more code which honestly I don't think makes justification for this feature. Storing in `cacheDir` is possible, you don't need permission of any kind for it. Please look at that.
- Content Providers are relational data structures. From Android Developer site:

> Content providers manage access to a structured set of data. They encapsulate the data, and provide mechanisms for defining data security. Content providers are the standard interface that connects data in one process with code running in another process.

##### And I don't think collecting logs is a kind of structured data. Do you think it is?

- Dont add extra libraries. Just attach it to intent and let the person use the email client of his choice to do the job.

PR #153, opened 2016-01-20
 Fixed Navigation Closes #152

### LINKED ISSUE #152: Navigation issue
Going to Settings and then trying to navigate back to the previous activity results in a weird lag. 
Have a look below: 
ezgif com-video-to-gif

This error could be caused due to the following reasons: 
* finish() is not called in the MainActivity, when SettingsActivity is started.
* Using onBackPressed(); instead of finish() ;

Yathannsh:


app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -257,7 +257,7 @@ private void doMenuAction(int menuItemId){
                 Intent intent = new Intent(MainActivity.this, SettingsActivity.class);
                 mDrawerLayout.closeDrawers();
                 startActivity(intent);
-
+				finish();
```
creativepsyco: This is not a clean solution, `closeDrawers()` is an asynchronous operation. You should wait for the drawers to close before starting an activity.

Yathannsh:
I will try and implement the clean solution after learning a bit more. 
Thanks for your feedback! :smiley:

Issue #154, opened 2016-01-21
Intent Missing in Location and Speaker's List.


opticod:
Currently an user can't check the session details from Speaker or Location Activity.He has to switch back to track activity track and then check the session details.

I am willing to fix this. @creativepsyco @mananwason

mananwason:
My bad. #155 is still in progress. Hence reopening.

opticod:
Merged via  dd7827 .Hence closing.

PR #155, opened 2016-01-21
FIXED #162, MISSING INTENT ADDED #154 + Refactor

### LINKED ISSUE #162: BUG: onClick of any item causes starting of wrong session details.
I have attached a animated video of it.
Clicking on 'Purpose of Entry' causes launch of 'Strategise...'
ezgif com-video-to-gif 2

opticod:
Intent added #154

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/activities/LocationActivtiy.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/LocationActivtiy.java
@@ -1,10 +1,16 @@
 package org.fossasia.openevent.activities;
 
+import android.content.Intent;
 import android.os.Bundle;
-import android.support.v7.widget.*;
+import android.support.v7.widget.DefaultItemAnimator;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.SearchView;
+import android.support.v7.widget.Toolbar;
 import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.view.View;
 
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.adapters.SessionsListAdapter;
@@ -52,10 +58,24 @@ public class LocationActivtiy extends BaseActivity implements SearchView.OnQuery
 
         sessionRecyclerView = (RecyclerView) findViewById(R.id.recyclerView_locations);
         mSessions = dbSingleton.getSessionbyLocationName(location);
+
         sessionsListAdapter = new SessionsListAdapter(mSessions);
+
         sessionRecyclerView.setLayoutManager(new LinearLayoutManager(this));
         sessionRecyclerView.setAdapter(sessionsListAdapter);
+        sessionsListAdapter.setOnClickListener(new SessionsListAdapter.SetOnClickListener() {
+            @Override
+            public void onItemClick(int position, View view) {
+
+                Session model = (Session) sessionsListAdapter.getItem(position);
+                String sessionName = model.getTitle();
+                Intent intent = new Intent(getApplicationContext(), SessionDetailActivity.class);
+                intent.putExtra(IntentStrings.SESSION, sessionName);
+                startActivity(intent);
+            }
+        });
         sessionRecyclerView.setItemAnimator(new DefaultItemAnimator());
+
         if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
             searchText = savedInstanceState.getString(SEARCH);
         }
--- a/app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
@@ -20,6 +20,7 @@ import org.fossasia.openevent.api.Urls;
 import org.fossasia.openevent.data.Session;
 import org.fossasia.openevent.data.Speaker;
 import org.fossasia.openevent.dbutils.DbSingleton;
+import org.fossasia.openevent.utils.IntentStrings;
 import org.fossasia.openevent.utils.SpeakerIntent;
 
 import java.util.ArrayList;
@@ -98,6 +99,17 @@ public class SpeakersActivity extends AppCompatActivity implements SearchView.On
         sessionsListAdapter = new SessionsListAdapter(mSessions);
         sessionRecyclerView.setLayoutManager(new LinearLayoutManager(this));
         sessionRecyclerView.setAdapter(sessionsListAdapter);
+        sessionsListAdapter.setOnClickListener(new SessionsListAdapter.SetOnClickListener() {
+            @Override
+            public void onItemClick(int position, View view) {
+
+                Session model = (Session) sessionsListAdapter.getItem(position);
+                String sessionName = model.getTitle();
+                Intent intent = new Intent(getApplicationContext(), SessionDetailActivity.class);
+                intent.putExtra(IntentStrings.SESSION, sessionName);
+                startActivity(intent);
+            }
+        });
         sessionRecyclerView.setItemAnimator(new DefaultItemAnimator());
         if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
             searchText = savedInstanceState.getString(SEARCH);
--- a/app/src/main/java/org/fossasia/openevent/adapters/LocationsListAdapter.java
+++ b/app/src/main/java/org/fossasia/openevent/adapters/LocationsListAdapter.java
@@ -1,6 +1,5 @@
 package org.fossasia.openevent.adapters;
 
-import android.support.v7.widget.RecyclerView;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -10,6 +9,7 @@ import android.widget.TextView;
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.data.Microlocation;
 import org.fossasia.openevent.dbutils.DbSingleton;
+import org.fossasia.openevent.utils.ViewHolder;
 
 import java.text.MessageFormat;
 import java.util.ArrayList;
@@ -21,7 +21,7 @@ import timber.log.Timber;
  * User: MananWason
  * Date: 8/18/2015
  */
-public class LocationsListAdapter extends BaseRVAdapter<Microlocation, LocationsListAdapter.Viewholder> {
+public class LocationsListAdapter extends BaseRVAdapter<Microlocation, ViewHolder.Viewholder> {
     @SuppressWarnings("all")
     Filter filter = new Filter() {
         @Override
@@ -48,6 +48,7 @@ public class LocationsListAdapter extends BaseRVAdapter<Microlocation, Locations
             animateTo((List<Microlocation>) results.values);
         }
     };
+    private ViewHolder.SetOnClickListener listener;
 
     public LocationsListAdapter(List<Microlocation> microlocations) {
         super(microlocations);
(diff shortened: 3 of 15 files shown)


app/src/main/java/org/fossasia/openevent/activities/LocationActivtiy.java
```diff
@@ -50,6 +54,17 @@ protected void onCreate(Bundle savedInstanceState) {
         sessionRecyclerView.setLayoutManager(new LinearLayoutManager(this));
         sessionRecyclerView.setAdapter(sessionsListAdapter);
         sessionRecyclerView.setItemAnimator(new DefaultItemAnimator());
+        sessionRecyclerView.addOnItemTouchListener(
+                new RecyclerItemClickListener(getApplicationContext(), new RecyclerItemClickListener.OnItemClickListener() {
+                    @Override
+                    public void onItemClick(View view, int position) {
+                        String session_name = ((TextView) findViewById(R.id.session_title)).getText().toString();
```
creativepsyco: you should make use of `adapter.getItem(position)` to get the data not through the view.

app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
```diff
@@ -65,6 +69,17 @@ protected void onCreate(Bundle savedInstanceState) {
         sessionRecyclerView.setLayoutManager(new LinearLayoutManager(this));
         sessionRecyclerView.setAdapter(sessionsListAdapter);
         sessionRecyclerView.setItemAnimator(new DefaultItemAnimator());
+        sessionRecyclerView.addOnItemTouchListener(
+                new RecyclerItemClickListener(getApplicationContext(), new RecyclerItemClickListener.OnItemClickListener() {
+                    @Override
+                    public void onItemClick(View view, int position) {
+                        String session_name = ((TextView) findViewById(R.id.session_title)).getText().toString();
```
creativepsyco: Same here

opticod:
Done,@creativepsyco ,  fixed #162 along with adding missing intent #154

creativepsyco:
Ok give me until 2moro to review this. Thanks for the patience.

opticod:
:smiley:

opticod:
There was some merge conflict...I have resolved them also

app/src/main/java/org/fossasia/openevent/adapters/SessionsListAdapter.java
```diff
@@ -28,7 +31,9 @@ public SessionsListAdapter(List<Session> sessions) {
         Viewholder viewholder = new Viewholder(view);
         return viewholder;
     }
-
+    public static interface AdapterOnClickHandler {
+        void onClick(String name, Viewholder vh);
```
creativepsyco: why pass the name? shouldn't you be passing the position? then you can get the item from the `adapter.get()` method

app/src/main/java/org/fossasia/openevent/adapters/SessionsListAdapter.java
```diff
@@ -94,7 +99,7 @@ public void moveItem(int fromPosition, int toPosition) {
         notifyItemMoved(fromPosition, toPosition);
     }
 
-    class Viewholder extends RecyclerView.ViewHolder implements View.OnClickListener {
+    public class Viewholder extends RecyclerView.ViewHolder implements View.OnClickListener {
```
creativepsyco: This class should be declared `static` to prevent adapter leaks.

app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
```diff
@@ -113,7 +117,7 @@ public void moveItem(int fromPosition, int toPosition) {
         notifyItemMoved(fromPosition, toPosition);
     }
 
-    class ViewHolder extends RecyclerView.ViewHolder {
+    public class ViewHolder extends RecyclerView.ViewHolder implements View.OnClickListener{
```
creativepsyco: Declare this static.

app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
```diff
@@ -36,7 +38,9 @@ public SpeakersListAdapter(List<Speaker> speakers) {
         View view = layoutInflater.inflate(R.layout.item_speaker, parent, false);
         return new ViewHolder(view);
     }
-
+    public static interface AdapterOnClickHandler {
+        void onClick(String name, ViewHolder vh);
```
creativepsyco: Use position instead of string

app/src/main/java/org/fossasia/openevent/adapters/LocationsListAdapter.java
```diff
@@ -105,18 +112,26 @@ public void moveItem(int fromPosition, int toPosition) {
...
             floor = (TextView) itemView.findViewById(R.id.location_floor);
 
         }
 
+        @Override
+        public void onClick(View view) {
+            int adapterPosition = getAdapterPosition();
+            Microlocation current = microlocations.get(adapterPosition);
+            String name = current.getName();
+            mClickHandler.onClick(name, this);
```
creativepsyco: should you allow viewholder to access the member variables of the outer class? This is a common trajectory seen in memory leaks.

opticod: As suggested by u in lower comments, i will make the class static and hence mClickHandler static also this will solve  adapter leaks...Any hint for preventing memory leak here?

I am accessing member variable to handle the onClick events.

app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
opticod: Ok i will implement it using position.

app/src/main/java/org/fossasia/openevent/adapters/LocationsListAdapter.java
```diff
@@ -105,18 +107,27 @@ public void moveItem(int fromPosition, int toPosition) {
...
-
+            itemView.setOnClickListener(this);
             name = (TextView) itemView.findViewById(R.id.location_name);
             floor = (TextView) itemView.findViewById(R.id.location_floor);
 
         }
 
+        @Override
+        public void onClick(View view) {
+            clickHandler.onClick(getAdapterPosition(), this);
```
opticod: done with the other suggestions,Thank You.
One thing is that ,I need to access member variable to handle the onClick events.

creativepsyco: You can pass the click listener onto the view holder during `onBindViewHolder` method.
```java
onBindViewHolder(...) {
viewHolder.itemView.setClickListener(adapterClickListener); // where this is the adapter.
}
```

opticod: Isn't having click listener in `onBindViewHolder` method is an expensive one with respect to performance as there will be too many instances of clickListener ?

http://stackoverflow.com/questions/33845846/why-is-adding-an-onclicklistener-inside-onbindviewholder-of-a-recyclerview-adapt

creativepsyco: Yes that is true but we could pass the click listener through to the viewholder instead of accessing a member field from the viewholder which is a static class.

mananwason: Something like [this](https://github.com/fossasia/open-event-android/blob/master/app/src/main/java/org/fossasia/openevent/adapters/SponsorsListAdapter.java) @opticod

creativepsyco: we don't want to make viewholder an inner class because it leaks the adapter, the adapter leaks the view, the view leaks the context and thats it.

mananwason: What if we make the viewholder static. Would it leak the adapter then also?

creativepsyco: no making it static does not allow it to have inner references so it is always good practice to use static inner classes unless you absolute need the outer reference. Please see: http://www.androiddesignpatterns.com/2013/01/inner-class-handler-memory-leak.html

Its for Handlers but applies generally to any Android code that has lifecycle issues like adapter, views, activities etc.

mananwason: Oh Great !! So useful  :smile:

opticod: Can I Implement in this way? I haven't used any explicit reference from outer class in inner class 
```
@Override
public void onBindViewHolder(Viewholder holder, int position) {
    Session current = sessions.get(position);
    String title = current.getTitle();
    String summary = current.getSummary();
    holder.sessionName.setText(title);
    holder.sessionSummary.setText(summary);
    holder.itemView.setOnClickListener(new myOnClickListener(position,clickHandler));
}
```

```
private static class myOnClickListener implements View.OnClickListener {
        private int position;
        private AdapterOnClickHandler clickHandler;
        myOnClickListener(int pos,AdapterOnClickHandler ch){
            position=pos;
            clickHandler=ch;
        }
        @Override
        public void onClick(View v) {
            clickHandler.onClick(position);
        }
    }
```

```
public static class Viewholder extends RecyclerView.ViewHolder implements View.OnClickListener {
        TextView sessionName;
        TextView sessionSummary;

        public Viewholder(View itemView) {
            super(itemView);
            itemView.setClickable(true);
            itemView.setOnClickListener(this);
            sessionName = (TextView) itemView.findViewById(R.id.session_title);
            sessionSummary = (TextView) itemView.findViewById(R.id.session_abstract);

        }

        @Override
        public void onClick(View view) {

        }
    }
```

creativepsyco: so you will need to do 2 things:
* in the outer adapter you already cache the listener
* set the listener to the item view like `viewHolder.itemView.setListener(cachedListener)`
* Remove the `setOnClickListener` from viewholder constructor

That should be good.

creativepsyco: Also I see that you may need access to the viewHolder  properties so you can also cache this listener as `final` in the ViewHolder class. So to re-iterate:
* Outer adapter caches the listener
* pass the listener in the viewHolder, either by setting it explicitly or attaching it like `viewHolder.setItemClickListener(cachedListener)`
* In the viewHolder cache this and then set the listener on the view. like `itemView.setListener(this); this.cachedListener = cachedListener;`
* The invoke the cachedListener during `onClick()` like: `this.cachedListener(getAdapterPosition(), view)`

The idea is something similar that you have already done except we don't create any new instances anywhere just pass along memory references to an already existing click listener.

This is important because a recycler view can contain items of different types which means the viewHolders will be different implementations. So if that happens you need to maintain these different implementations in separate java files instead of the Adapter and therefore it is good practice to have a structure early on. Since your viewHolder class is static you can use it in other adapters too! Facilitates reusability and tones down complexity and coupling between the components.

So if you are stuck about thinking how you can structure the code, think about this: How can I make sure this viewHolder can be used by a different adapter, without creating any new clickListeners anywhere else in the code.

opticod:
@creativepsyco  done with the changes.

opticod:
travis is failing due to codestyles...it also includes code that are not changed in this commit.

creativepsyco:
The codestyle issues will only be posted if you have made them in the current commit, i wouldn't worry about those.

opticod:
@creativepsyco ,I have manually fixed all that i have made in current commit...but still its failing. :(

creativepsyco:
okay let me take a look into it, i think some gem is missing.

opticod:
solved the merge conflict.

mananwason:
Build failing @opticod. Check the travis logs.

opticod:
@mananwason @creativepsyco . Build passed

opticod:
Am I missing anything ? @creativepsyco

creativepsyco:
there are some conflicts lemme resolve and merge.

opticod:
Ok,thanks.

creativepsyco:
merge via commit 9252760

PR #175, opened 2016-01-28
FIXED #174

### LINKED ISSUE #174: Missing Updates of Bookmarked List
On UnBookmarking a session ,it doesn't get removed from the Bookmarked List , until and unless we change the fragment.
gif_2

opticod:
#174

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/fragments/BookmarksFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/fragments/BookmarksFragment.java
@@ -34,6 +34,25 @@ public class BookmarksFragment extends Fragment {
     ArrayList<Session> bookmarkedSessions = new ArrayList<>();
     ArrayList<Integer> bookmarkedIds;
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        if (sessionsListAdapter != null) {
+            try {
+                DbSingleton dbSingleton = DbSingleton.getInstance();
+                bookmarkedIds = dbSingleton.getBookmarkIds();
+                bookmarkedSessions.clear();
+                for (Integer id : bookmarkedIds) {
+                    Session session = dbSingleton.getSessionById(id);
+                    bookmarkedSessions.add(session);
+                }
+                sessionsListAdapter.notifyDataSetChanged();
+
+            } catch (ParseException e) {
+                e.printStackTrace();
+            }
+        }
+    }
 
     @Nullable
     @Override
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -41,9 +41,9 @@
     <string name="default_mode_api">Developer</string>
 
     <string name="settings_mode_notification"> Notification Mode</string>
-    <string name="default_mode_notification">10 mins before</string>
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/fossasia/openevent/fragments/BookmarksFragment.java
```diff
@@ -29,11 +29,17 @@
  * Created by manan on 22-05-2015.
  */
 public class BookmarksFragment extends Fragment {
-    SessionsListAdapter sessionsListAdapter;
+    static SessionsListAdapter sessionsListAdapter;
```
creativepsyco: why is the adapter variable static?

opticod: I am referencing `notifyBookmarkedChanged()` from static context of  `onOptionsItemSelected(MenuItem item)` ..so need to keep `sessionsListAdapter` static.

app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -124,6 +124,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                     createNotification();
                 }
                 sendBroadcast(new Intent(BookmarkWidgetProvider.ACTION_UPDATE));
+                BookmarksFragment.notifyBookmarkedChanged();
```
creativepsyco: why is bookmarks fragment being accessed from here?

app/src/main/java/org/fossasia/openevent/fragments/BookmarksFragment.java
creativepsyco: nope this isn't right, firstly, `SessionDetailActivity` should not access the Bookmarks fragment plus, you shouldn't need to access cached objects like this.

creativepsyco:
I think the right way to fix this is to be just refresh the `BookmarksFragment` when it is shown. You cannot depend on Bookmarks fragment being available because it is in the activity behind the current one (SessionsDetail), also trying to fix it via `static` modifiers makes it not garbage collected when it should be (Static members are assigned on the class address space, not on the object address space) See http://stackoverflow.com/questions/453023/are-static-fields-open-for-garbage-collection

opticod:
@creativepsyco done.
Thanks for pointing out..this all helps a lot and to improve while contributing.. :smiley:

app/src/main/java/org/fossasia/openevent/adapters/SessionsListAdapter.java
```diff
@@ -94,6 +98,24 @@ public void moveItem(int fromPosition, int toPosition) {
         notifyItemMoved(fromPosition, toPosition);
     }
 
+    public void refresh() {
```
creativepsyco: SessionsListAdapter has no knowledge of the DbSingleton now here you are adding some more coupling.

opticod: Thanks, then simply we can have it in Bookmarks fragment only....and after refreshing the list, we will call `adapter.notifyDataSetChanged()`

Issue #182, opened 2016-02-03
Clicking on speaker social profile with no link results in app crash


suheb:
Steps to reproduce:
1. Go to speakers section.
2. Click on any speaker.
3. Click on any social profile icon.
4. If the selected profile's link not available, it results in app crash.

suheb:
PR to fix this #183

vjs3:
@suheb I tried to reproduce it several times but didn't able to reproduce the bug.

PR #183, opened 2016-02-03
Fixes: Clicking on speaker social profile with no link results in app crash #182

suheb:
Also, showing a Toast instead of hiding the ImageView, which I think provides better UI experience.
Signed-off-by: Suhaib Khan <suheb.work@gmail.com>

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
@@ -3,9 +3,14 @@ package org.fossasia.openevent.activities;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
-import android.support.v7.widget.*;
+import android.support.v7.widget.DefaultItemAnimator;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.SearchView;
+import android.support.v7.widget.Toolbar;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.view.View;
 import android.widget.ImageView;
 import android.widget.TextView;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/openevent/utils/SpeakerIntent.java
```diff
@@ -41,9 +41,9 @@ public void onClick(View view) {
                     Toast.makeText(view.getContext(), "What did you click?? :P", Toast.LENGTH_SHORT).show();
                 }
 
-                if (url.equals(null)) {
+                if (url.equals(null) || url.isEmpty()) {
```
creativepsyco: how can you compare url to null? are u sure this fixes it? It's incorrect.

suheb: Didn't notice it. The crash was due to the `url` being empty string so this fixes it.
But this is obviously wrong way to compare `null`.

mananwason: @suheb You just need to check for null like `url != null` and it will work. No need for do this.

suheb: @mananwason We do need check if `url` is a empty string (i.e "") even if it's not `null`.

mananwason: @suheb Makes sense.

app/src/main/java/org/fossasia/openevent/utils/SpeakerIntent.java
```diff
@@ -40,10 +40,10 @@ public void onClick(View view) {
                     url = "";
                     Toast.makeText(view.getContext(), "What did you click?? :P", Toast.LENGTH_SHORT).show();
                 }
-
-                if (url.equals(null)) {
+                
+                if (url == null || url.isEmpty()) {
                     Log.d("url null", "ABC");
-                    view.setVisibility(View.GONE);
+                    Toast.makeText(view.getContext(), "Link not available.", Toast.LENGTH_SHORT).show();
```
mananwason: Move string to strings.xml

mananwason:
@suheb If you can remove the views that don't exist instead of showing a toast, that would be nice.

suheb:
@mananwason I wanted to discuss that too. The existing behavior was not correct as it would just hide the view on click and it comes back up again when the activity is restarted.
Then I figured perhaps we wanted same layout for all speakers so I just showed a toast.
If we want to hide the views with no link, we can do it when inflating the layout.
What should it be?

mananwason:
We want to hide the view when it is being created by checking if the url is null or empty. Okay?

opticod:
we can do it in adapter . r8t ? @mananwason

mananwason:
Exactly @opticod we can do this in the oncreateviewholder of the adapter

suheb:
@opticod @mananwason How are we gonna do this in adapter?
The speaker layout is inflated in the SpeakerActivity class. We can just do it there only.

mananwason:
Yes @suheb My bad. I thought the icons were in a recylerview. Since, we have to manipulate icons from the activity, we will have to do it in the activity only.

suheb:
@mananwason Yeah:) Please review my latest commit.

mananwason:
Looks good. Please squash commits and then we can merge!

app/src/main/java/org/fossasia/openevent/activities/SpeakersActivity.java
```diff
@@ -59,13 +64,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        else {
+            speakerIntent.clickedImage(twitter);
+        }
+        if(selectedSpeaker.getGithub() == null || selectedSpeaker.getGithub().isEmpty()) {
+            github.setVisibility(View.GONE);
+        }
+        else {
+            speakerIntent.clickedImage(github);
+        }
+        if(selectedSpeaker.getFacebook() == null || selectedSpeaker.getFacebook().isEmpty()) {
```
creativepsyco: a better way is `TextUtils.isEmpty(string..)`

PR #192, opened 2016-02-05
Fixed #191

### LINKED ISSUE #191: About Us dialog comes up when Settings is pressed
Also, when you click on Settings in that nav bar, it doesn't directly lead you to SettingsActivity.class. First the animation occurs and then you see the same screen for a while, and then the intent acts.

It's trivial to fix. Will drop in a PR in a bit.

shivenmian:
Also, removed extra blank lines and redundant comments. @creativepsyco

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -281,6 +281,7 @@ public class MainActivity extends BaseActivity {
                         overridePendingTransition(R.anim.slide_in_left, R.anim.slide_out_left);
                     }
                 });
+                break;
             case R.id.nav_about:
                 AlertDialog.Builder builder =
                         new AlertDialog.Builder(this);
@@ -288,13 +289,10 @@ public class MainActivity extends BaseActivity {
                 builder.setMessage(getResources().getText(R.string.about_text));
                 builder.setPositiveButton("OK", null);
                 builder.setIcon(R.mipmap.ic_launcher);
-
                 AlertDialog welcomeAlert = builder.create();
                 welcomeAlert.show();
-                // Make the textview clickable. Must be called after show()
                 ((TextView) welcomeAlert.findViewById(android.R.id.message)).setMovementMethod(LinkMovementMethod.getInstance());
                 break;
-
         }
         currentMenuItemId = menuItemId;
         mDrawerLayout.closeDrawers();


app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -277,10 +277,11 @@ private void doMenuAction(int menuItemId) {
                 smoothActionBarToggle.runWhenIdle(new Runnable() {
                     @Override
                     public void run() {
-                        startActivity(intent);
                         overridePendingTransition(R.anim.slide_in_left, R.anim.slide_out_left);
                     }
                 });
+                startActivity(intent);
```
creativepsyco: do you even know what you are doing here?

shivenmian: Yes. The runnable executes the transition. If the intent is called within the runnable then the activity will come after the transition is up. 

You can try it out with the other options on the menu. When you press the button then the page immediately comes up, but as of now if you press Settings, there's a noticeable lag.

No problem though, I can remove it.

shivenmian: Yes. The runnable executes the transition. If the intent is called within the runnable then the activity will come after the transition is up.

You can try it out with the other options on the menu. When you press the button then the page immediately comes up, but as of now if you press Settings, there's a noticeable lag, and it first goes back to where you were and then the SettingsActivity comes . When I started the intent outside the runnable, it worked fine as I wanted. You can check or I can show you a GIF.

No problem though, I can remove the intent and put it back where it was. But the break statement had to be there IMO. @creativepsyco

shivenmian:
@creativepsyco thoughts?

app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
creativepsyco: the reason is that the settings activity is a different activity altogether, its not a fragment, the rest of the options are all fragment. No need to show me a gif, I know about it. when you call activity#onCreate the previous must be destroyed(or paused) before switching. And since the drawer is attached to it therefore we need to wait for the animation to finish. 

Yes break statement is needed, that's all. I don't think you need to do anything more than that.

shivenmian:
@creativepsyco Is this fine now? I've removed the redundant comment and extra blank lines too.

PR #193, opened 2016-02-06
fixed issue #191 and fixed view of navigation bar

mukesh14149:
1. "break" is missing in method(doMenuAction) of MainActivity.java file after case of nav_settings  
2. Navigation bar  is not looking good as "about us" is goes down 

Earlier it was looking like foss
Now it is 
foss2


shivenmian:
@mukesh14149 @creativepsyco I've done the same thing in my own PR #192

creativepsyco:
Invalid PR.

PR #199, opened 2016-02-16
Added Some Missing Refresh Options

opticod:
The search results was not refreshing on clearing all the texts. Fixed


opticod:
If we type something in SpeakerFragment or TracksActivity and then click cross symbol in search view,then it is not refreshing to its default list.

creativepsyco:
cool :+1:

Issue #204, opened 2016-02-28
Saving Navigation Drawer Image


scopeInfinity:
When connected to Internet
![aqua_power_hd_2016-02-29-02-59-22](https://cloud.githubusercontent.com/assets/9819066/13382035/ecafaa6c-de90-11e5-9a9c-9527afd775c4.png)

After disconnecting from Internet and starting Application again
![aqua_power_hd_2016-02-29-02-59-45](https://cloud.githubusercontent.com/assets/9819066/13382037/ef37d71e-de90-11e5-8edc-779af9281190.png)

We can temporarily store image that can be used in absence of Internet connectivity.

mananwason:
Closing this

PR #205, opened 2016-02-28
Saving Navigation Drawer Image

scopeInfinity:
Fixed Issue  #204


mananwason:
@scopeInfinity We don't require this. Picasso is already caching the image for us. If you want you can try to implement better caching scheme but I am not sure if that is also required here. I am closing this for now.

PR #220, opened 2016-03-06
LocationActivity name is wrong

shivenmian:
I thought this should be a part of a separate commit, so sent a PR for this separately. It's a typo but still.


shivenmian:
@championswimmer Please look at this.

Issue #242, opened 2016-03-09
Missing Runtime Permissions for APK >=23 


opticod:
For missing permissions , the view remains un-populated.
![screenshot_20160309-200510](https://cloud.githubusercontent.com/assets/13851773/13638582/efdf90c2-e632-11e5-8076-97eb69aaba4f.png)

opticod:
we can close this issue.

PR #243, opened 2016-03-09
Added missing Runtime Permission for  API >=23

opticod:
#242 Permissions Added , and view gets populated after giving permissions
screenshot_20160309-200835
screenshot_20160309-200850

### DIFF
--- a/app/src/fdroid/java/org/fossasia/openevent/fragments/OSMapFragment.java
+++ b/app/src/fdroid/java/org/fossasia/openevent/fragments/OSMapFragment.java
@@ -1,10 +1,16 @@
 package org.fossasia.openevent.fragments;
 
+import android.Manifest;
+import android.app.Activity;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.widget.Snackbar;
 import android.support.v4.app.Fragment;
+import android.support.v4.content.ContextCompat;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


app/src/fdroid/java/org/fossasia/openevent/fragments/OSMapFragment.java
```diff
@@ -42,6 +50,23 @@ public void onCreate(Bundle savedInstanceState) {
...
 
         View rootView = inflater.inflate(R.layout.fragment_map, null);
+        Activity activity = getActivity();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            if (ContextCompat.checkSelfPermission(activity,
+                    Manifest.permission.WRITE_EXTERNAL_STORAGE)
+                    != PackageManager.PERMISSION_GRANTED) {
+                ActivityCompat.requestPermissions(activity,
+                        new String[]{Manifest.permission.WRITE_EXTERNAL_STORAGE},
+                        PERMISSION_REQUEST_WRITE_EXTERNAL_STORAGE_RESULT);
```
opticod: here , `BitmapFactory` is trying to fetch a file repeatedly which is located in device storage. As soon we provide the permissions, it gets the access and view gets populated after resume of activity.
So,without overriding the `onRequestPermissionsResult()` we can achieve this.

opticod: if we deny, the view does't gets populated . If we again open `Map` , it will again ask for the permission.

opticod: ezgif com-video-to-gif 4

mananwason: This is not exactly correct. Why don't you use the callback provided? You are checking for permission again after requesting it, which would be more resource intensive than just getting the result from the callback.

opticod:
@championswimmer @mananwason , I have made changes, Please have a look.

PR #266, opened 2016-03-11
LinkedIn app support added

abhishekjiitr:
Now, when the user clicks on speaker's linked in profile icon it opens directly in linkedin app, rather than in the browser.


championswimmer:
What happens when user does not have Linkedin app ?

the-dagger:
@championswimmer It asks the user to open it inside a browser of his choice.
I have tested it and it works perfectly.

PR #270, opened 2016-03-12
Converted to Timber

### LINKED ISSUE #269: Convert Log.* calls to Timber.* calls
Now that we use Timber - let's actually make use of it :D

shivenmian:
Fixes #269.

I've added Timber.tag for those which have tags explicitly mentioned. And I removed the tags for some features which didn't need tags and could have tag as Activity name.

@championswimmer Please tell me if there are issues in this. I'll resolve them and squash commits.


championswimmer:
LGTM

PR #274, opened 2016-03-12
Fixed redundant NavDrawer selection

shivenmian:
When you go to Settings from NavDrawer, or About Us, and then open NavDrawer again, both of the items remain selected.

There is no real need of implementing something which checks only one of those, because clicking on any of those two leads you to a separate window from where NavDrawer can't be accessed. When you go to Settings, then to go to NavDrawer, you'll have to go back, so Settings will obviously have to be unchecked then. Same with About Us (you'll have to click OK and dismiss the dialog to see NavDrawer, but About Us needs to be unchecked).

There is thus no real need of making any of these two checked (i.e setChecked). We anyway can't see them checked when we are at the Settings page / About Us dialog. 

@championswimmer @mananwason Have a look.


shivenmian:
@championswimmer Have a look.

championswimmer:
Makes sense.  Thanks.  Will test and merge

shivenmian:
@championswimmer Please do.

championswimmer:
Please do not make multiple comments when not required. It increases the
time people have to spend clearing their inboxes (each comment also becomes
an email)

Be considerate that everyone is in different time zones. Some are
travelling to Singapore for FOSSASIA16 and everyone has their pace of
working on things.

Once your code is reviewed, please wait till it's get merged for at least
2-3 days.

shivenmian:
@championswimmer Could you have a look now if possible?

shivenmian:
@championswimmer Can you have a look now?

Issue #285, opened 2016-03-15
Logo changes to Twitter logo when offline.


SilentFlame:
This happens I'm offline, I know that it's not a major bug but thought to report as such we can improve it.

![shot](https://cloud.githubusercontent.com/assets/9693795/13779362/91391e82-eae0-11e5-931d-dbe2e068f782.png)

@mananwason I guess this will be easy to resolve, as we need to save the logo to the app only.
I could help but am really not familiar with android a bit. 
Let me know if I can help in something that need not use Android application as such.

the-dagger:
Wait, i'll choose a better image and send a PR

the-dagger:
@championswimmer I have fixed this in #288

mananwason:
@the-dagger I didn't want to increase the apk size by adding more resources. Plus the main aim of the system is to be Event independent and if we add such logos, then it defeats the purpose.

PR #288, opened 2016-03-15
Use a better nav_drawer image when the app is offline

the-dagger:
Fix for #285 
When the app is offline, the nav_header image is set as twitter logo.
I modified it to use the fossasia logo from the splash screen.
@mananwason @championswimmer 
alt tag


mananwason:
@the-dagger I didn't want to increase the apk size by adding more resources. Plus the main aim of the system is to be __Event independent__ and if we add such logos, then it defeats the purpose.

the-dagger:
@mananwason Understood!
Should we set the default image to be a solid material color?

mananwason:
@the-dagger I would prefer just removing the twitter image and let it be same as the drawer color. If we set any particular material color, then also it amy be different from what the conference theme is.

the-dagger:
Aight then. (y)

PR #291, opened 2016-03-17
Plugin Error fixed

shivenmian:
The gradle version is outdated, gives an outdated version error. Fixed. @championswimmer 

screen shot 2016-03-17 at 1 41 07 pm


shivenmian:
@championswimmer Could you have a look?

shivenmian:
@championswimmer Please have a look

PR #293, opened 2016-03-21
Add try and catch block

baguzzzaji:
Add try and catch block to prevent crash from NullPointerException while calling getLogo() in MainActivity.java


mananwason:
I am not sure if this bug actually arises, If you can reproduce it and post the stack trace here, we can merge this. Closing this for now.

PR #303, opened 2016-03-22
Fixed Travis + Bumping versions

opticod:


app/build.gradle
```diff
@@ -23,7 +23,7 @@ android {
         minSdkVersion 15
         targetSdkVersion 23
         versionCode 80
-        versionName "0.8.0"
+        versionName "0.9.0"
```
championswimmer: This is not be clubbed with any other change. 
Version bumps are NOT dependent on libs, but on the features we add to the app.

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta2'
+        classpath 'com.android.tools.build:gradle:1.5.0'
```
championswimmer: This will cause other conflicts, by going version down.

championswimmer: This will cause other conflicts, by going version down.

championswimmer:
Done by this https://github.com/fossasia/open-event-android/pull/302 anyway

app/build.gradle
championswimmer: Also we cannot go to 0.9.0 untill we solve all these 
https://github.com/fossasia/open-event-android/milestones/v0.9.0

opticod: Ok.

PR #307, opened 2016-03-23
Fixed multiple selection in NavDrawer

shivenmian:
There's no need of adding any lines, removed a redundant line which solves this. Works fine. @championswimmer @mananwason


shivenmian:
@championswimmer Please have a look.

mananwason:
Merging this.

PR #308, opened 2016-03-23
Removed the empty space in navigation drawer

the-dagger:
@championswimmer 
Before
navbef

After
navfixed


mananwason:
Looks good. Merging it.

PR #311, opened 2016-03-23
Fixed unnecessary NetworkDialog

### LINKED ISSUE #310: Network Dialog stays there if WiFi turned on from pulldown menu
When user starts app with WiFi off, then the "Turn WiFi on" dialog comes up. But if the user turns on the WiFi from the pull down menu instead of going to Settings, then the download starts, but the "Turn WiFi on" dialog stays there.

shivenmian:
Fixes #310 @championswimmer @mananwason


shivenmian:
@championswimmer Please have a look.

mananwason:
I am not sure, if what you've done solves the issue but the ideal way to fix this is to listen for CONNECTIVITY_CHANGE broadcast. [Source ] (http://developer.android.com/training/monitoring-device-state/connectivity-monitoring.html)

shivenmian:
@mananwason This solves the issue. This is anyway logical. Just remove the dialog if it exists on the screen during DataDownload.

Or do you want me to do the other one? That's trivial too.

mananwason:
@shivenmian I get what you're doing but it doesn't seem right, would be best to change remove dialog when connectivity changes.

shivenmian:
I'll update my PR in a while.

mananwason:
ping me here when you're done, I'll review it.

PR #326, opened 2016-04-04
Update gradle plugin version from 2.0.0-beta6 to 2.1.0-alpha5

suheb:
Signed-off-by: Suhaib Khan <suheb.work@gmail.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta6'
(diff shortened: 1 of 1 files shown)


the-dagger:
@suheb I don't know, but should we force everyone to use the alpha plugin?
IMO we should hold back until the stable version is out.
Cheers.

championswimmer:
Let's not use alpha as long as things build with the existing version.

opticod:
Today gradle stable plugin version 2.0.0 is released , can be switch to stable version?

mananwason:
Yes @opticod. Changed it in #331. Closing this @suheb @championswimmer

PR #357, opened 2016-04-17
Fixed Redundant empty Bookmarks Page

### LINKED ISSUE #338: Redundant empty bookmarks screen
When there are no bookmarks, the Bookmarks screen just shows "No bookmarked session yet !!". 

It's a really bad idea to do this. You're letting the user press the Bookmarks option on the NavigationDrawer, and lead them to nothing, which is awful. And it's wasting time.

We don't need to lead the user to an entire blank screen with nothing on it to tell him that no bookmarks exist.

shivenmian:
Fixes #338.

Added a simple check once the user presses the Bookmarks option on NavDrawer. If no bookmarks exist, it shows up a prompt rather than lead him to an empty screen, thus saving time of the user. So the BookmarksFragment gets loaded only when some sessions are bookmarked.

Also, if we un-bookmark a session and the Bookmarks list becomes empty again, it shows up the dialog on pressing the back button on that Session Detail, and clicking on OK loads the TracksFragment back again.

This also happens to be one of the layout changes suggested in [my proposal](https://docs.google.com/document/d/1P1-UE0PFEE6KAnYhrdkRBmqzgYoUnN1ZFWQeC_p_mh4/edit), to reduce the blankiness in the application.

@championswimmer @mananwason @juslee

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -255,9 +255,14 @@ public class MainActivity extends BaseActivity {
                 getSupportActionBar().setTitle(R.string.menu_tracks);
                 break;
             case R.id.nav_bookmarks:
-                fragmentManager.beginTransaction()
-                        .replace(R.id.content_frame, new BookmarksFragment(), FRAGMENT_TAG).commit();
-                getSupportActionBar().setTitle(R.string.menu_bookmarks);
+                DbSingleton dbSingleton = DbSingleton.getInstance();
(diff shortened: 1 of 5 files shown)


championswimmer:
@shivenmian Screenshots please

shivenmian:
@the-dagger Had deleted my clone because I'm on a different project rn, but you can merge it. Works. Instead of showing the present "No bookmarks yet" screen, it just shows a dialog if no bookmarks there (saving the user from a blank screen which does nothing). You can pull it and check.

the-dagger:
Cool. 👍

PR #378, opened 2016-06-02
Added json config file parsing

mananwason:
This Pull request adds the functionality to the app to read a config file and extracts the name, api url from there. @mariobehling @the-dagger Please review.

### DIFF
--- a/android/app/src/main/assets/config.json
+++ b/android/app/src/main/assets/config.json
@@ -1,0 +1,5 @@
+{
+  "Email": "dev@fossasia.org",
+  "App_Name": "Open Event",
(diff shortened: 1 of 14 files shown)


the-dagger:
@mananwason Here we are assuming that the endpoints in the API provided by the organizer are the same as the endpoints in the orga-server right?

mananwason:
@the-dagger We will provide users with two options : 
* To upload their own json files
* To use the open event orga server to build an api

So right now in this PR I'm targeting the second feature. this will allow us to use any api which has similar endpoints as open event orga server, we can target uploading json files later.

the-dagger:
Looks good to me for merge. :+1:

Issue #385, opened 2016-06-06
App Generator Component 5: Send apk to organiser once build has successfully completed


mananwason:
Once the apk has been build successfully and signed, we need to send the apk to the organiser via mail

the-dagger:
Closing @mananwason

PR #389, opened 2016-06-07
Script to automatically send email by fetching address from a txt file

the-dagger:
Script that fetches the email address from a .txt file and sends the email to it with the compiled email as an attachment. #385 
@mariobehling @mananwason Please review.

### DIFF
--- a/apk-generator/scripts/send_email.sh
+++ b/apk-generator/scripts/send_email.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+echo "Preparing to send the email"
+EMAIL=`cat email.txt`
+mutt -s "Test mail" $EMAIL -a /root/dev@fossasia.org/open-event-android/android/app/build/output/apk/app-debug.apk < /root/mail_message.txt
+echo "Mail sent"
+


mananwason:
@the-dagger We need to have the apk folder as an argument to the script. Please wait for a day before merging yourself.

the-dagger:
Oh ok. 
But why is that so? 
The location for the generated apk is going to be constant and so is the apk name. 
So why do we need to pass the folder as an argument?

mananwason:
So the thing is that we want to have as many people as there can be to generate apk's at the same time and you have hardcoded a folder that has been created by one particular string. What I am doing in the script is that I am creating a new folder with the name same as that of the email of the organiser entered on the webpage, now according to your script, it will always get the same apk which was meant for the developer with the email address dev@fossasia.org, so it' not right, please rectify it, then I'll integrate it in the python script I made

the-dagger:
@mananwason Ok, got it. :+1: 
Something like this looks good? [http://hastebin.com/owaquyuvip.bash](http://hastebin.com/owaquyuvip.bash)

mananwason:
Looks fine @the-dagger

Issue #380, opened 2016-06-06
App Generator Component 4: Implement function to call or trigger the script to build and sign the apk after upload of JSON files


the-dagger:
Trigger the build script in #383  whenever the user submits his/her information in the app-generator website.

mariobehling:
Please reference this issue when you make PRs. If this is finished please tell us what technology you used and close the issue when done.

the-dagger:
:+1:

the-dagger:
@mananwason @mariobehling I am facing a problem in executing a particular .sh file through the php script.
Seems like a permission issue so I will continue to research on this

the-dagger:
Turns out that we need to use Node.js for executing the python and bash scripts rather than PHP.
So will work on shifting the server from LAMP to MEAN.

mariobehling:
@the-dagger Please check what the Web App folks are using @championswimmer Could you give feedback, please?

the-dagger:
@mariobehling yes I had a talk with arnav, they are using node.js for it. 
I'll use the same and try to fix it by today.

mariobehling:
Where did that conversation happen? Which issue? I did not see it on the gitter channel?

the-dagger:
@mariobehling It was not a full discussion, I just personal messaged him on gitter asking if using Node.js was better and he said that it was.
I didn't have a detailed discussion with him regarding this.

PR #393, opened 2016-06-09
Update index.html

the-dagger:
Update index.html to 
 - Save data according to user's timestamp
 - Launch a php script to generate the apk when submit button was pressed.
 - Pass the timestamp variable to the php script as an argument 
Fixes issue #380 
@mananwason @mariobehling Please have a look


the-dagger:
No need for this, as the changes are included in #394

PR #419, opened 2016-06-20
Schedule Completed

mananwason:
This commit adds the schedule functionality but this adds a lot of overhead on the app since the data has to be sorted on the app. I have to test this for bugs due to the extra overhead, if the overhead becomes too much, we would have to move the filtering part on the server


mananwason:
@the-dagger  Please review. I'll fix the codacy issues

the-dagger:
Looks good.
Should I merge, or you'll be fixing the codacy issues?

mananwason:
I'll see if the issues are worth fixing and thenlet you know

android/app/src/main/java/org/fossasia/openevent/data/Session.java
```diff
@@ -46,6 +46,8 @@
 
     int track;
 
+    String startDate;
```
mananwason: This is not required

mananwason:
@the-dagger I have fixed the codacy issues that required fixing. So, this PR is good to go now.

@mariobehling @the-dagger [This] ( https://drive.google.com/file/d/0ByZsjX7yMIzzOTBPTmpNaThfeVU/view?usp=sharing) is a screencast of how the schedule is looking like currently. Please review and merge if it looks good.

the-dagger:
LGTM

PR #427, opened 2016-06-21
fixes #310

### LINKED ISSUE #310: Network Dialog stays there if WiFi turned on from pulldown menu

mananwason:


the-dagger:
Looks good, merging.

PR #438, opened 2016-06-24
Add script for uploading files

### LINKED ISSUE #434: Adopt file import to import zip with all JSON files
In the Orga-Server we changed the JSON import/export into a zip file with all files inside. Please implement this for the app generator too. So users can import a the zipped JSON files instead of single files.

the-dagger:
@mananwason please review and merge.
Fixes #434


the-dagger:
@mananwason Please have a look

PR #445, opened 2016-06-28
fixes #439

### LINKED ISSUE #439: App crashes when a bad API URL is entered
When I add an arbitary wrong url in the app generator like /api/v1.com, the app crashes. This error should be handled graciously.

mananwason:

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/api/Urls.java
+++ b/android/app/src/main/java/org/fossasia/openevent/api/Urls.java
@@ -1,5 +1,7 @@
 package org.fossasia.openevent.api;
 
+import android.webkit.URLUtil;
+
 /**
  * User: championswimmer
  * Date: 23/5/15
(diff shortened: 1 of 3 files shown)


android/app/src/main/java/org/fossasia/openevent/api/Urls.java
```diff
@@ -46,6 +50,12 @@ public static String getBaseUrl() {
     }
 
     public static void setBaseUrl(String baseUrl) {
-        BASE_URL = baseUrl;
+        if (URLUtil.isValidUrl(baseUrl)) {
```
the-dagger: So basically anything that is a valid URL will be qualified as good to go here tight?
So what if I use www.google.com here?
It is a valid URL but is not a valid API.
@mananwason

mananwason: Yes @the-dagger, you are correct. We can never know if an api is correct without actually making a request. This PR focuses on invalid api urls which was making the app crash

the-dagger: Ok cool, got it.

PR #447, opened 2016-07-01
Modify parser for sponsors to be fetched according to new json outputs

the-dagger:
@mananwason It is working, please check and merger.
P.S. only issue is the "htps://" being prepended at the beginning of each URL.
So please have a look at that.


the-dagger:
Moving this to #448

PR #448, opened 2016-07-01
[Do Not Merge]Modify parser for speakers to be fetched according to new json outputs

the-dagger:
@mananwason @mariobehling please have a look.
P.S. We still have to edit the SpeakerDetails Activity to display all these details that are being fetched from the new API, so I'll create a separate issue for it.


the-dagger:
screenshot-area-2016-07-01-170316
What this looks like

the-dagger:
Moved to #453

Issue #436, opened 2016-06-24
Extract json files and package them in the apk


mananwason:
Once the user uploads the zip of json files, we need to actually unzip those files and add them to the app folder to package them with the apk.

PR #484, opened 2016-07-15
added packaging of icon and json assets

### LINKED ISSUE #434: Adopt file import to import zip with all JSON files

mananwason:
@the-dagger Please review and merge. Fixes #434 #436.


the-dagger:
@mananwason could you change the base to development?

mananwason:
So that won't be possible @the-dagger because they have a different history. What I'll do is that I'll move this commit to the dev branch and create a new PR from that.

Issue #443, opened 2016-06-27
Prompt the user when the app is started for the first time if he wants the assets download to take place or not


the-dagger:
* [x] Prompt user to download resources at startup otherwise read from offline json
* [ ] Implement a preference setting where in user can select if he would like to download updates or not

mariobehling:
Yes, i think this is a better solutions than what I suggested first. Just ensure the apps tries to download everything when it starts the first time.

the-dagger:
Yep, sure thing!

mananwason:
@the-dagger Changing the title. We not only have to display the dialog, we also have to change how the download the data

mananwason:
Already completed in #490

PR #490, opened 2016-07-17
added prompt for #443

mananwason:
@the-dagger Please review and merge. Solves #443. 

@mariobehling After this the user gets a prompt asking if he wants to download assets. If he says no, then everytime on startup the prompt shows up. When the users selects yes, the prompt doesn't appear the next time he opens the app. Does this look fine?
device-2016-07-18-002211


mariobehling:
Sounds good. Please implement the option to "automatically update" or "ask before updates" in the settings.

mananwason:
Okay great. Will add it to the issue and implement by today itself.

the-dagger:
Looks good, merging

PR #593, opened 2016-08-15
Fixes #562 (Added ordering of sponsors by sponsor types, Fixed images and urls)

### LINKED ISSUE #562: Sponsor Tab: Images not scaled, not ordered according, not put into categories
The sponsor tab has a number of issues. Please fix the following:
* [x] shows sponsor images same sizes
* [x] Add sponsors into categories e.g. Platinum and order them into categories
* [x] Ensure URLs work and if not do not add a URL

device-2016-08-11-140831

* [x] Add a screenshot in the issue how your fix will look like before providing a PR

mananwason:
@mariobehling @the-dagger This fixes #562. All the images are ordered according to their sponsor types. 
@mariobehling these are the two UI I have implemented for sponsors. Please suggest which one looks better. Thanks!


the-dagger:
@mananwason For some reason clicking on a sponsor resulted in Invalid URL Snackbar even when the URL was correct.
Could you check if the problem is there in this PR?

mananwason:
@the-dagger I have fixed that in this PR

mariobehling:
@mananwason Thank you. Please use the first one. Reason: It is easier to sort according to important/sponsorship.

How do you show the sponsorship level? At the moment I do not see if it is Platinum, Gold etc? It this is missing from the sample, please add it.

mananwason:
@mariobehling I'll make separate lists based on different sponsor types. I have fixed this already in the sample. 
Right now I have just sorted in descending order of sponsorship. For eg : Redhat and Github are Platinum, then Google is Diamond etc.

mananwason:
@mariobehling @the-dagger please review and merge

android/app/src/main/java/org/fossasia/openevent/fragments/SponsorsFragment.java
```diff
@@ -68,25 +71,36 @@ public void onRefresh() {
         sponsorsRecyclerView.addOnItemTouchListener(new RecyclerItemClickListener(view.getContext(), new RecyclerItemClickListener.OnItemClickListener() {
             @Override
             public void onItemClick(View itemView, int position) {
-                String sponsorUrl = dbSingleton.getSponsorList().get(position).getUrl();
-                if (!sponsorUrl.startsWith("http") || !sponsorUrl.startsWith("https")) {
+                List<Sponsor> objects = dbSingleton.getSponsorList();
+                Sponsor sponsor = objects.get(position);
+                String sponsorUrl = sponsor.getUrl();
+                if (!sponsorUrl.startsWith("http") && !sponsorUrl.startsWith("https")) {
```
the-dagger: This will cause URL not found error again

the-dagger:
@mananwason I have added a comment, please have a look.

android/app/src/main/java/org/fossasia/openevent/fragments/SponsorsFragment.java
mananwason: No @the-dagger. I have fixed it here. Earlier it was check if it doesn't start with http/https, it would append but obviously the url can't start with both. It starts with one of them, so the and.

mananwason:
@the-dagger This is perfectly okay. Please recheck and merge.

android/app/src/main/java/org/fossasia/openevent/fragments/SponsorsFragment.java
the-dagger: Oh sorry, my bad.

Issue #560, opened 2016-08-10
Going back to track activity when back pressed from sponsors activity instead of closing.


chiragw15:
1.) The app closes when pressed back from the sponsor or speakers fragment. It will be better if it goes back to the 'Tracks' activity.
2.) The picture in the navigation drawer is not complete
![screenshot_20160811-010709](https://cloud.githubusercontent.com/assets/16978820/17568553/eeed262c-5f61-11e6-978f-893c720110ab.png)

P.S : This is first time I am reporting an issue. So, correct me if I did something wrong. I would also like to fix this, if allowed.

mananwason:
Good idea @chiragw15. Would you like to work on it?

chiragw15:
Yes yes. I was eagerly waiting for someone's reply. This is the first time I am going to contribute in an open source project. I am so excited.

PR #597, opened 2016-08-16
Solving issue #560

chiragw15:
Now the app doesn't closes when back pressed from sponsors or speakers fragment. Rather goes back to the main fragment i,e tracks fragment.


the-dagger:
@chiragw15 Please rebase your fork from upstream and then send the pull request again after creating a branch from the updated fork.
Refer to this answer on how to do it.[http://stackoverflow.com/a/31836086](http://stackoverflow.com/a/31836086)
Replace master with development, as that's our default branch

PR #598, opened 2016-08-16
Solving issue #560

chiragw15:
Now the app doesn't closes when back pressed from sponsors, location,speaker etc fragment. Rather goes to the main fragment i.e tracks fragment.


the-dagger:
Looks good, thanks.

PR #601, opened 2016-08-17
Fix formatting

the-dagger:
@mananwason @mariobehling please merge.


the-dagger:
Merging this as i have to send one more PR which affects index.html and there wil be conflicts later on.

PR #621, opened 2016-08-21
 Layout width change

saket14:
Changed the layout width of speaker list from match_parent to wrap_content


mariobehling:
@saket14 Thanks a lot. We will go through this PR. Please in future submit an issue first and match the issue with a pull request. This will make it easier for everyone to keep track and discuss issues.

mananwason:
@saket14 What layout changes does this do? Could you tell me what it solves because it also breaks a lot of things

PR #624, opened 2016-08-21
Fix multiple uploads

### LINKED ISSUE #610: Allow multiple users to upload file simultaneously
Our current setup only allows one user to upload the sample zip at a single instance.
We need to fix this once our setup goes into production.

the-dagger:
Fixes #610 , for some reason the fix wasn't yet reflected in the source code.


mariobehling:
Thanks. Testing it step by step.

Issue #620, opened 2016-08-20
Retry Action for failed downloads.


gogeta95:
The snackbars for "Download failed/couldn't refresh" or similar events should contain a "retry" action so user can retry easily and get improved experience.

gogeta95:
@the-dagger, could you assign this to me? I'd like to contribute.

the-dagger:
Hey @gogeta95 you can go ahead and work on this if you like.
Only the people belonging to an organization can be assigned to an issue. :)

PR #625, opened 2016-08-21
Add retry action on download fails

gogeta95:
added retry action to all the snackbars where downloads are failing.
resolves isssue #620


android/app/src/main/java/org/fossasia/openevent/fragments/DayScheduleFragment.java
```diff
@@ -39,7 +39,7 @@
 /**
  * Created by Manan Wason on 17/06/16.
  */
-public class DayScheduleFragment extends Fragment {
+public class DayScheduleFragment extends Fragment implements SwipeRefreshLayout.OnRefreshListener {
```
mananwason: @gogeta95 You can remove this as we have only one listener and detecting calls there works fine for now. There is no overhead for this as well

mananwason:
Just one issue @gogeta95. Please remove the implemented onRefreshListener. We don't need that for now.

gogeta95:
@mananwason, in that case we should remove all the swiperefreshlayout from layouts as well?

mananwason:
@gogeta95 I mean that the current implementation is okay. Since we have only one listener, we can let it be. We don't have to implement the listener and then override methods and then check for the listener type as you have done in this PR.

android/app/src/main/java/org/fossasia/openevent/fragments/DayScheduleFragment.java
gogeta95: you mean remove the global listener and keep the the anonymous listener (which i removed) it was because so I could reuse the ` onrefresh() ` method in case of retry action.

gogeta95:
@mananwason update the PR, please check.

PR #628, opened 2016-08-22
Fixes #600, #626, #618

### LINKED ISSUE #600: Connect the app with the FOSSASIA sample on the orga server
We have been working with the github sample for development purposes but since it's almost the end of GSoC we need to get the system in production and that means connecting the app to the orga server and check if everything works perfectly.

### LINKED ISSUE #618: Add setting preference to show local and conference's local time
Users should be able to select which time schedule they want. They should be able to switch within the schedule according to their local time as well as according to the conference's local time.

### LINKED ISSUE #626: Sort sessions according to time in the schedule
Right now all the sessions are sorted according to tracks and within each track they're sorted by time. We want to sort them according to time, so that it looks like a timeline

mananwason:
@mariobehling @the-dagger. This PR 
1. Fixes #600 : Connects the app to the orga server
2. Fixes #626 : Sort sessions according to start_time
3. Fixes #618 : Adds preference to see sessions according to local or conference time zones


android/app/src/main/assets/config.json
```diff
@@ -1,5 +1,6 @@
 {
   "Email": "dev@fossasia.org",
   "App_Name": "Open Event",
-  "Api_Link": "https://raw.githubusercontent.com/fossasia/open-event/master/sample/FOSSASIA16/"
+  "Api_Link": "http://open-event-dev.herokuapp.com/api/v2/events/132/"
+  //  "Api_Link": "https://raw.githubusercontent.com/fossasia/open-event/master/sample/FOSSASIA16/"
```
the-dagger: Remove this instead of commenting

mananwason: Done @the-dagger

PR #648, opened 2016-08-24
Remove unnecessary parameters and components

geeteshk:
Proof that components are not required: 
proof
Proof that MaxPermSize is not required:
proof

@mananwason


the-dagger:
LGBT

Issue #612, opened 2016-08-19
No error shown on entering wrong email


mananwason:
@the-dagger Even if I add a random word, it doesn't show any error. It should check for a valid email and then proceed further

the-dagger:
![screenshot_20160819_195751](https://cloud.githubusercontent.com/assets/6565270/17813037/47788a52-6647-11e6-9009-d8c27e3bec4d.png)
@mananwason Works for me.
Which browser are you using?
I tried with Firefox and Chrome

mananwason:
@the-dagger Safari :/

mananwason:
<img width="1437" alt="screen shot 2016-08-19 at 10 03 01 pm" src="https://cloud.githubusercontent.com/assets/10570249/17816810/bded3b0e-6658-11e6-8946-27ed4e04e82e.png">
Here is the screenshot

Abhi2424shek:
Should I work on it,
I could provide javascript validation soon after you click the generate app. Searching for @ and . throughout the field and displaying an alert if nothing is there. Since it is javascript it should work for all browsers ranging from old ones. And since the code is small there should not be much of a delay too.

the-dagger:
@Abhi2424shek Yep that's what I planned to do.
You can go ahead and work on this issue if you want :)

Abhi2424shek:
sure @the-dagger

Abhi2424shek:
Fixed it @the-dagger see the PR

Abhi2424shek:
Please close this issue, it is solved and the PR is merged @mananwason @the-dagger . Happy coding :)

PR #663, opened 2016-08-26
Added checking email input to resolve issue #612

Abhi2424shek:
Closes issue #612 @the-dagger. Please do check it out and merge.

I added code to check for the following:
     No spacial characters before and after @
     (-) and (.) should not be together after @
     2 characters must before @ 
     Email length should be less 128 characters
{Follows regex validation technique}

How it works,
On press of the generate button the code checks if the email is a regular email address or not with the help of the checkemail() function. If not reload the page as in case if there was nothing was there as input else proceed and generate the apk file!


apk-generator/index.html
```diff
@@ -121,13 +121,19 @@ <h3 id="status"></h3>
                 $('#generator-progress').css('display', 'block')
                 $('#upload-progress').css('display', 'block')
                 var ary = $(form).serializeArray();
+                var emailse = $("#Email").val();
                 var obj = {};
                 updatePercentUpload(10);
                 for (var a = 0; a < ary.length; a++) obj[ary[a].name] = ary[a].value;
                 console.log("JSON", obj);
                 if (obj.Email == "" || obj.App_Name == "") {
```
the-dagger: Can you integrate the email check in this if statement rather than creating elseif?

Abhi2424shek: @the-dagger, sure but that would decrease the user experience. For example think of you having to fill up 20 forms with various attitudes[email,name,phone number etc] and you filled up wrong and then it shows that you entered no details at all! I agree that it can increase speed but only a bit and that bit is useless if you give up a lot of user experience the other end. We have to code for the user and not for us or anything that would look good for a programmer! :)

the-dagger: This will in fact make a good user experience.
Saying user enters wrong email and empty API link, so according to your
implementation he will be shown 2 alerts,which is not a good ux.

Speaking of your point, sure we can remove the line where whole page
refreshes on wrong input.
```setTimeout("location.reload(true);", 1);```
That should address the issue of user having to fill the whole form again.

>

-- 
Best,
Harshit Dwivedi
Contact : +919968195588
GitHub <https://github.com/the-dagger> | LinkedIn
<https://in.linkedin.com/in/hadwivedi>
Google Play <https://goo.gl/wl3loD> | Facebook
<https://www.facebook.com/daggerdwivedi>

Abhi2424shek: Yes do it out remove the `setTimeout("location.reload(true);", 1);

Plus what if I show all errors in a alert box. That will be a good UX right. @the-dagger .

The code now does is that first it checks whether email and api-links are empty if empty the user is asked to correct it. Then it checks for email to be right, if not it will alert and then proceed. Two alerts will never be shown at the same time as "if and else if" are used. If only "if and if" are used it will throw up 2 alerts...

the-dagger: Ok sounds good. :+1:
Also , just noticed that it was elseif.
I was thinking that it was if the whole time.
looks good, proceed ahead.

Abhi2424shek: K so you could merge this please @the-dagger .

PR #664, opened 2016-08-26
Better ss

SparshBansal:
Improved splash screen performance. Also i would like to suggest that we remove splash activity altogether. If you would look at apps such as Play Music or Youtube, they don't have a dedicated splash activity. Instead until their content view is constructed they show the splash image to the user. 
Please consider my request.


Abhi2424shek:
That would be a nice Idea but loading the intent first with lots of data can leverage performance right! That's why a new intent with small contents which would load and make the main intent load in the background as an async task is used for!

mananwason:
@SparshBansal LGTM. Merging.
About the splash screen I think it's best to keep it for now because if the data keeps downloading the user might get frustrated, hence we need a splash screen to be there.

Issue #672, opened 2016-08-28
Hide the generate and download app button once the build starts


mananwason:
Right now the generate and download app button is shown even when the build starts. It should be hidden or atleast made unclickable

the-dagger:
Fixed already

PR #681, opened 2016-08-29
Fixes

### LINKED ISSUE #671: Code cleanup and refactoring
Right now there are a lot of commented code on the app generator scripts. We need to remove it.

the-dagger:
Fixes #671 and #672 
@mariobehling @mananwason Please test the site at [http://45.55.58.149/](http://45.55.58.149/) with fossasia zip and merge.


mariobehling:
If the goal is to make the buttons inactive, then this is working. Thank you!

Issue #724, opened 2016-10-12
[Improvement] setContentView and Butterknife.bind redundant code


PareshMayani:
We can declare an abstract method for getting layout ID and can set the same directly in base activity.

This would allow us not to write setContentView() and binding the same layout using Butterknife. It will also help us in overcoming a mistake of forgetting setting layout sometimes!

Issue #723, opened 2016-10-12
[Improvement] android.R.id.home should be defined in BaseActivity


PareshMayani:
Below code is a redundant code in activity and it should be written in BaseActivity only.

```
    @Override
    public boolean onOptionsItemSelected(MenuItem item) {
        switch (item.getItemId()) {
            case android.R.id.home:
                onBackPressed();
        }
        return super.onOptionsItemSelected(item);
    }
```

P.S. I have already written code, sending a PR soon :)

Issue #722, opened 2016-10-12
[improvement] Deprecated methods and APIs used


PareshMayani:
We have used some deprecated methods, it's not going to affect anything but it's good always to replace it with updated solutions provided.

For example, getColor and getDrawable is deprecated, and they have provided ContextCompat.getcolor method for the solution.

Issue #721, opened 2016-10-12
[improvement] Declare variables for dependencies and refer it


PareshMayani:
There are dependencies which is being referred by two-three libraries/artifacts. For the same, it would be good if we declare a common variable and refer it. It's good practice to follow, as we just need to change only line/one variable, just update the variable with new version and that's it.

def SUPPORT_LIB_VERSION = "24.2.1"
def BUTTER_KNIFE_VERSION = "8.4.0"
def RETROFIT_VERSION = "2.1.0"
def LEAK_CANARY_VERSION = "1.3.1"
def ESPRESSO_VERSION = "2.2.2"

PR #725, opened 2016-10-12
Code improvements optimizations

PareshMayani:
Solutions for #721 , #722 , #723, #724


the-dagger:
@PareshMayani Looks good :+1: 
The test cases are failing because the API link mentioned in the code returns a 404 and hence the corresponding test is failing.
I will fix it asap.

PareshMayani:
@the-dagger thanks for merging but that's squash merging. I think it's not a good practice to make a squash merge as it merges all the commits into a single commit and there we loss commit history!

PR #727, opened 2016-10-13
re-factor: no need to have "ok" string, instead we can use in-built o…

PareshMayani:
No need of having "ok" string in strings.xml, instead we can use in-built ok string which is also having translations provided in-built.

PR for #726


PareshMayani:
@the-dagger you can squash commit if build goes successful!

Issue #720, opened 2016-10-07
App Crashing if internet is off


amitiwary999:
I found crash by doing following step(internet off)
1.Open app 
2.Click on any option form navigation drawer.
3.Click on track option and when track page open refresh it.
   And app crashes.

amitiwary999:
I would like to work on it.

Issue #715, opened 2016-09-26
App crashing


passiondroid:
Found a crash. 
Steps which i performed - 
1. Launched the app
2. Selected a track
3. Opened a session detail
4. Bookmark the session
5. Press back twice to reach the first activity
6. Now do swipe refresh. As soon as the refresh gets finished, app crashes.

Many times app did not crashed by performing the above steps. So there may be few other things which is causing the crash. Below is the crash log - 

> 09-26 18:07:00.236 19148-19148/org.fossasia.openevent E/AndroidRuntime: FATAL EXCEPTION: main
>                                                                         Process: org.fossasia.openevent, PID: 19148
>                                                                         java.lang.RuntimeException: Could not dispatch event: class org.fossasia.openevent.events.TracksDownloadEvent to handler [EventHandler public void org.fossasia.openevent.fragments.TracksFragment.onTrackDownloadDone(org.fossasia.openevent.events.TracksDownloadEvent)]: Attempt to invoke virtual method 'void android.support.v4.widget.SwipeRefreshLayout.setRefreshing(boolean)' on a null object reference
>                                                                             at com.squareup.otto.Bus.throwRuntimeException(Bus.java:460)
>                                                                             at com.squareup.otto.Bus.dispatch(Bus.java:387)
>                                                                             at com.squareup.otto.Bus.dispatchQueuedEvents(Bus.java:368)
>                                                                             at com.squareup.otto.Bus.post(Bus.java:337)
>                                                                             at org.fossasia.openevent.OpenEventApp$1.run(OpenEventApp.java:58)
>                                                                             at android.os.Handler.handleCallback(Handler.java:746)
>                                                                             at android.os.Handler.dispatchMessage(Handler.java:95)
>                                                                             at android.os.Looper.loop(Looper.java:148)
>                                                                             at android.app.ActivityThread.main(ActivityThread.java:5443)
>                                                                             at java.lang.reflect.Method.invoke(Native Method)
>                                                                             at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:728)
>                                                                             at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:618)
>                                                                          Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.support.v4.widget.SwipeRefreshLayout.setRefreshing(boolean)' on a null object reference
[...]

PR #733, opened 2016-10-16
fix issue #720 #715 app crash when revisit any page: location, speaker, sp…

dhluong90:
it's easy to reproduce. 
1. click track.
2. click speaker
3. click track. ====> crash. 
it's also happen with speaker, track, sponsors and location page when user revisit them.
Because OnDestroyView it unbind all view, but this instance still keep as handler event. 
So when swipe refresh, download data and call back handler event, Null pointer exception appear. 
We need to unRegister event in onDestroyView too.  
Please help me check it.


mariobehling:
@dhluong90 Good job!

Issue #747, opened 2016-10-19
Revamp speaker details layout


PareshMayani:
We need to revamp the speaker details layout as it looks somewhat messy and not utilizing the space perfectly.

P.S. I have started working on it and will send a PR

PareshMayani:
Attached is the revamped layout design of the speaker details screen. Hope it looks awesome!

![screenshot_20161021-000243](https://cloud.githubusercontent.com/assets/888080/19572927/c6c8d66c-9721-11e6-8a62-acde04758650.png)

the-dagger:
Looks pretty nice to me.
@mariobehling Thoughts?

mariobehling:
Yes, good job. Let's proceed.

PR #750, opened 2016-10-20
revamp speaker details layout with material design layouts and standards

PareshMayani:
Solution for #747. Revamp speaker details layout with material design layouts and standards.


android/app/src/main/java/org/fossasia/openevent/activities/SpeakerDetailsActivity.java
```diff
@@ -65,10 +66,11 @@
     @BindView(R.id.imageView_github) ImageView github;
     @BindView(R.id.imageView_twitter) ImageView twitter;
     @BindView(R.id.imageView_web) ImageView website;
-    @BindView(R.id.speaker_name_title) TextView speakerName;
+    @BindView(R.id.speaker_details_title) TextView speakerName;
     @BindView(R.id.speaker_bio) TextView biography;
-
+    @BindView(R.id.speaker_details_header) ViewGroup header;
```
PareshMayani: @the-dagger I think we need to write a custom rule to avoid this because Butterknife BindView must not be private AFAIK.

the-dagger: Yes it shouldn't be.

Issue #759, opened 2016-10-23
Back arrow button is not working in setting activity


amitiwary999:
Back arrow button in setting activity is not working.(Reason: R.id.home is moved to Baseactivity and in setting activity we are not extending Baseactivity)

amitiwary999:
I would like to work on this.

PR #760, opened 2016-10-23
Fixes issue #759 back arrow button is not working in setting activity

amitiwary999:
Fixes issue #759 back arrow button is not working in setting activity


mariobehling:
Thanks.

PR #772, opened 2016-10-28
fixes issue 771

amitiwary999:
fixes issue 771 Share option is not working in TracksActivity.


android/app/src/main/java/org/fossasia/openevent/activities/TracksActivity.java
```diff
@@ -106,8 +106,8 @@ public boolean onOptionsItemSelected(MenuItem item) {
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
         super.onCreateOptionsMenu(menu);
-        getMenuInflater().inflate(R.menu.menu_tracks, menu);
-        searchView = (SearchView) menu.findItem(R.id.action_search_tracks).getActionView();
+        getMenuInflater().inflate(R.menu.menu_tracks_activity, menu);
```
the-dagger: I think that this is already fixed in #768

Issue #804, opened 2016-12-03
Schedule's top nav can be improved 


rhemon:
I realised the top nav in Schedule, wouldn't it be nice to have each of them evenly distributed using layout_weight
![screenshot_2016-12-03-20-14-54](https://cloud.githubusercontent.com/assets/6615517/20859909/1b2a7ba2-b996-11e6-898b-c3b93a9ee601.png)

rhemon:
I would like get assigned to solve this for my task https://codein.withgoogle.com/dashboard/task-instances/6039001716752384/

rhemon:
this thing is achieved by doing this 
```
app:tabMaxWidth="0dp"
app:tabGravity="fill"
app:tabMode="fixed"
```
![schedule](https://cloud.githubusercontent.com/assets/6615517/20888189/017ee100-bb28-11e6-8b17-b873dfbad7c8.png)

But now the problem is previously `tapMode` was set to `scrollable`. So I am not sure about actually if I should do that change (what if it was set that because some cases have more days?), so if we want it scrollable, then following material design, I was thinking maybe I shoud try adding at 72dp gap, not exactly sure how to do this. One way I can think of is by adding blank tab with width 72dp, if there is any better way please say so? And if making it fixed is fine, then thats also cool. 

@mariobehling @the-dagger

the-dagger:
Keep it scrollable @rhemon.
We don't want to hardcode anything.

rhemon:
ok so then I should add the 72dp gap at the beginning?

the-dagger:
Not sure why that is needed though.

Best,
Harshit Dwivedi
Contact : +919968195588
GitHub <https://github.com/the-dagger> | LinkedIn
<https://in.linkedin.com/in/hadwivedi>
Google Play <https://goo.gl/wl3loD> | Facebook
<https://www.facebook.com/daggerdwivedi>

PR #810, opened 2016-12-05
Fixing Issue #804

rhemon:
Since we are using scrollable TabLayout, according material design it should have a marginStart of 72p.


the-dagger:
There are commits from your earlier PR as well.
Ideally you should update your fork and then send a pull request,

rhemon:
Oops sorry, should I try rebasing?

the-dagger:
Yep

rhemon:
I don't know why but I seem to have some unknown comits that I am not well aware of, so gonna get rid of do a new pull request, sorry.

PR #811, opened 2016-12-05
Fixing Issue #804

rhemon:
Since we are using scrollable TabLayout, according material design it should have a marginStart of 72p.


rhemon:
ok I am sorry, I am unable to do the rebasing, I don't know there are these file (Track.java, ReadJson.java, readJson.java, Microlocation.java) changes it shows me from the beginning
if I commit it shows error
if I dont comit it doesn't let me rebase
i tried git clean, git stash but it didn't do any change, if its not ok with three commits please tell me what to do,

rhemon:
I think I managed to do it with another branch, by creating with `--orphan` and then do one commit, will do a pull request with that.

Issue #793, opened 2016-11-27
Double backpress to exit.


free4murad:
In most of the apps if the user presses back button to exit, a toast message appears which prompts the user to press back button again to exit.

I suggest implementing the same in this app

the-dagger:
Cool, go ahead.

free4murad:
@the-dagger 
I have implemented the functionality locally, but when I push it onto my copy of project the Travis build is not successful.

the-dagger:
What is the error log?
Also we are not using travis anymore, we have shifted to CircleCi.
Cheers.

Danki567:
Hello, I have implemented the feature. As well as add a toast message that notifies the user, that has been translated into all languages

PR #817, opened 2016-12-07
Exit on double back press implemented.

free4murad:
Solved issue #793


free4murad:
@the-dagger 
this is a new PR with regard to issue #793

Issue #828, opened 2016-12-10
Custom language select in user settings


rohanagarwal94:
Since the app is thinking of having regional language too, it would be better to have selection of language in the user settings to change it if he wants. @the-dagger please assign me this.

mariobehling:
Please use the phone settings instead of having a separate implementation in the app.

PR #832, opened 2016-12-11
Language preferences in user settings

rohanagarwal94:
#828 is fixed. @the-dagger and @mariobehling please review.
screenshot_2016-12-11-20-04-22-020_org fossasia openevent


the-dagger:
Rather than setting it to `en`, could you show the actual language to the user?
I personally feel like displaying `English` makes more sense than displaying `en`

rohanagarwal94:
@the-dagger the default language comes from the system and might vary for custom android devices. Therefore cant really do string mapping too.

the-dagger:
You won't need to map it.
Have a look over here.
[http://stackoverflow.com/a/23168383/5471095](http://stackoverflow.com/a/23168383/5471095)

rohanagarwal94:
@the-dagger thanks for helping out. Made the necessary changes.

Issue #868, opened 2016-12-19
Improvement in Splash screen


utkarshnath:
Currently splash screen waits for 1 sec and then all background works starts, thus we see a blank screen for some seconds before we could see the actual app.
I would suggest implementing splash screen such that it is seen by the user until the background data is loaded and there should be no blank screen in between.

AlabhyaVaibhav:
@utkarshnath i would like to work on this issue

utkarshnath:
@AlabhyaVaibhav I have already resolved this issue #869

AlabhyaVaibhav:
Oh.. cool is there anything else i could work upon ?

PR #869, opened 2016-12-19
Improved splash screen

utkarshnath:
Solved #868


mananwason:
Could you add a screenshot or better yet a gif for the changes? Thanks

utkarshnath:
@mananwason Done
**Now we don't have blank screen in between splash screen and main activity**
www gifcreator me_rbwm29

mananwason:
@utkarshnath How have you handled the cases when the download takes a long time? Is there a timeout? Also what about if the user doesn't want to download the data and instead wants offline json content that is packaged into the app?

utkarshnath:
@mananwason I am waiting only till the views of the main Activity is constructed(so that we don't get a blank screen) and not till the data gets downloaded. So as soon as view is constructed we move on to main activity and downloading starts and we can see the progress bar if the internet is present otherwise we receive a dialog box for switching on the internet.

Issue #895, opened 2016-12-23
Change in LocationActivity class code


arp95:
The main issue is when we put the entire adapter code in onCreate() method then the bookmark status gets updated only when onCreate() is called again.
To ensure the bookmark status is properly updated in other cases the code has to be moved from onCreate() to onStart() method.

PR #896, opened 2016-12-23
fixes issue #895

arp95:
made changes to the bookmark status by copying the adapter code from onCreate() to onStart() method.


mananwason:
LGTM. Merging.

Issue #900, opened 2016-12-24
Bookmark should be marked from session list itself


utkarshnath:
currently bookmark is marked only from the sessionDetailActivity and the button in session list doesn't work

PR #901, opened 2016-12-24
marking bookmark from session list

utkarshnath:
solved #900


mananwason:
@utkarshnath Looks good. Merging

PR #919, opened 2016-12-30
solved #911

utkarshnath:
solved #911 and removed the splash activity as its not being used now.


android/app/src/main/java/org/fossasia/openevent/activities/SplashActivity.java
```diff
@@ -1,32 +0,0 @@
-package org.fossasia.openevent.activities;
```
the-dagger: Why did you remove the splash activity?
Even though it doesn't serves any meaningful purpose as of now, but we can use this for prefetching the session data and other time consuming tasks on app launch.

utkarshnath:
@the-dagger added back the splash Activity

PR #934, opened 2016-12-31
Fixes #932

### LINKED ISSUE #932: Session Timing is null while sharing the session url
**Actual Behaviour:**
Session Timing is null while sharing the session URL 

**Expected Behaviour:**
It should show the start time and end time of the session.

**Screenshots:**
screenshot_20161231-231436

_**I would like to work on this issue.**_

samsruti:
Screenshot of the Change:
capture


mananwason:
LGTM, merging.

PR #937, opened 2017-01-01
Update strings.xml

pipix51:
Added more strings into the Simplified Chinese version of the file for internationalisation.


mananwason:
@pipix51 Why have you deleted strings? Have you checked if these strings are still there in the other string resources?

pipix51:
@mananwason  Yes, I have check if the strings are still in the other string resources. I just repositioned those strings so that it is similar to the original strings.xml layout.

mananwason:
Okay Thanks. I'm merging this!

PR #983, opened 2017-01-08
Added notification when bookmarked from Sessionlist

### LINKED ISSUE #967: We don't get notification when a session is bookmarked from list
alarm is not set when a session is bookmarked from SessionList directly.

utkarshnath:
@niranjan94 made fresh PR, fixed #967


Abhi2424shek:
Just put a # before the issue number pls @utkarshnant

utkarshnath:
@Abhi2424shek typo error :)

the-dagger:
@utkarshnath Can you explain why this would be useful to the user?
Because as of now, we are already displaying a notification when the user's bookmarked event is about to start.

utkarshnath:
@the-dagger yes, we do display the notification when the user's bookmarked event is about to start but only when session is bookmarked from sessionDetailActivity and not when bookmarked from the sessions list directly. So I feel we should display notification irrespective of from where user has bookmarked a session

mananwason:
LGTM merging. Thanks!

PR #987, opened 2017-01-09
Share App feature

rohanagarwal94:
Fixes issue #899


mananwason:
Same here. Please rebase from upstream branch first. Then add commits

PR #990, opened 2017-01-10
App generator v2

niranjan94:
Live instance: http://android-generator.eventyay.com/


mariobehling:
Thank you! Great.

Let's get travis work in following steps again as well.

PR #994, opened 2017-01-11
Fixes overlapping progressbar

rohanagarwal94:
Fixes issue #831 and refactors about_us dialog by using dialog factory method.


mananwason:
@rohanagarwal94 please rebase with upstream and then add commits

PR #1000, opened 2017-01-13
Share app feature

rohanagarwal94:
Fixes issue #899 . Will create separate issue for app invite with facebook.


mananwason:
As Per our discussion in the issue, this model will not work for different apps. Please retrieve the package name from the app and share that. Don't share a static app url. What if this app is generated for another event? This will not work. Please rectify that, otherwise this looks good. Cheers!

rohanagarwal94:
@mananwason updated.

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -445,6 +445,20 @@ public void run() {
...
+                try {
+                    Intent shareIntent = new Intent();
+                    shareIntent.setAction(Intent.ACTION_SEND);
+                    shareIntent.setType("text/plain");
+                    shareIntent.putExtra(Intent.EXTRA_TEXT,
+                            String.format(getString(R.string.whatsapp_promo_msg_template),
+                                    String.format(getString(R.string.app_share_url),getPackageName())));
+                    startActivity(shareIntent);
+                }
+                catch (Exception e) {
```
the-dagger: This is not the correct way to handle such events.
Rather, you should use `startActivityForResult` and check for the correct request code in `onActivityresult`.
It might be possible that the user chooses not to share the event and presses back button, you need to acknowledge that action and let the user know that the event wasn't shared and he/she can  retry it again by adding an action to the SnackBar.
Same can be done for a successful share too.

rohanagarwal94: @the-dagger well pointed. Changing it ASAP.

rohanagarwal94:
@the-dagger please see the update.

Abhi2424shek:
@rohanagarwal94 updates?

rohanagarwal94:
@Abhi2424shek already mentioned in the thread that Intent.ACTION_SEND is not set up for use with startActivityForResult(). Check the thread in which @the-dagger asked for changes in the PR.

Issue #958, opened 2017-01-03
improvement in the UI


arp95:
Improving the ui of the speakers and schedule activity.

mananwason:
How? Could you add prototypes or just describe the changes you plan. Prototypes would be appreciated.

arp95:
I am referring to the tablet view of both where the display of activities seems extended and has to have a autospan gridlayout to overcome that. 

![15902662_10210135775065346_1175802889_o](https://cloud.githubusercontent.com/assets/8268392/21739880/f0ce1546-d4cd-11e6-866c-968b2d875b1b.jpg)
![15942956_10210135774985344_967764428_o](https://cloud.githubusercontent.com/assets/8268392/21739882/f7c6ea76-d4cd-11e6-91ab-52508396e17a.jpg)

mananwason:
Please go ahead

mariobehling:
Issues like "Improving UI" or "Improving App" are not specific enough. Please also provide screenshots using the phone screenshot feature in future. Thank you. 

Closing this. Please provide specific ideas for improvements and open a specific issue for this if needed.

PR #1008, opened 2017-01-15
fixes issue #958

arp95:
made changes to ui as follows:

15902662_10210135775065346_1175802889_o 1
16108009_10210219484958041_533561470_o


the-dagger:
The speaker fragment looks OK to me,  but I'm not sure about the schedule viewpager.
The default ui is good enough, we don't need gridview everywhere.

arp95:
on the tablet view it didnt seem fine.  So should i submit only the changes in speaker then?

the-dagger:
Yes please commit the changes only for speaker fragment.
And as mentioned previously, also have a look at the other
fragments/activities as well if they looked weird on tablet layout and
commit changes for them in this pull request.

I'll put this pr on hold till then.

>

-- 
Best,
Harshit Dwivedi

arp95:
@the-dagger you can have a look at the speakers fragment changes.
According to me all the activities look fine on tablet view. You can have a look now.

Issue #893, opened 2016-12-23
Improvement in Zoom In and Zoom Out function of Map


tarsdev:
When we open map from map option from phones that do not have physical buttons the zoom in and zoom out function comes back of on-screen navigation buttons and we cannot use zoom buttons because it is back of navigation.
1.
![capture _2016-12-23-10-53-26](https://cloud.githubusercontent.com/assets/24253823/21449389/a8647514-c912-11e6-9a08-faa6a1494d82.png)
2.
![capture _2016-12-23-11-00-01](https://cloud.githubusercontent.com/assets/24253823/21449396/d13cd152-c912-11e6-94c8-7c3ce5384650.png)
This must be fixed.

And when fully zoom out of map then it shows many maps of world, it should show only one complete map!
3.
![screenshot_2016-12-22-18-53-56](https://cloud.githubusercontent.com/assets/24253823/21449473/93b0506a-c913-11e6-940e-01901a29af41.png)
This is not useful but it should be proper.

the-dagger:
You can go ahead and work on it if you like. 👍

free4murad:
@tarsdev
are you working on it?

tarsdev:
I am still learning coding so I will try to solve it after I know about languages!

free4murad:
@the-dagger 
If we need to fix this ASAP, Can I make a PR for this?

the-dagger:
@free4murad you can work on this.
@tarsdev I'd suggest you to find and work on some simpler UI related issues if you'd like to. 🙂

free4murad:
@the-dagger 
making a PR now.

jsroyal:
@the-dagger Close this issue.

free4murad:
Solved via #1025
Closing this.

PR #1025, opened 2017-01-18
Map controls fixed.

free4murad:
Fixes issue #893

Changes: Map controls are no longer hidden under navigation buttons.
@the-dagger please review.


android/app/src/fdroid/res/layout/fragment_map.xml
```diff
@@ -9,4 +9,11 @@
         tilesource="Cyclemap"
         android:layout_width="match_parent"
         android:layout_height="match_parent" />
+
+    <ZoomControls
+        android:id="@+id/zoomControls"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center|bottom"
+        android:paddingBottom="50dp"/>
```
the-dagger: Why the fixed padding?
We should set it programmatically above the nav bar.
I understand that ideally the nav bar is 48dp, but this might not be the case with every device out there.
Many OEMs choose not to include it at all.
I'd suggest to add a fixed margin of 4dp around the `ZoomControls` and then add a bottom padding equal to the nav bar size.

free4murad: @the-dagger 
Sure I am on it.
One question though, we need only bottom padding, why should we add padding on other sides?

the-dagger: When I said 4dp margin, I meant bottom margin only.
Sorry about the miscommunication.

free4murad: Made the changes
please review

free4murad:
Tested on 2 devices
1. Moto G4 Plus (with navigation bar)
moto g4 plus
2. Samsung Galaxy J2 (without navigation bar)
samsung j2

Issue #1032, opened 2017-01-19
Adding text drawables to tracks activity


brainbreaker:
UI of tracks activity can be significantly improved by adding the text drawables to the items just like Gmail application.
 
<Img src = "https://s23.postimg.org/qch7eqkiz/16129442_1239014579508633_1312515403_o.png"/>

@the-dagger Your views?

brainbreaker:
Please review https://github.com/fossasia/open-event-android/pull/1033

PR #1033, opened 2017-01-20
TextDrawables added to improve the UI of tracks activity

brainbreaker:
Fixes issue #1032 

Changes: [Text drawables added to tracks activity]
[Screenshot](https://s28.postimg.org/618kwelvh/Whats_App_Image_2017_01_20_at_11_28_52.jpg) for the change


android/app/src/main/res/layout/item_track.xml
```diff
@@ -1,45 +1,42 @@
...
-            android:textColor="#89000000"
             android:textSize="14sp"
-            tools:text="Track description" />
+            android:textColor="@color/secondary_text"
+            android:layout_width="match_parent"
+            tools:text="Track description"
+            android:layout_height="wrap_content" />
     </LinearLayout>
-
-    <View
```
the-dagger: For our use case I believe a simple view added to the end of each item would be a better option rather than implementing a custom item decorator for the divider.
What do you think @brainbreaker ?

brainbreaker: Yes, indeed. I have updated the pull request
- Removed the custom divider view class
- Optimised the imports

Issue #1027, opened 2017-01-18
Bookmark widget is not updated instantly.


free4murad:
**Actual Behaviour**

If we bookmark a session, it is not added to widget to instantly.

**Expected Behaviour**

As soon as a session is bookmarked, it should be added to the widget.

@the-dagger @mananwason @mariobehling 
Should I make a PR.

free4murad:
@the-dagger 
your reviews about this issue???

the-dagger:
Ok, sure.
Go ahead.

free4murad:
Solved in #1077 
Closing

PR #1077, opened 2017-01-28
Widgets update issue fixed.

free4murad:
Fixes issue #1027 

Changes: Widget is updated as soon as we add or remove a bookmark


android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -133,6 +135,12 @@ public void onClick(View v) {
                     createNotification();
                     Toast.makeText(SessionDetailActivity.this, R.string.added_bookmark, Toast.LENGTH_SHORT).show();
                 }
+
```
saketkumar95: @free4murad Can you remove this extra line here and at line 143? It would better if the code is formatted in better way. :)

Issue #1109, opened 2017-02-15
Renaming the toolbar's title (Session Details) to the Session's name in Session Detail Activity. 


harshitagupta30:
**Actual Behaviour**

![screenshot from 2017-02-15 14 38 34](https://cloud.githubusercontent.com/assets/12701036/22986881/f748244c-f3d2-11e6-852d-a795c7cac227.png)

**Expected Behaviour**

![screenshot from 2017-02-15 23 00 28](https://cloud.githubusercontent.com/assets/12701036/22986883/f9af5bba-f3d2-11e6-92ad-9c88d24cd242.png)

![screenshot from 2017-02-15 23 04 56](https://cloud.githubusercontent.com/assets/12701036/22986951/36d7263a-f3d3-11e6-826b-c4c0b08378ec.png)

**Would you like to work on the issue?**
Yes!

@the-dagger @mananwason Please review! Can I send the PR for the same?

the-dagger:
Sure, please go ahead 👍

PR #1127, opened 2017-02-18
Toolbar's title changed to the Session's name in Session Detail Activity.

harshitagupta30:
Fixes  Issue #1109 

Changes: Used the library `multilinecollapsingtoolbar` to add the title to collapsing toolbar and removed the title from the details of the session.

Screenshots for the change: 

screenshot from 2017-02-18 23 19 54

@the-dagger @mananwason Please Review! :)


android/app/build.gradle
```diff
@@ -91,6 +91,7 @@ dependencies {
     compile 'com.jakewharton:butterknife:8.4.0'
     compile 'com.amulyakhare:com.amulyakhare.textdrawable:1.0.1'
     apt 'com.jakewharton:butterknife-compiler:8.4.0'
+    compile 'net.opacapp:multiline-collapsingtoolbar:1.3.1'
```
the-dagger: The method count looks quite high to me.
http://www.methodscount.com/?lib=net.opacapp%3Amultiline-collapsingtoolbar%3A1.3.1

Isn't there an alternative solution though?

harshitagupta30: The another one is by creating a custom collapsing toolbar similar to one we are using in speaker's activity title! and I tried and it look similar to this but for that, we need to change the menu items `app:showAsAction="ifRoom"` to`never`. Shall I update the PR with this solution?

the-dagger: Um, no.
Let's not do that.
Try looking for an alternative library and if not then we can merge this instead.
How about you take a look at the I/O app to see how they are doing it.

harshitagupta30: This is how it would look like if use custom Collapsing toolbar! we don't need to hide the menu items in this case.
It looks good to me. 
screenshot from 2017-02-20 22 02 50
screenshot from 2017-02-20 22 02 59

What's your thought on this?

the-dagger: Yes, looks good.

harshitagupta30: Shall I update the PR with this code?

the-dagger:
Yes.​

harshitagupta30:
@the-dagger the codacy issues `Use explicit scoping instead of the default package private level` has been raised! I guess I don't need to make any changes? Please review!
Also 2 files from the last commit also got added while resetting the head. Do I need to delete them or it would work and doesn't create any issue while merging? 
Sorry for bothering :)

PR #1135, opened 2017-02-19
CrossIcon changed with BackIcon in CustomCustomTab

### LINKED ISSUE #1122: Use of BackIcon instead of CrossIcon in ChromeCustomTabs
**Actual Behaviour**
 
We got a button with Cross-Icon on Top-Left screen of each and every ChromeCustomTab in SponsorFragment which comes in default.

ezgif com-video-to-gif 1

**Expected Behaviour**

Cross Icon 

 should be replaced with Back-Icon because of following reasons:

1.) It makes more sense because we are going back where we came from(i.e., from ChromCustomTab to where we launched it), even from User's Perspective

2.) And as our exitAnimations are in Horizontal Direction(i.e, from Left or Right), Back sign makes more relevance.
( Cross-Sign makes Significance in Top-Bottom transition.)

ezgif com-video-to-gif

Would you like to work on the issue?

Yes, I would like to work on the issue.
@the-dagger  @mananwason  @mariobehling

doomers:
Fixes #1122 

Changes: BackIcon has been used instead of CrossIcon in ChromeCustomTabs.

Screenshots for the change: 
after_back_icon


free4murad:
@doomers 
You can convert the png to a vector drawable.

doomers:
@free4murad SVGs doesn't support ChromeCustomTab, Moreover, it has to follow a certain dimension as stated in the Documentation

https://developer.android.com/reference/android/support/customtabs/CustomTabsIntent.html#KEY_ICON

and we need only 1 image of BackArrow for all Screens, so why to Convert it to vector drawable
Because Vector Drawable is used to save space for 5 PNGs, which we are not going to save it anyway.

free4murad:
@doomers 
Using one xml file instead of 5 PNGs is not the only advantage of Vector Drawable.
Vector drawables are smaller than PNGs.

But if ChromeCustomTab don't support them noting can be done.

PR #1139, opened 2017-02-20
Improved UI of Session's list

### LINKED ISSUE #1116: Changes in UI of  Tracks Session list
**Actual Behaviour**

screenshot_20170216-213306

**The list represent sessions for a particular track thus track's name shouldn't be present in all list items**

**Expected Behaviour**

screenshot_20170216-224922

@the-dagger @mananwason @mariobehling do you think the UI is fine. Should I send a PR for same?

utkarshnath:
Fixes #1116 

**Earlier**

screenshot_20170216-213306

**NOW**
screenshot_20170216-224922


the-dagger:
@utkarshnath any specific need for doing this?
Technically there is a color associated with each track and this might cause some confusion among users.

https://github.com/fossasia/open-event/blob/master/sample/FOSSASIA16/tracks#L5

PR #1141, opened 2017-02-20
Use palette to tint card header and icons (Fixes #1119)

### LINKED ISSUE #1119: Unused Palette generation in SpeakerDetailsActivity
**Actual Behaviour**

In SpeakerDetailsActivity, a palette is generated based on the loaded speaker's photograph to tint the share and back buttons. However, it is unused. The generated palette is not used to generate the color of tint, rather Color.WHITE is directly used to color them white. The icons are already white, and if not, they were addressed in #1118 

**Expected Behaviour**

We should either remove the palette generation altogether as is it not crucial or use the generated palette to tint the icons.

**Screenshots of the issue**

SpeakerDetailsActivity.java

Yes. I would like to work on the issue.
@mananwason @the-dagger @mariobehling

iamareebjamal:
Fixes #1119 

Changes: 
- Uses Palette to tint the card header and icons in SpeakerDetailActivity

Screenshots for the change: 
Collapsed Toolbar
Expanded Toolbar


iamareebjamal:
The codacy bot counts variable scoping as a bad qualifier, but as we can't do anything in case of using ButterKnife, I suggest this pattern should be disabled for builds.

But for the sake of this PR, what should I do @the-dagger ?

the-dagger:
Hey, since the palette is using muted dark for the background, can we set the textColor of the title to white instead?

iamareebjamal:
OK. Doing that

the-dagger:
👍

iamareebjamal:
Updated Image

Is it OK?

the-dagger:
Looks good. :+1:

PR #1177, opened 2017-02-28
Bookmarks Item in navigation drawer remains clicked if no bookmarks p…

### LINKED ISSUE #1174: UI bug: Navigation Drawer (Bookmarks)
**Actual Behaviour**
Bookmarks item in navigation drawer remains clicked if you try to open it and there are no bookmarks

**Expected Behaviour**
It should show the activity as clicked which the user is currently inside.

**Steps to reproduce it**
Make sure you have no bookmarks. Open Navigation drawer and click on Bookmarks. It will show the error dialog of "No Bookmarks Available". After disposing it, open navigation drawer and you'll find Bookmarks as clicked. But it should show the activity as clicked in which the user is currently viewing.

**LogCat for the issue**

**Screenshots of the issue**
[screenrecorder.zip](https://github.com/fossasia/open-event-android/files/802019/screenrecorder.zip)

Yes I would like to work on it.

jodiwaljay:
…resent

Fixes #1174 

Changes: If bookmarks database is empty, it returns false so that item is not shown clicked and rest is unaffected. Therefore the usual dialog box appears for no bookmarks.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -395,6 +395,12 @@ public class MainActivity extends BaseActivity {
                     public boolean onNavigationItemSelected(MenuItem menuItem) {
                         int id = menuItem.getItemId();
                         doMenuAction(id);
+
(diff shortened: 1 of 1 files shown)


mananwason:
LGTM merging

PR #1199, opened 2017-03-01
fixes issue #1195

### LINKED ISSUE #1195: Adding Deep Links to the app
**Actual Behaviour**

Currently if we give any link to a user there are chances it will open it in Chrome or any other browser instead of the Open Event app (if installed)

**Expected Behaviour**

Adding various intent filters in the manifest file for the activites which will enable links to be viewed on our app instead of viewing it on any browser.

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

@the-dagger please review.

arp95:
Fixes #1195 

Changes: 

Added support for the uri: http://2017.fossasia.org/tracks

Now whenever the user tries to open the uri sent through eg a messaging app, they will be prompted to view  it via the app or the browser which was earlier happening only through the browser.


harshitagupta30:
@arp95 Can you please update your commit message according to [guidelines](https://github.com/fossasia/open-event-android/blob/development/docs/commitstyle.md) :+1:

arp95:
@harshitagupta30 
sure !

PR #1200, opened 2017-03-01
Fix: navigation drawer lag #1173

### LINKED ISSUE #1173: Animation stutter in Navigation Drawer
**Actual Behaviour**

While closing the drawer, it experiences a stutter in the animation. You may have felt it while opening schedule or tracks activity.

**Expected Behaviour**

The animation should be smooth

**Steps to reproduce it**

Try to open an activity from navigation drawer which is expensive in number of operations like "Tracks" or "Schedule". You will feel abruptness in animation of Navigation Drawer
 
I am adding an excerpt from https://developer.android.com/reference/android/support/v4/widget/DrawerLayout.html to show that it is a very common issue. I myself came across this a while ago in another project. It is 4th paragraph from the top in the link.

_DrawerLayout.DrawerListener can be used to monitor the state and motion of drawer views. Avoid performing expensive operations such as layout during animation as it can cause stuttering; try to perform expensive operations during the STATE_IDLE state. DrawerLayout.SimpleDrawerListener offers default/no-op implementations of each callback method._

There are different approaches to this problem. I have used one of them. I'll try to choose the best one and would like to work on it. Does this issue makes sense @the-dagger ?

P.S. more info about the issue here. http://stackoverflow.com/q/18343018/7622397

jodiwaljay:
**Fix: Navigation drawer abruptness while closing**

Whenever drawer closes or opens, the app should not perform heavy operations. Therefore a stutter or lag was observed. Now, the activity remains idle for that period of time when animation is being displayed and that eliminated the abruptness

Fixes #1173

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -401,83 +401,126 @@ public class MainActivity extends BaseActivity {
     }
 
     private void doMenuAction(int menuItemId) {
-        FragmentManager fragmentManager = getSupportFragmentManager();
+        final FragmentManager fragmentManager = getSupportFragmentManager();
         addShadowToAppBar(true);
         switch (menuItemId) {
             case R.id.nav_tracks:
-                atHome = true;
-                fragmentManager.beginTransaction()
-                        .replace(R.id.content_frame, new TracksFragment(), FRAGMENT_TAG_TRACKS).commit();
-                if (getSupportActionBar() != null) {
-                    getSupportActionBar().setTitle(R.string.menu_tracks);
-                }
-                appBarLayout.setExpanded(true, true);
+                smoothActionBarToggle.runWhenIdle(new Runnable() {
+                    @Override
+                    public void run() {
+                        atHome = true;
+                        fragmentManager.beginTransaction()
+                                .replace(R.id.content_frame, new TracksFragment(), FRAGMENT_TAG_TRACKS).commit();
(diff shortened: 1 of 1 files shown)


harshitagupta30:
@jodiwaljay Can you please update your commit message according to [guidelines](https://github.com/fossasia/open-event-android/blob/development/docs/commitstyle.md)? Thanks :+1:

jodiwaljay:
@harshitagupta30 I have made the changes. But while making the pull request these guidelines were not present. Instead there was an old template.

harshitagupta30:
@jodiwaljay Yeah! they've been added lately. That's I asked to change! kindly update the commit message using following command: 
`git commit --amend -m "fix: Navigation drawer lag"`

arp95:
@jodiwaljay 
nice idea !
But i still think creating a separate thread for the opening and closing of navigation drawer is beneficial as we are already having various threads for api calls, plus i didnt observe any lag as such.

jodiwaljay:
@arp95 Thanks for the concern. But I think sudden drawer disappearance is quite observant. You may want to check #1173 for reproducing steps. @harshitagupta30 The commit message already follow the guidelines and essentially the pull request has all the details required, if issue is needed to look up again in future

mananwason:
LGTM. Merging

mananwason:
@jodiwaljay I have revert this because this is causing the fragments to not be initialised and if is making the download buggy.

jodiwaljay:
Sorry for the late reply! I was having my exams. I'll try to see where is the problem. It would be a great help if you (@mananwason)can give me the steps to reproduce the bug.

PR #1234, opened 2017-03-05
fix: session title display problem in LocationActivity

### LINKED ISSUE #1233: Session Title not displayed properly in LocationsActivity
**Actual Behaviour**

As seen in the screenshot the title is not shown properly in LocationsActivity.

**Expected Behaviour**

Change in the layout file so as to display the title properly.

**LogCat for the issue**

-

**Screenshots of the issue**

17121500_10210641885437789_896852141_o

yes.

arp95:
Fixes #1233 

Changes: 
made changes to the layout file to display the title properly.

Screenshots for the change: 

17148890_10210642344649269_281172065_o


arp95:
@harshitagupta30

i didnt face your issue where the title was displayed with dots. As you can see over here the layout file had heoght set to 100dp which was the problem for the title not displayed fully.

harshitagupta30:
@arp95 You've tried to solve the same issue in a different way. Title with dots is not the issue. It's what I have changed if the title length is more than 2 lines it'll ellipsize at the end instead of going beyond the view.

arp95:
@harshitagupta30
i think this is a better way.

mananwason:
This disrupts the uniformity of the cards and hence will not look good. Instead I prefer @harshitagupta30 approach.

PR #1261, opened 2017-03-08
fix: App crashing due to OutOfMemory error in loading Bitmap

### LINKED ISSUE #1260: App crashing due to runtime OutOfMemory error
**Actual Behaviour**
Currently, we are using `bitmap` in SpeakerDetailsActivity due to which while loading a bitmap the app is raising an `OutOfMemory` exception at `runtime`. 

screenshot from 2017-03-02 21 59 14

**Expected Behaviour**
It should be resolved otherwise it will lead to the crash app.
 
Yes

harshitagupta30:
Fixes #1260

Changes: Added 2 properties in `AndroidManifest.xml` file.
1. android:hardwareAccelerated="false", corresponds to that every window created from activity will be hardware accelerated which has set to false.
2. android:largeHeap="true", request that our application's processes be created with a large Dalvik heap.

@the-dagger @mananwason Please review :+1:


iamareebjamal:
Using `android:hardwareAccelerated="false"` has lead to considerable drawer lag

PR #1263, opened 2017-03-08
fix: Change in description field variable in DayScheduleAdapter

### LINKED ISSUE #1262: Change in description field in DayScheduleFragment for better UI
**Actual Behaviour**

Currently, the field description in schedule list points to the `long_abstract` which is empty for almost all the sessions. The UI was designed in such a way to display the description of the session as well.

screenshot_20170308-161014

**Expected Behaviour**

While the summary i.e. `short_abstract` is non-empty for almost all sessions. Replacing description with summary will result in better UI. 

screenshot_20170308-160529
screenshot_20170308-160554

Yes

harshitagupta30:
Fixes #1262
Changes: 
- Instead of `getDescription()`, `getSummary()` is called.
- Removed unnecessary result variable.

Screenshots for the change: 

screenshot_20170308-162728
screenshot_20170308-162724

@the-dagger @mananwason Please review! :+1:


harshitagupta30:
@arp95 Thanks for your opinion. The UI was designed in such a way that the item is having a description field between the title and location and it looks good in that way. It's no harm to change the field to a non null field instead of null description. I hope its okay with you to let mentors decide if its necessary or not :) :+1:

arp95:
@harshitagupta30 
well why i was disapproving was according to the commit guidelines "fix" stands for a bug fix. I think this is not a bug fix going on. Use of a relevant commit message of "whatever" this pr is solving will gice the mentors a better idea.
Cheers!

harshitagupta30:
@arp95 Alright! Thanks for suggestion! I'll update the commit message :+1:

PR #1274, opened 2017-03-09
Images compressed before being downloaded.

### LINKED ISSUE #1264: Size of Speaker Images to very large.
**Actual Behaviour**
Just installed the app on my device and after 10mins of use the app has used 300MB of mobile data. 
The size of speaker images is the major reason behind this.
Size of some images is as high as 6.2MB. (`Speaker` - Blair Layton, `Image` - https://storage.googleapis.com/eventyay.com/events/6/speakers/2312/photo/TXZMK1JFUm/f3069efd-8cdf-4786-9e59-046f95f41440.png)

**Expected Behaviour**
The images must be compressed using `tinypng` or `pngcrusher`. I guess the same has been done in the project `2017.fossasia.org`. We can also reduce size of Sponsor images.

Yes

free4murad:
Fixes #1264 

Changes: Before being downloaded the images are compressed.

Screenshots for the change: 
Data used to download images before changes
before
Data used to download images after changes
after


free4murad:
@the-dagger @mananwason 
please review.

android/app/src/main/java/org/fossasia/openevent/utils/CompressImage.java
```diff
@@ -0,0 +1,30 @@
...
+        String temp = "";
+        int indexOfPath = -1;
+        try {
+            url = new URL(file);
+        } catch (MalformedURLException e) {
+            return file;
+        }
+        indexOfPath = file.indexOf(url.getPath());
+        if(!url.getHost().endsWith(".rsz.io")) {
+            temp = file.substring(0, indexOfPath) + ".rsz.io" + file.substring(indexOfPath) + "?width=300";
```
harshitagupta30: @free4murad Can we lower down the width in turn size a little more? I tried with Width=200 and found the images are good to work with :+1:  (Just a suggestion) :)

android/app/src/main/java/org/fossasia/openevent/utils/CompressImage.java
```diff
@@ -0,0 +1,30 @@
...
+            file = temp;
+        }
+        if(url.getProtocol().equals("https")){
+            file = temp.replace("https", "http");
```
harshitagupta30: Can you please add the `https` & `html` in strings.xml instead of hardcoding. :+1: 
Also `.rsz.io` :)

android/app/src/main/java/org/fossasia/openevent/utils/CompressImage.java
free4murad: Sorry for the late reply just saw this.
Actually we have to keep in mind the large devices like tabs.
Also the size difference between images with width 200 and 300 is very less.

android/app/src/main/java/org/fossasia/openevent/utils/CompressImage.java
free4murad: Sure

free4murad: @the-dagger @mananwason @harshitagupta30 
To use string resources in `CompressImage.java` we need to pass context of the activity as a parameter.
It will require a editing functions at many levels. 
1. Activity classes 2. Adapter classes 3. Speaker class.
We can totally do this but I think its not necessary.
These strings need not be translated.

What do you guys think?

free4murad:
@mariobehling 
I will look for some other method and update the PR.

Issue #1269, opened 2017-03-08
Make DB interactions asynchronous using loaders or Rx


iamareebjamal:
**Actual Behaviour**

UI is very jittery and lags on transitions from activity to activity due to heavy initialization on main thread and specifically in onCreate. 

**Expected Behaviour**

We could use Loaders or RxJava to make these tasks asynchronous. However our database doesn't change in real time and CursorLoaders are de-facto standard in Android for DB interactions, I think integration with Rx makes it very easy to scale the application in future and also makes it quite more testable using side effects. Using Rx also means we won't have to change our data model a lot, just wrap it around and we will be good. Almost all core db interactions reside in `DatabaseOperations.java`, even the internal handling of cursors, which will make implementing Loaders somewhat tedious than Rx
I think it will be a right step in improving the performance of the app.

**Would you like to work on the issue?**

Yes, I would like to work on the issue

free4murad:
@iamareebjamal 
because of this heavy initialization the app is missing the 16ms rendering window.

iamareebjamal:
@free4murad Yes, database operations are generally slow and onCreate in each view in generally monolithic. Migration to realm DB would be very invasive, so Rx integration will not only prove great for performance but will also help in creating complex asynchronous patterns in event handling for future scope

arp95:
looks good. I think this can make the app more fast.

iamareebjamal:
@the-dagger Should I start work on this?

the-dagger:
Yes, without a doubt! :+1:

iamareebjamal:
I think we should also include retrofit as it plays nicely with RxJava, or should I defer it?

PR #1295, opened 2017-03-13
Make all remaining fragments reactive

iamareebjamal:
Changes: 
- Following up from #1269 
- Make all DB interactions in fragments reactive
- Fix navigation drawer lag


iamareebjamal:
Please review @the-dagger @mananwason @mariobehling

PR #1305, opened 2017-03-14
fix: Remove redundant menu item tinting code

### LINKED ISSUE #1304: Remove redundant menu icon tinting in SpeakerDetailsActivity
**Actual Behaviour**

Share and back menu icon are being tinted unnecessarily in the SpeakerDetailsActivity

**Expected Behaviour**

We now use a shadow scrim to increase visibility of the icons and hence it is not required to tint them. Furthermore, the tinting is redundant currently and they are only tinted white. Also, due to reactive loading of Speaker Details, it may produce NullPointerException as activity lifecycle is not in our hands

Yes

iamareebjamal:
Now, we use a shaded scrim to make text and icons visible, so there is no need to dynamically tint menu icons. Also, reactively loading speaker from db means it may result in NullPointerException

Fixes #1304 

Changes: 
- Remove menu icon tinting from SpeakerDetailsActivity


harshitagupta30:
@iamareebjamal The code for tinting the image is not redundant. It is there because for some speaker images the name is not visible due to which tint has been added to make it visible #1141. You yourself added the code I guess.

iamareebjamal:
@harshitagupta30 That is tinting card headers, the code to tint the menu icons is redundant as it only tints them white and we already use scrim for visibility.

PR #1314, opened 2017-03-15
fix: display cached sponser images when offline

arp95:
Fixes #1287 

Changes: 

displaying cached sponser images when offline.

Screenshots for the change: 

17349144_10210746698738056_1311922091_o

PR #1313, opened 2017-03-15
fix: resolved crashing in SpeakerDetailsActivity

arp95:
Fixes #1312 

Changes:

Resolved crashing and now the app is working fine.

PR #1285, opened 2017-03-12
fix: loading of SpeakerImage in SpeakerDetailsActivity resolved

arp95:
Fixes #1277 

Changes: 
Added a network listener to listen for network change and consequently display speaker image when user changes the network settings.

Screenshots for the change: 

17237184_10210723772084904_1968310279_o

17310798_10210723774924975_586314479_o

Issue #1310, opened 2017-03-14
Prepare for release


the-dagger:
@mananwason @championswimmer we need to merge devel into master and publish the apk by today.

Some things which we might look into before doing this : 

- Test devel with the merged commits to see if it has any bugs
- Merge devel to master
- Update GPlay and fDroid variants

iamareebjamal:
There's a broadcast leak in `SpeakerDetailsActivity`, but no crash ->

```
Activity org.fossasia.openevent.activities.SpeakerDetailsActivity has leaked IntentReceiver org.fossasia.openevent.utils.NetworkUtils@be764d3 that was originally registered here. Are you missing a call to unregisterReceiver()?
android.app.IntentReceiverLeaked: Activity org.fossasia.openevent.activities.SpeakerDetailsActivity has leaked IntentReceiver org.fossasia.openevent.utils.NetworkUtils@be764d3 that was originally registered here. Are you missing a call to unregisterReceiver()?```

I am working to fix it

iamareebjamal:
An issue exists in Speaker Detail : #1321

championswimmer:
@the-dagger @mananwason have you physically tried out app on devel branch on your phones ?

mananwason:
Yes @championswimmer. I did try it on my phone, works fine for me right now. Let me upload another apk into beta using generator.

the-dagger:
Works well for me too. 👍
The generator has some issues tho, with package being corrupt. 
I've asked niranjan and he said he'll take a look at it.

PR #1320, opened 2017-03-15
Leaks, blocks and NPE fixes

iamareebjamal:
Changes:
- Fix memory and receiver leaks in SpeakerDetailActivity
- Change blocking nature of Picasso cache in #1314 
- Change Network listener to already implemented Network change event #1285 and #1313


iamareebjamal:
For #1310 

An issue in SpeakerDetailActivity exists

PR #1322, opened 2017-03-15
fix: resolved bug in BookmarksFragment

### LINKED ISSUE #1319: Bug in the BookmarksFragment
**Actual Behaviour**

On removing all the bookmarks from the bookmarksFragment we observe a behaviour as described in the screenshot below.

**Expected Behaviour**

Loading the TracksFragment when the bookmarksFragment becomes empty

**Steps to reproduce it**

1. Add two to three bookmarks 
2. Now remove all the bookmarks from the bookmarksFragment 
3. You would observe an empty bookmarksFragment which should be replaced by TracksFragment

**LogCat for the issue**

-

**Screenshots of the issue**

17237036_10210748348259293_1129547458_o

Yes.

arp95:
Fixes #1319 

Changes:
Instead of displaying an empty screen the user is taken to a TracksFragment.

Screenshots for the change: 

17357461_10210748464782206_1202725976_o
17357448_10210748464862208_1355164863_o


android/app/src/main/java/org/fossasia/openevent/fragments/BookmarksFragment.java
```diff
@@ -80,24 +82,49 @@ public void accept(@NonNull ArrayList<Integer> ids) throws Exception {
...
-                                    final int index = i;
-                                    dbSingleton.getSessionByIdObservable(id)
-                                            .subscribe(new Consumer<Session>() {
-                                                @Override
-                                                public void accept(@NonNull Session session) throws Exception {
-                                                    mSessions.add(session);
+                                if(bookmarkedIds.size()>0) {
 
-                                                    sessionsListAdapter.notifyItemInserted(index);
+                                    for (int i = 0; i < bookmarkedIds.size(); i++) {
```
the-dagger: isn't there a simpler way to do this?
I think that refreshing the adapter that holds bookmarks should be more than enough.

arp95: @the-dagger
actually when the user removes the bookmakrs one by one and then no bookmakrs are left, i am presenting a " no bookmakrs available" dialog to the user and then replacing bookmarks fragment with tracks fragment.
i didnt know whther you are referring to this or not.

mananwason:
IMO not required. The current system seems pretty decent.

PR #1335, opened 2017-03-16
fix: Re-enable hardware acceleration for shadows

iamareebjamal:
Changes: 
- Re-enable hardware acceleration for enabling shadows on API >= 21

The `android:hardwareAccelerated="true"` tag disables shadows and corner radius for cards, toolbars, FABs, etc and also introduces lag where fluidity could be possible

Screenshots for the change: 
Before:

After:


the-dagger:
@mananwason merge this too before releasing.

free4murad:
@iamareebjamal Nice
I also felt that the UI has changed.
I was just going to create an issue for this.

PR #1388, opened 2017-04-01
feat: display of user location using LocationManager

### LINKED ISSUE #1386: Getting user location for Map Activity
**Actual Behaviour**

Right now the Map Activity doesnt have useful features which can provide more interaction with users.

**Expected Behaviour**

Currently I want to get the users location to use that ahead for many applications in Map Activity.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

@the-dagger 
please review this. and also instead of using the variant i think in this case i have to use the google play services plugin which requires a google-services.json file. How to go about this file so that i can work with my location?

arp95:
Fixes #1386 

Changes: 

Used Location Manager to display user location which checks for network as well as gps connection for finding the users current location.

Screenshots for the change: 

-

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -6,6 +6,9 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Rect;
+import android.location.Location;
+import android.location.LocationListener;
+import android.location.LocationManager;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -217,8 +306,13 @@ protected void onCreate(Bundle savedInstanceState) {
         setUpProgressBar();
         setUpCustomTab();
 
+        getLocation();
         disposable = new CompositeDisposable();
 
+        if(latitude!=0.0 && longitude!=0.0){
+            Toast.makeText(this, "Location is " + latitude + " " + longitude, Toast.LENGTH_LONG).show();
```
the-dagger: Why are you displaying a toast to the user?

arp95:
@the-dagger
right now i wanted to get the method of taking location approved. now i plan on using the location in different use cases if approved.

the-dagger:
@arp95 
Closing this for now, we can probably follow up on this later on.
Right now, getting user location isn't a top priority so we can defer it for later.

arp95:
@the-dagger 
sure! No problem.

PR #1417, opened 2017-04-15
fix : search when sort option is changed

### LINKED ISSUE #1416: Search is not working when sort option is changed
**Actual Behaviour**

Search is not working when sort option is changed in Day Schedule Fragment and Speakers Fragment

**Expected Behaviour**

When changing sort option it should filter sessions/speakers according to search query.

**Steps to reproduce it**

- Go to ScheduleFragment/SpeakersFragment
- Enter search query
- Now change search option

**Screenshots of the issue**

Actual Behaviour
actual behaviour

Expected Behaviour
expected behaviour

Yes i would like to work on this.

Shailesh351:
Fixes #1416 

Fixes search results when sort option is changed in Day Schedule Fragment and Speakers Fragment

**Changes:**
Made separate method refreshAdapter() for refreshing adapter according to result, so that it can be reused In both Day Schedule Fragment and Speakers Fragment.

**Screenshots for the change:**
Before
actual behaviour

After
expected behaviour


harshitagupta30:
@shailesh351 please try to avoid writing fix twice in the PR. You can simply write fix: search when sort option is changed which makes sense in itself. Kindly change it. :+1:

the-dagger:
@Shailesh351 could you resolve the conflicts?

the-dagger:
@Shailesh351 
Please send the PR for this issue again, there seems to be some issue with this PR.
On resolving the conflicts, the travis builds are failing.

PR #1422, opened 2017-04-17
fix : Display thumbnails and sponsor images when app is offline

### LINKED ISSUE #1421: Speaker's thumbnail and sponsors's image are not displaying when app is offline.
**Actual Behaviour**

Speaker's thumbnail and sponsors's image are not displaying when app is offline. Speaker's thumbnail and sponsors's image are getting reloaded  when user reopens app after closing. It sends request to server for thumbnails each time, so more data is being used. So when there is no internet it is not displaying thumbnails and images.

**Expected Behaviour**

Display speaker's thumbnails and sponsor's images even if there is no internet by providing more cache.

**Steps to reproduce it**

Turn off internet connection and go to SpeakersFragment and SponsorsFragment.

**Screenshots of the issue**
Note: There is no internet connection

before_speaker

before_sponser

Yes, i would like to work on this.

Shailesh351:
Fixes #1421 

**Changes:**
Display thumbnails and sponsor images when app is offline by providing more cache.
Added PicassoCache.java class. Which is used for loading images along with storing it in app specific cache. 
This class should be used for displaying thumbnails and small size images because it stores images in cache.

**Screenshots for the change:** 
Note: There is no internet connection.

after_speaker

after_sponsor


Shailesh351:
@the-dagger please review  : )

android/app/src/main/java/org/fossasia/openevent/adapters/SessionSpeakerListAdapter.java
```diff
@@ -32,10 +31,12 @@
 public class SessionSpeakerListAdapter extends BaseRVAdapter<Speaker, SessionSpeakerListAdapter.RecyclerViewHolder> {
 
     private Activity activity;
+    private PicassoCache picassoCache;
 
     public SessionSpeakerListAdapter(List<Speaker> speakers, Activity activity) {
         super(speakers);
         this.activity = activity;
+        this.picassoCache = new PicassoCache(activity.getApplicationContext());
```
the-dagger: Do we really need the `applicationContext()`?

android/app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
```diff
@@ -76,6 +76,7 @@ protected void publishResults(CharSequence constraint, FilterResults results) {
     public SpeakersListAdapter(List<Speaker> speakers, Activity activity) {
         super(speakers);
         this.activity = activity;
+        this.picassoCache = new PicassoCache(activity.getApplicationContext());
```
the-dagger: Same as above.

android/app/src/main/java/org/fossasia/openevent/adapters/SessionSpeakerListAdapter.java
the-dagger: This might increase the chances of memory leaks in the app.

android/app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
Shailesh351: I am using application context for building picasso object. So it is needed. For avoiding memory leaks i will make weak reference for context.

From this [blog](https://android-developers.googleblog.com/2009/01/avoiding-memory-leaks.html) application context is more safer than activity context :+1:

Shailesh351: I went through [guide1](https://medium.com/freenet-engineering/memory-leaks-in-android-identify-treat-and-avoid-d0b1233acc8),  [guide2](https://possiblemobile.com/2013/06/context/) for memory leak in android. As we are not making **static** instance of context it will not leak memory. I checked it in HPROF Viewer and Analyzer in android studio.

Tell me if i am wrong at some point and suggest better way to do this.

iamareebjamal: You can initialize PicassoCache in Application class and then directly use static instance of it

the-dagger: Yeah, this makes sense. ^

iamareebjamal:
@Shailesh351 It would be better if you make the PicassoCache singleton, otherwise, this works too.

Singleton because the PicassoCache is not being changed according to different Context in different activities and fragments, so it would only need one instance of PicassoCache throughout the application

mariobehling:
@Shailesh351 Please help to close this issue and make the suggested changes. Thanks.

android/app/src/main/java/org/fossasia/openevent/cache/PicassoCache.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public Picasso getPicassoInstance(Context context) {
+
+        if (picassoInstance == null) {
+            synchronized (Picasso.class) {
+                if (picassoInstance == null) {
+                    new PicassoCache(context);
+                }
+            }
+        }
```
betterclever: Simply returning Picasso instance will work here since you initialized it in the constructor. PicassoInstance will never be null. It would have made sense if it would have been a static method where you would have returned a new instance of Picasso by creating a new PicassoCache object.

betterclever: You should make it static method instead.

android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
```diff
@@ -97,6 +107,9 @@ public void onCreate() {
             Picasso picasso = new Picasso.Builder(this).downloader(new OkHttp3Downloader(httpClient)).build();
             Picasso.setSingletonInstance(picasso);
```
betterclever: Best approach would be to pass picassoInstance made using PicassoCache in the Picasso.setSingletonInstance method. This would make sure every time that instance without changing code everywhere in application.

PR #1488, opened 2017-05-06
fix:Settings Page Animation fixed and Settings Descriptions Added(#1444)

### LINKED ISSUE #1444: Improve settings page and standard settings
* There is a delaying effect when the settings move in. Please take out any delays. We don't want the app to appear slow. No delays, please.
* Settings moves in, but "splashes out". Expected settings move in and move out without delay.
* App should not use "Mobile Data" should be set as a standard setting.
* It is unclear what the "notification mode" is about. Please add description about notifications. What notifications are there?

shreyasnbhat:
Fixes #1444 

Changes: default anim files were changed to give a shortConfigTIme and the Handler for the intent was removed.

Screenshots for the change: 
Already in Issue

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -596,13 +596,7 @@ public class MainActivity extends BaseActivity {
                 break;
             case R.id.nav_settings:
                 final Intent intent = new Intent(MainActivity.this, SettingsActivity.class);
-                smoothActionBarToggle.runWhenIdle(new Runnable() {
-                    @Override
-                    public void run() {
-                        startActivity(intent);
-                        overridePendingTransition(R.anim.slide_in_right, R.anim.slide_out_left);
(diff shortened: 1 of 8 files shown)


the-dagger:
@shreyasnbhat 
Did you ensure that changing the default download to mobile data ensures data being downloaded properly even after a clean install?
I remember having some issues with this, hence we forced the app to use wifi by default.

shreyasnbhat:
@the-dagger After a clean install the default settings are programmed in the settings.xml show up i.e WiFi for this setting and this is done only once. If the user changes it once the settings are saved.

shreyasnbhat:
`PreferenceManager.setDefaultValues(this,R.xml.settings,false);`
This is the line that does the one-time check of the settings prefs.
Earlier all the default values were being ignored.

the-dagger:
Ok, alright.
Yeah that makes sense, anyways I don't think that this PR will have to do anything with the issue I was mentioning anyways, that was a separate issue. 👍

PR #1509, opened 2017-05-08
feat: Migrate to Realm

iamareebjamal:
Fixes #1466

PR #1511, opened 2017-05-08
fix : filter in schedule fragment

### LINKED ISSUE #1510: Filter is not working properly in Schedule Fragment
**Actual Behaviour**

Filter is not working properly in Schedule fragment. 

**Expected Behaviour**

It should work properly when switching between pages.

**Screenshots of the issue**
videotogif_2017 05 08_13 34 30

yes

Shailesh351:
Fixes #1510 

Changes: 
Fixed filtering  issue in schedule fragment


Shailesh351:
@arp95 i think memory leak is not due to this. it is because of swipe refresh layout on refresh listener

leak

You can continue working on resolving memory leak

arp95:
@Shailesh351 
atleast before commenting get your stuff right.
have you viewed what we are doing in SwipeRefreshLayout listener?

Here for your reference:

swipeRefreshLayout.setOnRefreshListener(new SwipeRefreshLayout.OnRefreshListener() {
            @Override
            public void onRefresh() {
                refresh();
                if(mSelectedTracks != null && mTracksNames != null) {
                    filter(mTracksNames,mSelectedTracks);
                }
            }
        });

We are calling filter function inside this which is a source of memory leak.

Shailesh351:
@mahikaw Memory leak is still there

iamareebjamal:
I think the current implementation of filter is very poor and causes memory leaks here and there

I tried numerous ways of fixing it while implementing Realm but to no avail

That's why I have also delegated the filtering process in that PR. I think someone should remove and reimplement it from scratch

Shailesh351:
@iamareebjamal @arp95 I  checkedout version of app before filtere feature was added . It is also producing memory leak

iamareebjamal:
True @Shailesh351 

I also thought the same, what we can do for now is to make LeakCanary silent as I think it is a bug in FragmentPagerAdapter in the Android SDK

You can also search for memory leak in view pager adapter. It'll nevertheless be great if we fix it

iamareebjamal:
@Shailesh351 Can you please disable leak check in DayScheduleFragment and send PR for it?

mahikaw:
The memory leak issue has been there before and after pr #1498 . Incase, it hasn't been resolved yet then I think atleast we should get the filter working while we look for a fix to the memory leak problem. @iamareebjamal it will be nice if you could simply revert the PR #1498 as the filter stopped working after that.

arp95:
i think the filter problem existed from before.

mahikaw:
Um, I doubt actually. I remember coming across this bug after this PR got merged. Could you please recheck? @arp95

iamareebjamal:
@mahikaw The current filter is very poor implementation IMO and should be reimplemented as currently it is following several bad practices, reverting the PR would only deviate from the real issue. It'd be great if someone thinks about a better implementation

It is true that memory leak was before the filter was added, but currently filter  -
- Keeps track names in array for filtering - bad
- Forces load on many unneeded situations
- Does not play nice with sorting

Shailesh351:
Filtering will be fixed when PR #1509 will merged. So closing it. 

For avoiding memory leak I think we have to re-implement schedule fragment and  make better model for it so that searching , filtering all works simultaniously

PR #1513, opened 2017-05-08
fix : App crashing when connected to network but internet is not working

### LINKED ISSUE #1479: App crashes when device connected to network but internet is not working in DayScheduleFragment
**Actual Behaviour**

App crashes when device connected to Mobile Data but internet is not working in DayScheduleFragment because of  NullPointerException.

**Expected Behaviour**

App shouldn't crash.

**LogCat for the issue**

`java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.view.View.getResources()' on a null object reference
                                                                            at android.support.design.widget.Snackbar.make(Snackbar.java:165)
                                                                            at org.fossasia.openevent.utils.ShowNotificationSnackBar.showSnackBar(ShowNotificationSnackBar.java:41)
                                                                            at org.fossasia.openevent.utils.ShowNotificationSnackBar.<init>(ShowNotificationSnackBar.java:35)
                                                                            at org.fossasia.openevent.fragments.DayScheduleFragment$6$1.<init>(DayScheduleFragment.java:328)
                                                                            at org.fossasia.openevent.fragments.DayScheduleFragment$6.inactiveConnection(DayScheduleFragment.java:328)
                                                                            at org.fossasia.openevent.utils.NetworkUtils$4$1.accept(NetworkUtils.java:158)
                                                                            at org.fossasia.openevent.utils.NetworkUtils$4$1.accept(NetworkUtils.java:152)`

**Steps to reproduce**

- Run app & turn off wifi connect to network which has no internet
- Go to schedule fragment
- Refresh schedule using swipe refresh layout

Yes

Shailesh351:
Fixes #1479 

Changes: 

Check if view is null when calling snackbar.make() and if view is null use swipeRefreshLayout, because view in make method is used for finding suitable parent view (CoordinatorLayout).
So it will use swipeRefreshLayout for finding suitable parent if view is null.

Reference : https://developer.android.com/reference/android/support/design/widget/Snackbar.html#make(android.view.View,int,int)


mananwason:
Please check my comments on the issue

android/app/src/main/java/org/fossasia/openevent/utils/ShowNotificationSnackBar.java
```diff
@@ -38,18 +38,31 @@ public ShowNotificationSnackBar(Context context, View view, SwipeRefreshLayout s
...
+                            }
+                            snackbar.dismiss();
+                            refreshClicked();
                         }
-                        snackbar.dismiss();
-                        refreshClicked();
-                    }
-                });
+                    });
+        }else{
```
iamareebjamal: Don't duplicate code

Create a local `View` variable and assign it any non null view instance and pass that as an argument to the Snackbar method

Shailesh351:
Yes i'm not able to reproduce the bug. But from the log it is clearly because view is null in showSnackbar() method. So it will be good to check if it is null or not.

iamareebjamal:
But the problem is. If one view is null, the other should be too

So changing it to some other view is also unneeded IMO

iamareebjamal:
Better than this, we should analyse why the view is null -> meaning there is some activity or fragment passing it which gets destroyed making the view null. The actual problem is that we are retaining a dead view. This can happen due to not removing a callback on destroy.

See the usage of this function and try to determine which such uncleared function is calling it

arp95:
we have to ensure whether there can be a point if that view will be null and being called up. i dont think this is happening.

iamareebjamal:
@arp95 Yes there can be such case

Register network listener in home fragment and don't remove it

Go to BookmarksFragment -> HomeFragment destroyed but held in memory by the listener (but views are null and it is a memory leak)

Callback returns, tries to call on view -> NPE

Shailesh351:
View parameter passed in snackbar's make methode is root view of the fragment returned by  getView() method.

I just faced issue right now here is log cat:
`05-23 15:48:07.725 11381-11381/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                   Process: org.fossasia.openevent, PID: 11381
                                                   java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.view.View.getResources()' on a null object reference
                                                       at android.support.design.widget.Snackbar.make(Snackbar.java:170)
                                                       at org.fossasia.openevent.utils.ShowNotificationSnackBar.showSnackBar(ShowNotificationSnackBar.java:43)
                                                       at org.fossasia.openevent.utils.ShowNotificationSnackBar.<init>(ShowNotificationSnackBar.java:35)
                                                       at org.fossasia.openevent.fragments.DayScheduleFragment$6$1.<init>(DayScheduleFragment.java:321)
                                                       at org.fossasia.openevent.fragments.DayScheduleFragment$6.inactiveConnection(DayScheduleFragment.java:321)
                                                       at org.fossasia.openevent.utils.NetworkUtils$4$1.accept(NetworkUtils.java:158)
                                                       at org.fossasia.openevent.utils.NetworkUtils$4$1.accept(NetworkUtils.java:152)
                                                       at io.reactivex.internal.observers.ConsumerSingleObserver.onSuccess(ConsumerSingleObserver.java:59)
                                                       at io.reactivex.internal.operators.single.SingleObserveOn$ObserveOnSingleObserver.run(SingleObserveOn.java:81)
                                                       at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:109)
                                                       at android.os.Handler.handleCallback(Handler.java:751)
                                                       at android.os.Handler.dispatchMessage(Handler.java:95)
                                                       at android.os.Looper.loop(Looper.java:154)
                                                       at android.app.ActivityThread.main(ActivityThread.java:6123)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)`

Step to reproduce :

- Run app &  turn off wifi connect to network which has no internet
- Go to schedule fragment
- Refresh schedule using swipe refresh layout

arp95:
@iamareebjamal  yeah that will produce this thing. what i meant was is it possible that a user while using the app might face this.

iamareebjamal:
Yes it is. I have faced it too

@Shailesh351 The easiest way to debug this would be to add different log messages in different callers of this method and then check when it happens next time that which one is the culprit

Shailesh351:
@iamareebjamal View parameter passed in snackbar's make method is view returned by getView() method which is method of Fragment class, it returns root view of fragment returned by onCreateView().

View parameter in snackbar's make() method is used to find suitable parent i.e. CoordinatorLayout. So we can use swipe refresh layout for finding suitable parent. And it is guaranteed that swipe refresh layout is not null because refresh method is called when swiping.

iamareebjamal:
OK @Shailesh351 then only use swiperefreshlayout to show the snackbar

iamareebjamal:
@Shailesh351 Let's merge this. I too am getting this error a lot. But please, add a check that if swiperefreshlayout is null, we should return and add a log in that case too

Shailesh351:
@iamareebjamal made changes. Can't directly return. Because function is returning snackbar object and it won't be good to return null.

iamareebjamal:
Well, then look at the usages of this function and see if all of them immediately call show, if so make the function void and call show inside it, if not, then add null checks around the function call. BTW it is a memory leak that this error is happening

iamareebjamal:
@Shailesh351 was this the only usage of this function? :astonished: 

I wonder why it was transferred to another file!

Shailesh351:
@iamareebjamal No the function is used in all fragments but only tracks fragment was using object returned by it. And also it don't make sense to dismiss snakbar manually it will disappear automatically after timeout. So I made return type void and removed snack bar from tracks fragment

PR #1514, opened 2017-05-08
share button updated

### LINKED ISSUE #1493: App sharing via Facebook
**Actual Behaviour**

Toast saying "Please retry again" is raised when I try to share the app using "INVITE VIA FACEBOOK" button. 
On the other hand, I am able to share the app successfully via Facebook using "SHARE APP" button and then selecting "share with Facebook".

**Expected Behaviour**

I think there is no need for a separate button to enable Facebook sharing. Moreover, it is not even functional.

**Screenshots of the issue**

device-2017-05-06-170435

device-2017-05-06-170458

Yes.

-----
@the-dagger @mariobehling @mananwason Is this intended?

mahikaw:
Fixes #1493
Removed "sharing with Facebook" button and restructured "sharing with app".


android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -607,7 +607,7 @@ public void accept(@NonNull Boolean empty) throws Exception {
                 LayoutInflater inflater = this.getLayoutInflater();
                 View dialogView = inflater.inflate(R.layout.dialog_share, mainFrame, false);
                 Button shareApp = (Button)dialogView.findViewById(R.id.share_app);
-                Button facebookInvite = (Button)dialogView.findViewById(R.id.facebook_invite);
+//                Button facebookInvite = (Button)dialogView.findViewById(R.id.facebook_invite);
```
the-dagger: Why comment this out?
Delete it instead.

mahikaw:
@the-dagger people could share the app on fb and twitter as well. Do we really need to specifically remove them? just saying..

the-dagger:
@mahikaw I don't think that you need the fb sdk for sharing something on fb.
You can simply share a URL and select Facebook from the app chooser and it'll share the content on fb.

However the SDK might have been used for app invite using fb, could you check that once and if not then we can remove it.

iamareebjamal:
@the-dagger Yes, the SDK is used to invite the user. But is the feature really needed?

the-dagger:
No, not really.
I got a similar feedback at the summit too.

iamareebjamal:
Then we should remove the SDK as it adds to the app size unnecessarily

the-dagger:
Yeah, that's what I mentioned in the first place.

mahikaw:
@the-dagger @iamareebjamal Sure,  I'm looking into this. 👍

mahikaw:
I'm creating a new PR.

PR #1520, opened 2017-05-08
fix: Network not available dialog disappears when network is available

### LINKED ISSUE #1505: Network not available dialog box does not go when network is available
**Actual Behaviour**

If there is no network, then the network not available dialog box is displayed. When the network state changes the dialog box is still displayed.

**Expected Behaviour**

When network is available the dialog box should disappear.

**Steps to reproduce it**

Change network state while using the app.

I would like to work on it if the actual behaviour is not intended

mayank8318:
Fixes #1505
Changes: 
- The network not available dialog box now disappears when network is available.
- Create a static Dialog object in MainActivity which maintains a reference to the dialog created.
- dismiss this dialog in OpenEvent if connection is available and the dialog is not null.


android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -67,6 +68,7 @@
 import org.fossasia.openevent.data.Track;
 import org.fossasia.openevent.dbutils.DataDownloadManager;
 import org.fossasia.openevent.dbutils.DbSingleton;
+import org.fossasia.openevent.events.ConnectionCheckEvent;
```
mayank8318: Solved!

android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
```diff
@@ -67,7 +68,7 @@ public static void postEventOnUIThread(final Object event) {
         handler.post(new Runnable() {
             @Override
             public void run() {
-                getEventBus().post(event);
+                    getEventBus().post(event);
```
iamareebjamal: Please indent this correctly

iamareebjamal:
And please squash the commits too.

And one suggestion, if you are using static dialog, name it something specific rather than just dialog

mananwason:
@the-dagger Please review thoroughly before merging. There are 2 merge branch commits as well in the PR. @mayank8318 You should pull code using `` git pull --rebase `` to stop git from add a merge commit.

Issue #1448, opened 2017-04-30
Unscheduled Session show up at 0:00


mariobehling:
Unscheduled sessions show up in the schedule at 0:00. Expected: Do not show unscheduled sessions at all. 

Ensure unscheduled sessions show up in the API accordingly and align with the event orga server. Do not just mark a session "unscheduled" if it starts at 0:00. It is possible that there are sessions at that time actually. Please take care to details in this issue.

heysadboy:
@mariobehling Can I work on this?

harshitagupta30:
@heysadboy go ahead :+1:

heysadboy:
@mariobehling @harshitagupta30 Hey, events at 00:00 shows up like this 


![screenshot_2017-05-05-20-05-46](https://cloud.githubusercontent.com/assets/22588035/25750687/182700f8-31d0-11e7-98e3-fb47e6e6ff83.png)


I am not able to reproduce this bug. Am I missing something? Is this the expected behavior?

mananwason:
@heysadboy This is because events at 12 PM are decoded as 12 AM and rendered at 00 times. Take a look at ISO conversions.

heysadboy:
Okay, solving this here https://github.com/fossasia/open-event-android/pull/1538. Please see @mananwason .

heysadboy:
Target Milestone to solve the issue

https://github.com/fossasia/open-event-android/milestone/24

I can't edit this issue so putting this here.

PR #1538, opened 2017-05-10
Managing error in represenation of time

heysadboy:
Earlier it used to show 00:00 pm after 11:00 am. Showing time like 12:00 pm now. Solving this https://github.com/fossasia/open-event-android/issues/1448. See the screenshot to know more,

screenshot_2017-05-10-05-28-27


iamareebjamal:
Isn't there some other way than string manipulation to achieve that? 

Looks hacky

heysadboy:
@iamareebjamal If we do it in ISO conversions or anywhere else it would change the sorting order. This is the only anomaly I found so I though it would be better to just do it at the time of display.

android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
```diff
@@ -96,6 +96,15 @@ public void onBindViewHolder(DayScheduleViewHolder holder, int position) {
         String startTime = ISO8601Date.get12HourTime(ISO8601Date.getDateObject(currentSession.getStartTime()));
         String endTime = ISO8601Date.get12HourTime(ISO8601Date.getDateObject(currentSession.getEndTime()));
 
+        /*We are following 12 hour format so changing 00 PM to 12 PM */
```
Shailesh351: No need to manipulate string manually. You can just change 'KK' to 'hh'  in get12HourTime function in ISO8601Date.java
KK  is for 0-11 am/pm
hh is for 1-12 am/pm

Reference : https://developer.android.com/reference/java/text/SimpleDateFormat.html

heysadboy:
@Shailesh351 @iamareebjamal Thanks for reviewing, please see.

PR #1547, opened 2017-05-10
fix : wrong dates displayed in ScheduleFragment

### LINKED ISSUE #1541: Wrong dates displayed in DaySchedularFragment.
**Actual Behaviour**
If the user is using the app with the following configuration
screenshot_20170510-054441
then dates displayed in the `DaySchedularFragment` are incorrect.

eg when you open the sessions under the `16Jan` tab, the date displayed in the `SessionDetailActivity` for that session is 17 Jan.

**Expected Behaviour**
The dates in `DaySchedularFragment` and `SessionDetailActivity` should be same.

No

Shailesh351:
Fixes #1541 

Changes:

- Changed 'D' to 'd' because 'D' is for 'Day in year' & 'd' is for 'Day in month'  
- set event time zone on simple date formate
 
Screenshots for the change: 
fix-date


android/app/src/main/java/org/fossasia/openevent/utils/ISO8601Date.java
```diff
@@ -51,12 +51,12 @@ public static String dateFromCalendar(Calendar currentDate) {
...
 
     public static String getTimeZoneDateFromString(String dateString) throws ParseException {
-        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-DD", Locale.getDefault());
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd", Locale.getDefault());
         formatter.setTimeZone(getEventTimezone());
         Date date = formatter.parse(dateString);
-        SimpleDateFormat newFormat = new SimpleDateFormat("d MMM", Locale.getDefault());
-
-        return newFormat.format(date);
+        SimpleDateFormat newFormat = new SimpleDateFormat("d MMM, z", Locale.getDefault());
```
iamareebjamal: Is `z` needed?

It is again split out below, so...

Shailesh351: Yes i have checked it on my device. Without 'z' it is formatting date with "GMT+5:30" for both time zones that's why i am using 'z'

Just check it using Log in getTimeZoneDateFromString() function
`dateString +"  "+ date.toString()`

iamareebjamal: Ok. For an event based application, these date errors are crucial. I would have recommended Joda Time library but it's size is huge relative to the convenience it provides. What we can do is write tests. Currently, I am busy with exams so probably you or some other person can take up the issue to create unit tests for date conversion code?

Shailesh351:
Thanks @iamareebjamal @arp95

PR #1561, opened 2017-05-15
updated upload-apk.sh with second app

### LINKED ISSUE #1558: Adding Google IO sample as test app
**Actual Behaviour**

Sending pr for adding the google io version as a test app.

**Expected Behaviour**

-

**Steps to reproduce it**

-
**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes

arp95:
Fixes #1558 
Changes: 
Updated the script to add second app.

Screenshots for the change:


arp95:
@betterclever in this file itself?

arp95:
@betterclever do you mean adding another variant in the app like fdroid and googlePlay?

iamareebjamal:
The command you have written will have no effect since there is no apk of this form present in the build folder

arp95:
@iamareebjamal 
i have to ways either i put my apk in the apk folder 
or 
i create another variant of googleIO version

which one seems viable?

iamareebjamal:
I am not aware of what you are trying to do?

arp95:
@iamareebjamal 
actually i have created an apk for google io version. now either i put the apk in build/outputs/apk and then change the upload.sh file or create another variant in the build.gradle file.

i think simply adding the google io apk in the folder right?

iamareebjamal:
Even if you somehow create Google IO app variant, it won't be containing any of the info from Google IO unless you statically put it in that variant

Secondly if you put it manually on apk branch, it will be deleted on the next commit on master

arp95:
@iamareebjamal 
yeah this is what is making be confused.

Shailesh351:
@arp95 Try to generate App for Google_IO in same way it being generated for FOSSasia.
For this i think you have to write script for changing api url to google_io sample and then regenerate app with same functions.

arp95:
@Shailesh351
why write a script when the same can be done using a product flavour in the build.gradle file.
anyways lets see what @mariobehling @betterclever are looking for.

betterclever:
@iamareebjamal Please guide. I guess making a I/O version using a Product Flavor in build.gradle shouldn't be done. We shouldn't add an event specific variant there.
What should be best strategy?

iamareebjamal:
I think project mentors will be best suited to guide as I don't know what we're trying to achieve here 

@the-dagger @mananwason

arp95:
@iamareebjamal 
just to follow up you can read the chat on gitter yesterday. I think that will clear it up.

iamareebjamal:
OK thanks

Shailesh351:
@iamareebjamal Goal is too generate Two apks one for Fossasia sample and one for Google_IO sample after every push on development and transfer it to Apk branch. Currently apk for fossasia sample is being generated .So along with it we want to generate apk for Google_IO sample also

iamareebjamal:
That can only be properly done via product flavors, but it will contain sample data only, second way is to trigger a build from the generator on each push, which won't be possible from bash script but is the best possible way to do it. So if someone can make changes to the generator accepting command line options to include a URL endpoint or JSON payload, then the generator will build the app in travis on each push and you can rename it to google-io.apk and upload with debug app 

This is most robust solution, both for temporary reasons and future adaptability

arp95:
@iamareebjamal using product flavours should be a good option. now coming to the approach that it will be static, i think it wont be because in the google io version i am adding api link which will get updated with more info, so i think this shouldnt be an issue.

iamareebjamal:
OK. That should be fine if mentors give it a thumbs up. The only issue about product flavor is that it makes the app source event specific, other than that, it is good

mananwason:
@arp95 @iamareebjamal I don't understand the need for flavors. Are we making two different versions of the app? I thought we want to test the different flavors of the app with 2 different datasets. Am I missing something?

iamareebjamal:
@mananwason Mario said to make changes so that the test apk for Google I/O test apk is also uploaded to  apk branch on each push

mananwason:
@iamareebjamal That's kind of evident from the issue plus I have gone through the chat as well but I want to say is that this seems like a shortcut solution. Instead work on the generator scripts to fix this. 
This is not a college project where we can use any solution that works. We need to make sense of the solution as well.

arp95:
@mananwason i know the product flavour solution doesnt make sense because for every sample we would have to create another flavour which doesnt make sense. can you tell how to go about it for an efficient manner?

mananwason:
One solution would be what @iamareebjamal suggested. You can pass urls/zip through the command line to the script which can use the data and create the apk. Otherwise you can also work on an api for the generator which will be useful to integrate in the server as well. This api will take in all the parameters and return an apk.

iamareebjamal:
I forgot that we already have a API, the generator is just a front end wrapper to it. @arp95 Please use the API endpoint to send the data on to the hosted app generator and download the generated app.

You can create a small python script for it

the-dagger:
Closing this as the app for google io is no longer needed.
@arp95 you can maybe store the generated app along the samples instead?
And then update them as you update the samples.

PR #1563, opened 2017-05-15
fix: add sort option for org and country only when needed

### LINKED ISSUE #1559: Remove the sort by org option when all orgs are same
**Actual Behaviour**

When buidling sample app for google io event i encountered that in events like these all speakers can be from same org. In such cases enabling sorting by org doesnt mean anything.

**Expected Behaviour**

Removal of sorting by org in cases when orgs are same.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

arp95:
Fixes #1559 

Changes: 

Added sort options only if they are needed in the SpeakersDetailActivity.

Screenshots for the change: 

18516546_10211298370689510_776370621_o


arp95:
@iamareebjamal yeah repititions are happening.
what i can do is use a getter method in the SepakersListAdapter and us that to give me the orgs and country instead of running on the main thread. what say?

iamareebjamal:
You can do that, but instead whenever there is an update of list if speakers, you'll get that in the subscribe function of the observable. Trigger a function there to calculate the organisations and countries in 1 loop. and set it in a variable.

And when showing the dialog, use this variable to determine what should be the options

android/app/src/main/java/org/fossasia/openevent/fragments/SpeakersListFragment.java
```diff
@@ -172,9 +172,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case R.id.action_sort:
 
+                //checking for org and country
+                int[] array = DbSingleton.getInstance().getCount().get(0);
```
iamareebjamal: Better :clap: 

But here's the issue, it gets called every time the user clicks on menu and on the Main Thread!

Furthermore, you already have a list of speakers, why querying it again?

Instead of querying it again, calculate it in the subscribe method, where the speakers are loaded and set the corresponding state variables and then just check them while loading menu, like you have done here

android/app/src/main/java/org/fossasia/openevent/fragments/SpeakersListFragment.java
```diff
@@ -118,6 +120,7 @@ public void onRefresh() {
                     public void accept(@NonNull List<Speaker> speakers) throws Exception {
                         mSpeakers.clear();
                         mSpeakers.addAll(speakers);
+                        array = DbSingleton.getInstance().getCount(mSpeakers).get(0);
```
iamareebjamal: Again, you already have the list of `speakers` here, please don't query again

arp95: @iamareebjamal please check the method in DbSingleton.java
i am not querying the list, i am using the speaker list to get the count instead of over here. i think this should be fine.

iamareebjamal: But why querying database again when you have the list of speaker here?

iamareebjamal: Just count them here instead of going to the database again

Here : 20 ms
Database : 750 + 20 ms

(Figuratively)

iamareebjamal: Db class should be doing db work

arp95: sure i thought you said add that the count part over there. got it

arp95:
@iamareebjamal i think it should be fine now. anyways have a look.

android/app/src/main/res/values/arrays.xml
```diff
@@ -13,7 +13,18 @@
         <item>1 Hour</item>
         <item>12 Hour</item>
     </string-array>
-    <integer-array name="speaker_sort">
+    <integer-array name="speaker_sort1">
```
Shailesh351: Just give appropriate names like 

- speaker_sort_name
- speaker_sort_name_country
- speaker_sort_name_organisation
- speaker_sort_all

Shailesh351: Everything else looks good :+1:

mananwason:
@iamareebjamal @arp95 Is it necessary to perform so many computations? I suggest that you can just provide a default sorting option if the parameter is null.

arp95:
@mananwason i think its needed because many things we have been seeing just for the fossasia event. In cases when orgs are same or country are same we have to remove the sort option i think.

iamareebjamal:
@mananwason Besides, the computations will be less once Realm implementation has been merged, then we'll be able to easily query the amount of countries, orgs separately

mananwason:
@arp95 the problem is acceptable but the solution is not. You could have also just check for first few records if they contained the parameter then you could include it in sort options. If we get a large enough database, the app will start lagging from your solution.


@iamareebjamal Realm will not greatly improve performance in my opinion. The solution is flawed. A for loop on the ui thread is a definite NO!

arp95:
@mananwason do you mean i just check first 4-5 values and update according to that?

mananwason:
@arp95 Yes. Just of the top of my head, this will be a better solution. I'm sure you can find an even better one.

iamareebjamal:
@mananwason That's why I suggested RxJava to compute the results off the thread

iamareebjamal:
Or you can use SQL query to COUNT DISTINCT country attribute

arp95:
@mananwason @iamareebjamal is this approach fine? removed the for loop.

android/app/src/main/java/org/fossasia/openevent/fragments/SpeakersListFragment.java
```diff
@@ -118,6 +120,29 @@ public void onRefresh() {
                     public void accept(@NonNull List<Speaker> speakers) throws Exception {
                         mSpeakers.clear();
                         mSpeakers.addAll(speakers);
+                        List<Speaker> sorted_speakerCountry = dbSingleton.getSpeakerList("country");
+                        List<Speaker> sorted_speaker_org = dbSingleton.getSpeakerList("organisation");
```
iamareebjamal: If you are querying database on Main Thread, then it is worse

iamareebjamal: I think using using RxJava to compute countries and organisations off the main thread will be fine

arp95: yeah seeing the efficient solution over here.

arp95:
@mananwason @iamareebjamal i think this should be efficient.

android/app/src/main/java/org/fossasia/openevent/fragments/SpeakersListFragment.java
```diff
@@ -172,9 +174,25 @@ public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case R.id.action_sort:
 
+                int num_orgs=speakersListAdapter.getDistinctOrgs();
```
iamareebjamal: Please refrain from querying database on UI thread. Wrap inside an Observable like other examples from DbSingleton and use that to achieve this :smile:

android/app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
```diff
@@ -108,8 +111,27 @@ public RecyclerViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
     public void onBindViewHolder(RecyclerViewHolder holder, final int position) {
         final Speaker current = getItem(position);
 
-        String thumbnail = current.getThumbnail();
+        //adding distinct org and country (note size of array will never be greter than 2)
+        if(distinctOrgs.size()==0){
+            distinctOrgs.add(current.getOrganisation());
+        }else{
+
+            if(distinctOrgs.size()==1 && (!current.getOrganisation().equals(distinctOrgs.get(0)))){
```
iamareebjamal: Use else if please

Nice solution :+1:

android/app/src/main/java/org/fossasia/openevent/adapters/SpeakersListAdapter.java
```diff
@@ -108,8 +111,27 @@ public RecyclerViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
     public void onBindViewHolder(RecyclerViewHolder holder, final int position) {
         final Speaker current = getItem(position);
 
-        String thumbnail = current.getThumbnail();
+        //adding distinct org and country (note size of array will never be greter than 2)
+        if(distinctOrgs.size()==0){
```
iamareebjamal: Use `isEmpty()` instead, looks cleaner and is verbose

mahikaw:
@arp95 , using COUNT DISTINCT seems like a better approach in my opinion. Let the mentors make the final call though. @mananwason

mahikaw:
@arp95 , using COUNT DISTINCT as suggested by @iamareebjamal  seems like a better approach in my opinion. Let the mentors make the final call though. @mananwason

iamareebjamal:
@mahikaw When we merge Realm, SQL queries won't be available, and we'll have to resort o this methods only. Requiring breaking changes with merging. And this approach requires no computation, just a condition check on all holding and one assignment, which is much more efficient than a SQL query which may take several 100 milliseconds

mahikaw:
@iamareebjamal seems like I didn't think this through. Thanks for pointing it out. 👍

PR #1566, opened 2017-05-16
test: date tests added

### LINKED ISSUE #1549: Tests for Date conversion methods
We have had several issues regarding date formatting and timezone settings. This is a very crucial area for an event based application and thus must be ensured to be working correctly on all fronts. In order to do so, we must write unit tests so that no changes break the desired working of the application.

Tests should include:
- Correct conversion from ISO 8601 format to string representation
- Conformance to timezone setting

No. I am busy in preparations of my exams

arp95:
Fixes #1549 

Changes:

Added tests for date testing.

Screenshots for the change: 
-

### DIFF
--- a/android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
+++ b/android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
@@ -1,0 +1,356 @@
+package org.fossasia.openevent;
+
+import org.fossasia.openevent.utils.ISO8601Date;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Created by arpitdec5 on 02-06-2017.
+ */
+
+public class DateTest {
+
+    @Test
+    public void shouldConvertLocalTimeZoneDateStringToSpecifiedTimeZoneDateString() {
+        ISO8601Date.setTimeZone("UTC");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        //date strings to be used for testing
+        String dateString1 = "2017-06-02T07:59:10Z";
+        String dateString2 = "2017-06-02T20:00:10Z";
+        String dateString3 = "2017-06-03T00:00:10Z";
+        String dateString4 = "2017-06-03T07:00:10Z";
+        String expectedString;
+        String actualString;
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString1);
+        expectedString = "Thu, 01 Jun 2017, 23:59, UTC";
+        Assert.assertEquals(expectedString, actualString);
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString2);
+        expectedString = "Fri, 02 Jun 2017, 12:00, UTC";
+        Assert.assertEquals(expectedString, actualString);
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString3);
+        expectedString = "Fri, 02 Jun 2017, 16:00, UTC";
+        Assert.assertEquals(expectedString, actualString);
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString4);
+        expectedString = "Fri, 02 Jun 2017, 23:00, UTC";
+        Assert.assertEquals(expectedString, actualString);
+    }
+
+    @Test
+    public void shouldConvertLocalTimeZoneDateStringToInternationalTimeZoneDateString() {
+        //first case
+        ISO8601Date.setTimeZone("Europe/Amsterdam");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        String dateString = "2017-06-02T02:29:10Z";
+        String expectedString;
+        String actualString;
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Thu, 01 Jun 2017, 20:29, GMT+02:00";
+        Assert.assertEquals(expectedString, actualString);
+
+        //second case
+        ISO8601Date.setTimeZone("Asia/Kolkata");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Thu, 01 Jun 2017, 23:59, GMT+05:30";
+        Assert.assertEquals(expectedString, actualString);
+
+        //third case
+        ISO8601Date.setTimeZone("Europe/Berlin");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Thu, 01 Jun 2017, 20:29, GMT+02:00";
+        Assert.assertEquals(expectedString, actualString);
+
+        //fourth case
+        ISO8601Date.setTimeZone("Australia/Sydney");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Fri, 02 Jun 2017, 04:29, GMT+10:00";
+        Assert.assertEquals(expectedString, actualString);
+    }
+
+    @Test
+    public void shouldConvertInternationalTimeZoneDateStringToLocalTimeZoneDateString() {
+        //first case
+        ISO8601Date.setTimeZone("UTC");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+
+        String dateString = "2017-06-02T02:29:10Z";
+        String expectedString;
+        String actualString;
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Thu, 01 Jun 2017, 18:29, UTC";
+        Assert.assertEquals(expectedString, actualString);
+
+        //second case
+        ISO8601Date.setTimeZone("UTC");
+        ISO8601Date.setEventTimeZone("Europe/Amsterdam");
+
+        actualString = ISO8601Date.getTimeZoneDateStringFromString(dateString);
+        expectedString = "Fri, 02 Jun 2017, 00:29, UTC";
+        Assert.assertEquals(expectedString, actualString);
+
+        //third case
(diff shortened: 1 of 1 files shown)


arp95:
@iamareebjamal 
i still have to check the timezone stuff.but this is the way to go right?

iamareebjamal:
Correct

iamareebjamal:
@mahikaw More tests need to be added. This is not enough

mahikaw:
@iamareebjamal yes I'm aware. I did it by mistake and couldn't find a way to undo the review. Please ignore.

arp95:
yeah actually having my exams so not able to devote much time. working on tests part.

arp95:
@iamareebjamal why do not merge tag? tests are not fine?

iamareebjamal:
@arp95 For WIP PRs, do not merge tags tell maintainers to not try to merge the PR because it is work in progress. This saves time as they know they don't have to review it now

Someone may see check status and a correct test and merge the PR, so it's just a reminder, nothing bad for the author of PR. It doesn't reflect anything wrong with the PR

arp95:
@iamareebjamal no i am not taking anything personal. i made changes and then the label was added after that.

iamareebjamal:
Oh, I didn't know you had added more tests, let me take a look

Sorry

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
```diff
@@ -0,0 +1,59 @@
...
+            readDate.setTimeZone(TimeZone.getTimeZone("GMT+04:01"));
+            Date gmt_date_3 = readDate.parse(gmt_time);
+            Assert.assertEquals("12:00:02", gmt_date_3);
+
+            //checking base case 00:00:00
+            readDate.setTimeZone(TimeZone.getTimeZone("GMT+16:01"));
+            Date gmt_Date_4 = readDate.parse(gmt_time);
+            Assert.assertEquals("00:00:02", gmt_Date_4);
+
+        }catch (ParseException e){
```
iamareebjamal: Add a space after curly brace
`} catch (`

arp95: sorry for delaying this for so long. having a look

arp95: @iamareebjamal i am confused over here. we use ISO8601Date.java class to test the correctness. but each function needs getTimeZone() which if i get it, i will have to use if-else statements to check the output, like if timezone is ist do this etc?

iamareebjamal: OK. So what you can do is increase a parameter in each method, taking in a timezone and create helper method for others that by default call getTimezone()

This generally happens while testing

Second option is using Mockito, which returns what you want when getTimezone() is called, but it will be a little difficult, so use first method

arp95: you mean adding more methods in ISO8601Date.java where i am passing timezone as a parameter?

iamareebjamal: Yes, and then use that general method in other methods as a helper one

I'm talking about function overloading using different parameter list

arp95:
@iamareebjamal @Shailesh351 @mahikaw @shreyasnbhat please review.

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test;
+
+/**
+ * Created by arpitdec5 on 02-06-2017.
+ */
+
+public class DateTest {
+
+    @Test
+    public void checkDate() {
```
iamareebjamal: Ideally, there should be only one assert per test/function

Please break these into several tests and specify what it is expected to do
You can group similar assertions in one test

Look at this for reference - https://github.com/fossasia/open-event-orga-app/blob/master/app/src/test/java/org/fossasia/openevent/app/utils/UtilsTest.java

You can take a look at other test classes as well, there are plenty for reference

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
```diff
@@ -0,0 +1,87 @@
...
+        //test of method getTimeZoneDateStringFromString()
+        ISO8601Date.setTimeZone("UTC");
+        ISO8601Date.setEventTimeZone("Asia/Singapore");
+        String dateString1 = "2017-06-02T07:59:10Z";
+        String dateString2 = "2017-06-02T20:00:10Z";
+        String dateString3 = "2017-06-03T00:00:10Z";
+
+        String predictedDateString1 = ISO8601Date.getTimeZoneDateStringFromString(dateString1);
+        String actualDateString1 = "Thu, 01 Jun 2017, 23:59, UTC";
+        Assert.assertEquals(predictedDateString1, actualDateString1);
```
iamareebjamal: The order should be this - `assert(actual, predicted)`

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
iamareebjamal: This makes your tests run in parallel and hence, fast

arp95:
@iamareebjamal i have to just break it into several tests right? the checking is fine or should i add more?
i think i covered most of the edge cases.

iamareebjamal:
Looks good for now. I'll think of more use cases

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
```diff
@@ -0,0 +1,122 @@
...
+        String day2 = "2017-06-02";
+        String predictedString;
+        String actualString;
+
+        predictedString = ISO8601Date.getTimeZoneDateStringFromStringForDayFragment(day1);
+        actualString = "1 Jun";
+        Assert.assertEquals(actualString, predictedString);
+
+        predictedString = ISO8601Date.getTimeZoneDateStringFromStringForDayFragment(day2);
+        actualString = "2 Jun";
```
iamareebjamal: This is actually expected and the above is actual

Please change others accordingly. Use `expected` instead of predicted to avoid confusion.

Also, make your test names show what they are expected to do. Like `shouldConvertLocalTimeZoneStringTo...`

In this way, you can break this tests into much smaller components. See the link I gave as a reference. A signle method can have multiple tests. It is easier to see what use case failed when debugging failed errors

arp95:
@iamareebjamal is it fine now?

iamareebjamal:
Looks good @arp95 but currently contains tests only for UTC and Singapore. Test for other combination of timezones can be added as well. The reason being that the application implementation may be hardcoded for UTC and singapore making the tests pass right now. You and I know that it isn't but in black box testing, we test without knowing about the implementation. 

These test conditions could be easily added : 
1. Same event timezone as local timezone -> Date should not change
2. Same event timezone and test for different local timezones
3. Same local timezone and changing event timezone

Use different methods to test each and also, no need to use many dates in the new tests now as the edge cases with many tests were already checked in the previous tests. We don't want overlapping tests

arp95:
@iamareebjamal sorry for not seeing your post.
so what you mean is i can have more methods for testing having localtimezone as IST and event ones same and vice versa right? 
Also if i have to create the methods more how to name them? Is there a better way to name?

arp95:
@iamareebjamal please check now.

android/app/src/androidTest/java/org/fossasia/openevent/DateTest.java
```diff
@@ -0,0 +1,362 @@
...
+import org.junit.Test;
+
+/**
+ * Created by arpitdec5 on 02-06-2017.
+ */
+
+public class DateTest {
+
+    @Test
+    public void shouldConvertSingaporeDateStringToUTCDateString() {
```
iamareebjamal: Name format is correct but should not be particular like singapore or kolkata

It should be like shouldConvertLocalToInternational and then contain several tests on same string for different locales using same method.

For example, you are testing getTimeZoneDateStringFromString, so the test will only contain one date to be tested. And it will be tested on different timezones using the same method, so that if this particular test fails, we are sure that this particular feature of this method, namely changing the local date to international isn't working. All tests are good, just a little rearranging will do

One more thing. Keep the several date testing in one method for each function, like here you are using 4 strings, test them only once for a particular locale and this test will tell that the date conversion is correct. And when this test along with the above specified test (of timezone conversion) passes, we will know that both these functions of that method are working well

So, one containing 4 strings and testing on single locale - date conversion is correct
One containing single string and converting them to different locale or single string converting from different locale to local - timezone conversion is correct

This way each method will have 3 symmetrical tests which if failing will tell us that which component of the function is faulty

arp95: basically you are trying to say keep all the stuff in one method right?
one testing all 4 strings with specified timezone and eventTimeZone and the the single dateString testing with different timeZones?

iamareebjamal: No each method of Date class will have 3 tests, so 3 functions

iamareebjamal: 1 testing 4 dates for correct conversion, 2nd for local to international testing only one string but on 4 timezones, 3rd one international to local 1 string for 4 TZ

arp95: like for testing of getTimeZoneDateStringFromString(), i do following stuff:

1: use for date strings to check conversion from singapore time to utc
2: use two date strings to check conversion from singapore to amsterdam
3: use two date strings to check conversion from singapore to ist
4: use two date strings to check conversion from amsterdam to ist

you are saying to include all the above in one method right?

arp95: got it. means for each testing method 3 functions.

iamareebjamal: Yes, 3 tests methods for each date operation

iamareebjamal: And keep the test names general

arp95:
@iamareebjamal please check now.

Issue #1567, opened 2017-05-17
Use retrolambda to reduce callback complexities


iamareebjamal:
**Actual Behaviour**

We use RxJava and many callbacks and they increase the verbosity and complexity of apps

```Java
Observable.fromIterable((List<String>) objectCache.getValue("ITEMS"))
            .map(new Function<String, String>() {
                @Override
                public String apply(@NonNull String s) throws Exception {
                    return s + "Ok" + s;
                }
            }).filter(new Predicate<String>() {
            @Override
            public boolean test(@NonNull String s) throws Exception {
                return s.contains("OK");
            }
        }).flatMapCompletable(new Function<String, CompletableSource>() {
            @Override
            public CompletableSource apply(@NonNull String s) throws Exception {
                return Completable.complete();
            }
        }).subscribeOn(Schedulers.io())
            .observeOn(AndroidSchedulers.mainThread())
            .subscribe(new CallbackCompletableObserver(new Action() {
                @Override
                public void run() throws Exception {
                    // Done
                }
            }));
```

**Expected Behaviour**

Use retrolambda to use lambda functions instead

```Java
Observable.fromIterable((List<String>) objectCache.getValue("ITEMS"))
            .map(s -> s + "Ok" + s)
            .filter(s -> s.contains("OK"))
            .flatMapCompletable(s -> Completable.complete())
            .subscribeOn(Schedulers.io())
            .observeOn(AndroidSchedulers.mainThread())
            .subscribe(() -> {
                // Done
            });
```

**Would you like to work on the issue?**

Yes

**Target Milestone to solve the issue**

[Milestone 2](https://github.com/fossasia/open-event-android/milestone/25)

iamareebjamal:
@the-dagger @mananwason What do you think?

PR #1570, opened 2017-05-17
feat: Replace callbacks to lambda in DbInstance

iamareebjamal:
Fixes part of #1567 

Changes: 
- Lambdas in DbInstance


arp95:
@iamareebjamal why not in a single pr?

iamareebjamal:
It gets large and reviewing is difficult. This way independent PRs are mergeable irrespective of others

PR #1608, opened 2017-05-26
fix : App crashing on starting when generated using api endpoint

### LINKED ISSUE #1606: App generated using api endpoint is crashing on starting
**Actual Behaviour**

App generated using api endpoint is crashing on starting because of required file(event) is not available in assets folder. But we are using event file in [OpenEventApp.java](https://github.com/fossasia/open-event-android/blob/development/android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java#L137) 

**Expected Behaviour**

There should be required files in assets folder so that app don't crash. And also if in first time running app user click "No" for downloading latest data, app should show data from assets. 
We should download event data while generating app and store it in assets folder.

**Steps to reproduce it**

Generate app using api endpoin, install and run app

**LogCat for the issue**

`Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
                                                     at org.json.JSONTokener.nextCleanInternal(JSONTokener.java:116)
                                                     at org.json.JSONTokener.nextValue(JSONTokener.java:94)
                                                     at org.json.JSONObject.<init>(JSONObject.java:156)
                                                     at org.json.JSONObject.<init>(JSONObject.java:173)
                                                     at org.fossasia.openevent.OpenEventApp.onCreate(OpenEventApp.java:164)`

Yes

[Milestone 4](https://github.com/fossasia/open-event-android/milestone/27)

Shailesh351:
Fixes #1606 

Changes:

- Added methods for downloading event data and saving it in temp assets directory. Which is then being moved to assets folder of app. So that event file is available in assets folder which fixes issue. Also it saves other files like tracks, sessions, speakers etc in assets which will be used to load data when first time running app.

- Removed redundant files(background.png & ic_launcher.png) from assets


iamareebjamal:
Please also remove the already included files from assets

Shailesh351:
Done @iamareebjamal  :+1:

iamareebjamal:
@niranjan94 Please review

PR #1621, opened 2017-05-29
fixes nullpointer exception in SpeakerListResponseProcessor.java

### LINKED ISSUE #1620: Nullpointer exception in SpeakerResponseProcessor.java
**Actual Behaviour**
If the speaker's "session" array list is empty, the code throws a null pointer exception and the app crashes on starting. This is because gson is interpreting the empty array list as null and so `speakers.getsession()` is null.

**Expected Behaviour**

There should be a check for empty array list.

**LogCat for the issue**

` Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.util.ArrayList.size()' on a null object reference  at org.fossasia.openevent.api.processor.SpeakerListResponseProcessor.lambda$onResponse$0(SpeakerListResponseProcessor.java:43)`

Yes.

https://github.com/fossasia/open-event-android/milestone/27

mahikaw:
Fixes #1620 

Changes: Handles the exception by using an` if` statement to execute the speaker's session `for` loop only when the session's array list is not empty.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
+++ b/android/app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
@@ -31,10 +31,12 @@ public class SpeakerListResponseProcessor implements Callback<List<Speaker>> {
                 ArrayList<String> queries = new ArrayList<>();
 
                 for (Speaker speaker : response.body()) {
-                    for (int i = 0; i < speaker.getSession().size(); i++) {
-                        SessionSpeakersMapping sessionSpeakersMapping = new SessionSpeakersMapping(speaker.getSession().get(i).getId(), speaker.getId());
-                        String query_ss = sessionSpeakersMapping.generateSql();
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/org/fossasia/openevent/api/processor/SpeakerListResponseProcessor.java
```diff
@@ -31,11 +31,13 @@ public void onResponse(Call<List<Speaker>> call, final Response<List<Speaker>> r
                 ArrayList<String> queries = new ArrayList<>();
 
                 for (Speaker speaker : response.body()) {
+                    if(speaker.getSession()!=null){
```
iamareebjamal: Can you please fix the indentation?

Also, appropriate spacing will be great
`if (speaker.getSession() != null) {`

Try to write code like it will be when you reformat using IDE :smile: 

Otherwise looks good

mahikaw:
Please check this now @iamareebjamal

PR #1630, opened 2017-06-01
feat: Show progress bar dialog while downloading event data

### LINKED ISSUE #1490: Show progress of completion while downloading
**Actual Behaviour**

Currently, when the app id downloading data and if we move to the tracks section, it displays "no tracks", as shown below.

**Expected Behaviour**

The screen should indicate that data is downloading.
**Steps to reproduce it**
Go to tracks section soon after starting the app.

**Screenshots of the issue**

screenshot_20170506-143043

yes
@mariobehling @the-dagger @mananwason can I work on this issue and send a PR?

iamareebjamal:
Fixes #1490 

- Create DownloadCompleteHandler for handling download events
- Create fluent API for DownloadCompleteHandler
- Remove download related events from MainActivity
- Handle events ro show progress of download
- Propagate success or error to MainActivity using CompletableSubject
- Hide/Show progressbar at appropriate events

Screenshots for the change: 
device-2017-06-01-100737


iamareebjamal:
@arp95 @rohanagarwal94 @Shailesh351 @mahikaw Please review

arp95:
@iamareebjamal having a look.

shreyasnbhat:
Looks good!

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -210,9 +202,10 @@ protected void onCreate(Bundle savedInstanceState) {
 
         setUpToolbar();
         setUpNavDrawer();
-        setUpProgressBar();
         setUpCustomTab();
 
+        completeHandler = DownloadCompleteHandler.with(context);
```
rohanagarwal94: Unable to understand why do this instead of just creating a new object of the class. Could learn something new.

iamareebjamal: This is fluent API

When not attaching detaching much, could be used as

````Java
DownloadCompleteHandler.with(context)
    .startListening()
    .show()
    .withCompleteListener()
    .subscribe();
````

The full effect of its ease is not visible as there are lots of `show` and `hide` called on it, needing to separate the object

iamareebjamal: See Picasso and other modern libraries. When not using this, it may confuse the user. How to create the object, which method to call first?

This provides a natural order with chaining configuration

rohanagarwal94: Great!

iamareebjamal:
@arp95 Looking into point 1. I'll get back to you

About point 2, if there's an update, we will never know. The server down't support push notifications and the user can only pull to refresh new information, but it is section wise and not as whole

iamareebjamal:
@arp95 Checked it. If there is no internet connection while install and user returns to app, the dialog is shown again

iamareebjamal:
@the-dagger Please take a look

iamareebjamal:
@the-dagger Please take a look

arp95:
@iamareebjamal the problem i listed above still occurs. if user doesnt have internet and then switches on the WIFI he/she is not shown the download option. Please check.

iamareebjamal:
I got your point :+1: But I think it should be implemented as a separate feature. This PR should only handle what's written in the issue. To show dialog while downloading

arp95:
@iamareebjamal can i work on that then?

iamareebjamal:
@arp95 Sure

PR #1653, opened 2017-06-04
Fixed minor UI bugs in location activity

### LINKED ISSUE #1651: Location activity UI bug
**Actual Behaviour**
Incase of no sessions in a location,

- the map of that particular location also shows up the default "no sessions" text view with the map of the location.

- On clicking back button, first the map goes, leaving just the "no sessions" text view. On clicking back button for the second time, the locations list appears.

**Expected Behaviour**

- Even if there aren't any sessions, it should simply display the map of the locations and not the "no sessions" text.

- The back button should directly lead to sessions list.

**Screenshots of the issue**

device-2017-06-04-110236

Yes.

https://github.com/fossasia/open-event-android/milestone/27

mahikaw:
Fixes #1651

Changes: 
When there aren't any sessions, only the map of the locations is displayed and not the "no sessions" text.

The back button directly leads to sessions list with the text "no session", as expected.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
@@ -145,6 +145,14 @@ public class LocationActivity extends BaseActivity implements SearchView.OnQuery
         searchView.setQuery(searchText, false);
         return true;
     }
+    @Override
+    public void onBackPressed(){
+        if((mSessions.isEmpty())){
+            noSessionsView.setVisibility(View.VISIBLE);
+        }
+        super.onBackPressed();
+
+    }
 
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()){
@@ -153,6 +161,7 @@ public class LocationActivity extends BaseActivity implements SearchView.OnQuery
                 FragmentTransaction fragmentTransaction = fragmentManager.beginTransaction();
                 fragmentTransaction.replace(R.id.content_frame_location, ((OpenEventApp) getApplication()).getMapModuleFactory().provideMapModule().provideMapFragment(), FRAGMENT_TAG_LOCATION).addToBackStack(null).commit();
                 sessionRecyclerView.setVisibility(View.GONE);
+                noSessionsView.setVisibility(View.GONE);
                 return true;
             case android.R.id.home:
                 onBackPressed();


android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
```diff
@@ -145,6 +145,14 @@ public boolean onCreateOptionsMenu(Menu menu) {
         searchView.setQuery(searchText, false);
         return true;
     }
+    @Override
+    public void onBackPressed(){
+        if((mSessions.isEmpty())&&(sessionRecyclerView.getVisibility()==View.GONE)){
```
iamareebjamal: Is checking the sessionRecyclerView's visibility necessary here. If we include this check here, then we are depending on other variable whose state may give us different results making the code buggy

mahikaw: In this case, if the recycler view is visible, then there is no point for the back button. The back button is to reach the session list activity for that microlocation. Similar coding has been done previously too.

iamareebjamal: The problem is that the visibility of recyclerview is in our hand. If someone forgets to hide or show it, this conditional will give different results

mahikaw: could you suggest a workaround for it in that case. I can't really think of anything else. Also, we might want to change the existing code and remove such conditions in that case?

iamareebjamal: I'll get back to you after testing this

iamareebjamal: @mahikaw I'm testing this currently, can you tell me the specific use case for a particular location where this condition will be needed?

mahikaw: Yes, when on the map fragment, if we press backbitten and there are no sessions, the recycler view doesn't appear, but what the current code doesn't do is to make the no sessions text view visible. I have tried to address that here.

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
```diff
@@ -145,6 +145,14 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+            noSessionsView.setVisibility(View.VISIBLE);
```
mejariamol: What happens when user presses back button second time once `noSessionsView` is set visible?

mejariamol: What happens when user presses back button second time once `noSessionsView` is visible?

mahikaw: then it skips this onbackpressed() and instead the super.onbackpressed() is called which leads to locations list as expected.

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
mahikaw: @iamareebjamal is this PR sorted now?

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
mejariamol: For that to happen there should be `noSessionsView` visibility check in the case right?

mahikaw: I don't think so. I tested it and it was working just fine. Could you confirm @mejariamol

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
mahikaw: @iamareebjamal any updates regarding this?

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
arp95: @mejariamol it works fine in that case.

iamareebjamal: @mahikaw Sorry for late reply. I tried just using `mSessions.isEmpty()`, and it worked for most use cases. Can you confirm a use case where it won't, because if it is so, this too is a bug

Because `sessionRecyclerView` should always be `View.GONE` when `mSessions.isEmpty()`

mahikaw:
@iamareebjamal Did you have a chance to check this?

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
mahikaw: Yes, you are right @iamareebjamal I am updating it.

mahikaw:
@iamareebjamal Please review and merge.

android/app/src/main/java/org/fossasia/openevent/activities/LocationActivity.java
```diff
@@ -145,6 +145,14 @@ public boolean onCreateOptionsMenu(Menu menu) {
         searchView.setQuery(searchText, false);
         return true;
     }
+    @Override
+    public void onBackPressed(){
+        if((mSessions.isEmpty())){
+            noSessionsView.setVisibility(View.VISIBLE);
+        }
```
Shailesh351: @mahikaw here if mSessions is not empty then recycler view should be visible.

mahikaw: Yes, @Shailesh351 It is visible. This is being handled by the super.onbackpressed()

Issue #1637, opened 2017-06-01
Search facility in map 


heysadboy:
**Actual Behaviour**

Currently the map looks like this

![screenshot_2017-06-02-00-34-14](https://cloud.githubusercontent.com/assets/22588035/26696351/d1670f48-472b-11e7-892c-5d8dc679eb97.png)

**Expected Behaviour**

- There should be a search button and this SearchView could be in toolbar. 
- Text overlaps with map components.
- It should go in single line not in paragraph.

**Would you like to work on the issue?**

Yes.

**Target Milestone to solve the issue**

Milestone 5

heysadboy:
@iamareebjamal I don't see changes made in this https://github.com/fossasia/open-event-android/pull/1625 . Is this reverted back?

iamareebjamal:
Please confirm you have Google Play variant selected

PR #1656, opened 2017-06-05
Improving the search facility in map for play version

heysadboy:
Fixes https://github.com/fossasia/open-event-android/issues/1637

- Improving the search UX by providing the user facility to select the location instead of typing and clearing up the views earlier present on the map fragment.
- Putting up the search button it toolbar.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
+++ b/android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
@@ -176,7 +176,9 @@ public class OpenEventApp extends Application {
             JSONObject jsonObject = new JSONObject(event_json);
             String org_description = jsonObject.has(ConstantStrings.ORG_DESCRIPTION) ?
                     jsonObject.getString(ConstantStrings.ORG_DESCRIPTION) : "";
+            String eventTimeZone = jsonObject.has(ConstantStrings.TIMEZONE) ? jsonObject.getString(ConstantStrings.TIMEZONE) : "";
             sharedPreferences.edit().putString(ConstantStrings.ORG_DESCRIPTION, org_description).apply();
+            sharedPreferences.edit().putString(ConstantStrings.TIMEZONE, eventTimeZone).apply();
         } catch (JSONException e) {
             e.printStackTrace();
         }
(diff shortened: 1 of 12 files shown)


iamareebjamal:
@heysadboy Please provide screenshots

heysadboy:
@iamareebjamal These changes are for Google Play variant only, for which I have attached screenshots in the issue. Clicking on search button on toolbar goes like this,

<table>
<td>

</td>
<td>

</td>
</table>

iamareebjamal:
@heysadboy Yeah sorry. Forgot to see the Play variant changes

Also mention what you have changed in OSMapFragment as well please

heysadboy:
@iamareebjamal It was a stray button in toolbar to share map, when I removed that from the menu_map.xml to add search option in play version I had to remove that from OSMapFragment to stop the droid version from crashing. I haven't modified any functioning for droid versionother than that.

iamareebjamal:
OK. But now, please create an issue for adding a share button for location sharing and take a look at the suggestion for autocompletetextview

heysadboy:
@iamareebjamal Creating issue for that here https://github.com/fossasia/open-event-android/issues/1659. I considered that initially but I didn't use it because a user may not know about all the locations' textual info. like spelling etc. Selecting from a list was a better option as unlike the name of locations displayed in other fragments, no info. is displayed here.

iamareebjamal:
@the-dagger @mariobehling @mananwason Please suggest

mahikaw:
I agree  that the user might not know all the locations out there. But on the other hand, if there are a large number of locations, this type of list will make it tough for the user to navigate through @heysadboy 

I would suggest to incorporate both. The list will appear when the user has typed nothing and then the search will come into play as soon as user starts typing.

heysadboy:
@mahikaw That's a great suggestion. I think that would be the best of both options.

heysadboy:
@iamareebjamal Making the required changes, I am using AutoCompleteTextView now.
<table>
<td>

</td>
<td>

</td>
</table>

iamareebjamal:
@heysadboy Please remove Mahika's commit from your PR

heysadboy:
@iamareebjamal I tried removing but ended up removing all the commits accidentaly :( , I am really very sorry about this. Please find new one here https://github.com/fossasia/open-event-android/pull/1668

PR #1684, opened 2017-06-09
feat : Cache sponsor images and speaker's thumbnails

### LINKED ISSUE #1326: Displaying cached images using Picasso
**Actual Behaviour**

Earlier in my issue i used Picasso to display cached images in SponserFragment. Extending that to all the parts having images.

**Expected Behaviour**

Display cached images when offline.

yes.

Shailesh351:
Fixes #1326 

Changes:
Display thumbnails and sponsor images when app is offline by providing more cache.
Only use this picasso instance for displaying thumbnails and sponsors.


android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java
```diff
@@ -106,6 +109,12 @@ public void onCreate() {
         }
         refWatcher = LeakCanary.install(this);
 
+        //Initialize Cache
+        File httpCacheDirectory = new File(getCacheDir(), "picasso-cache");
+        Cache cache = new Cache(httpCacheDirectory, 15 * 1024 * 1024);
+
```
arp95: this is the amount of memory the OS will allocate to the process right after the app opens?
are we sure OS will always provide this amount of cache?
Also if no, then dont you think rendering images will become a problem?

iamareebjamal: This is disk space @arp95c`cacheDir` -> cache directory

If it would have been allocating memory, it would be in memory cache making all saved objects clear after app reload, so wouldn't make sense

iamareebjamal: @Shailesh351 I think 10 MB would be better, as some individual photos themselves surpass 1 MB

iamareebjamal: And also, this is FIFO or LRU cache most probably, so even if it is full, images will show up

arp95: thanks for pointing that out. 
but i am still doubtful that every device will allow a cache file of 10MB. Lets say if its full then there is no point caching right?

iamareebjamal: This will just feed the max cache size to the LruCache class, and it will handle space management by itself. It won't create the 10 MB file on opening of the app

iamareebjamal:
@Shailesh351 There are other images like event background and logo, which I think should be cached too

Shailesh351:
@iamareebjamal  @arp95 @shreyasnbhat @rohanagarwal94 Please review

iamareebjamal:
Please resolve conflicts @Shailesh351

Shailesh351:
@iamareebjamal resolved conflicts

PR #1702, opened 2017-06-12
fix: Add icon support for JPEG (deprecated)

### LINKED ISSUE #1701: Convert JPEG icons to PNG
**Actual Behaviour**

App generator crashes due to `None` quality for JPEG icons

**Expected Behaviour**

There should be an option for generator to convert JPEG to PNG in generator

Yes

iamareebjamal:
Fixes #1701

Changes: 
- Adds support for JPEG launcher icons by converting them to PNG


iamareebjamal:
@niranjan94 Please review

iamareebjamal:
@niranjan94 said that we should not support JPEG. So, changing it along the lines to notify the user that your build failed due to the icon being JPEG

But later as currently, there are more pressing issues

arp95:
i think this is important. we cant restrict the event organiser to provide png logo. this should be taken care of.

niranjan94:
@arp95 then we should be doing a conversion to png in the generator if it is not png.

iamareebjamal:
Yeah, changing to this

iamareebjamal:
@niranjan94 Please review

PR #1731, opened 2017-06-22
feat : Make models jackson compatible for JSONAPI implementation

### LINKED ISSUE #1729: Make models jackson compatible for JSONAPI implementation
**Actual Behaviour**

Currently all models are being serialized/deserialized using gson. For jsonapi we are using `https://github.com/jasminb/jsonapi-converter` which requires https://github.com/FasterXML/jackson-databind.

**Expected Behaviour**

All model should be jackson compatible for jsonapi implementation

Yes

[Milestone 7](https://github.com/fossasia/open-event-android/milestone/30)

Shailesh351:
Fixes #1729  

Changes: 
- Transfer from Gson to Jackson

Here Dummy Observable class is needed.  Ref : https://realm.io/docs/java/latest/#jackson-databind


Shailesh351:
@arp95 @iamareebjamal @shreyasnbhat Please review

android/app/build.gradle
```diff
@@ -36,6 +36,9 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         buildConfigField "String", "GIT_SHA", "\"${gitSha}\""
         buildConfigField "String", "BUILD_TIME", "\"${buildTime}\""
+
+        // Enabling multidex support.
+        multiDexEnabled true
```
iamareebjamal, line 41: Is the build surpassing 65k methods?

android/app/src/main/java/rx/Observable.java
```diff
@@ -0,0 +1,6 @@
+package rx;
+
+public class Observable {
+    // Dummy class required for Jackson-Databind support if
+    // RxJava is not a project dependency.
```
iamareebjamal, line 5: But RxJava is project dependency

iamareebjamal (changes requested):
Some doubts. 

Did you add any field except getters and setters?

android/app/build.gradle
Shailesh351: Yes, it is because of jackson jar file is large and have more methods

Shailesh351:
@iamareebjamal I have not added new fields. I have planned to add new fields and change Json property values when we switch to JSONAPI

android/app/src/main/java/rx/Observable.java
Shailesh351: Yes it is but  For realm we have not used Observable. 
https://realm.io/docs/java/latest/#rxjava

iamareebjamal (approved):

rohanagarwal94 (approved):

arp95 (approved):

shreyasnbhat:
@Shailesh351 Sorry for the delay. Looking at this now.

PR #1742, opened 2017-06-27
fix: replaced field names with setter annotations

### LINKED ISSUE #1740: Adding appropriate annotations for modified field names of the models
**Actual Behaviour**

Currently some field names of models have been changed accoridng to the new api.

**Expected Behaviour**

Adding annotations to support the modified field names of models.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

https://github.com/fossasia/open-event-android/milestone/31

arp95:
Fixes #1740 

Changes: 
replaced field names with the setter and getter annotations. will remove the old model ones, once the new api is up.

Screenshots for the change: 
-

@iamareebjamal @Shailesh351 @shreyasnbhat @rohanagarwal94 
please review.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/data/Event.java
+++ b/android/app/src/main/java/org/fossasia/openevent/data/Event.java
@@ -1,6 +1,7 @@
 package org.fossasia.openevent.data;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSetter;
 
 import org.fossasia.openevent.data.extras.CallForPapers;
 import org.fossasia.openevent.data.extras.Copyright;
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/org/fossasia/openevent/data/Event.java
```diff
@@ -16,77 +16,84 @@
...
     private Double latitude;
-    @JsonProperty("organizer_description")
     private String organizerDescription;
     private String identifier;
     private String type;
     private String email;
-    @JsonProperty("code_of_conduct")
     private String codeOfConduct;
 
+    @JsonGetter("event_url")
```
iamareebjamal: No need to annotate getters, prefer JsonProperty there. Getter annotation creates unnecessary boilerplate. Only annotate setters when there are more than one attribute name for same property. JsonProperty should be preferred over JsonSet/Getter. But we have to delegate to setter when there are more than one attribute name, but there is no need of getter annotation as we don't serialize the POJOs, only deserialize

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/data/Event.java
arp95: but if the atttribute has both a setter and a getter than should i make the json property the old one which will be used for the getter and update the setter with multiple field names?

arp95:
@iamareebjamal 
made changes.

android/app/src/main/java/org/fossasia/openevent/data/Event.java
iamareebjamal: If there is a conflict in property name, then remove JsonProperty and create 2 setters with names from both properties

iamareebjamal (approved):

iamareebjamal:
There are more changes to be done I guess @arp95 ?

arp95:
@iamareebjamal 
i went through twice. but didnt find any setter method having conflicting attribute names. are there according to you?

iamareebjamal:
I mean there are much more additions to do make it compatible with new API

arp95:
you mean the new attributes? plus i havent changed the attributes with getters for now. will cahnge them when the new api is set up.

iamareebjamal:
Please do those changes as well as it won't concern the old API and only make things easier when the new API is deployed

Your doc has listed some changes pertaining to attribute name in sessions, sponsors, etc. Do add setters for ambiguous attributes

https://docs.google.com/document/d/1caaCeOOTIxQn2HR1Uj9r-cHXQH5hSYzzzCJpt3rkl38/edit

arp95:
so i have to these two things right?

add new attributes with jsonproperty and 
add setters for all conflicted attributes.

right?

iamareebjamal:
Just the second part in this PR. First part will require migrations too

iamareebjamal (changes requested):
Add changes for conflicting attribute names

arp95:
made changes. please see now.

iamareebjamal:
@arp95 The build is failing, please take a look

arp95:
@iamareebjamal 
please review now.

iamareebjamal (dismissed):

arp95:
@Shailesh351 @shreyasnbhat @rohanagarwal94 
please review.

android/app/src/main/java/org/fossasia/openevent/data/Sponsor.java
```diff
@@ -37,6 +36,26 @@ public String getUrl() {
         return url;
     }
 
+    @JsonSetter("session_type")
```
Shailesh351: It should be `sponsor_type`

android/app/src/main/java/org/fossasia/openevent/data/Sponsor.java
```diff
@@ -37,6 +36,26 @@ public String getUrl() {
...
+    public void setSponserType(String sponserType) {
+        this.sponsorType = sponserType;
+    }
+
+    @JsonSetter("session-type")
```
Shailesh351: here also

Shailesh351 (changes requested):

arp95:
@iamareebjamal @Shailesh351 @shreyasnbhat @rohanagarwal94 
please review now.

Shailesh351 (approved):

arp95:
@mariobehling @the-dagger 
please review now.

iamareebjamal (approved):

PR #1750, opened 2017-06-29
Implement caching in the livefeed

### LINKED ISSUE #1735: Implement caching in the livefeed
**Actual Behaviour**

The livefeed currently works only online.

**Expected Behaviour**

The previously fetched feed should be displayed in case the user is offline.

**Steps to reproduce it**

1. Open livefeed item from the nav drawer without the internet

**Screenshots of the issue**

screenshot_2017-06-25-20-27-27-793_org fossasia openevent

Yes.

rohanagarwal94:
Fixes #1735

This will implement caching in the live feed and solve app crash in the global search.


rohanagarwal94:
@iamareebjamal @arp95 @Shailesh351 @shreyasnbhat please review.

android/app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -30,47 +39,109 @@
...
+            OkHttpClient okHttpClient = okHttpClientBuilder.addInterceptor(new HttpLoggingInterceptor()
+                    .setLevel(HttpLoggingInterceptor.Level.BASIC))
+                    .addInterceptor(provideOfflineCacheInterceptor())
+                    .addNetworkInterceptor(provideCacheInterceptor())
+                    .cache(provideCache())
+                    .build();
+
+            Retrofit.Builder retrofitBuilder = new Retrofit.Builder()
+                    .addConverterFactory(JacksonConverterFactory.create(new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)))
+                    .client(okHttpClient);
```
iamareebjamal: Please pre-build the common functionalities statically. This code block is duplicated

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/api/APIClient.java
rohanagarwal94: the okHttpClient paramter is not same. I have provided the cache and added the interceptor for the the facebook okhttpclient.

iamareebjamal: I know. Still, the retrofit bit except client is same. Similar to what it was before.

rohanagarwal94: okay, doing it.

rohanagarwal94:
@iamareebjamal updated.

android/app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -30,47 +39,110 @@
...
+                    .header(CACHE_CONTROL, cacheControl.toString())
+                    .build();
+        };
+    }
+
+    private static Interceptor provideOfflineCacheInterceptor() {
+        return chain -> {
+            Request request = chain.request();
+
+            if (!NetworkUtils.haveNetworkConnection(OpenEventApp.getAppContext())) {
```
iamareebjamal, line 131: This will block the main thread for seconds if there's a slow connection

android/app/src/main/java/org/fossasia/openevent/api/APIClient.java
```diff
@@ -30,47 +39,110 @@
...
         return facebookGraphAPI;
     }
 
+    private static Cache provideCache() {
+        Cache cache = null;
+        try {
+            cache = new Cache(new File( OpenEventApp.getAppContext().getCacheDir(), "facebook-feed-cache"),
+                    10 * 1024 * 1024 ); // 10 MB
+        }
+        catch (Exception e) {
```
iamareebjamal: Move cache to previous line

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/api/APIClient.java
rohanagarwal94: So any suggestions for this?

iamareebjamal: Why does this need to be dynamic?

rohanagarwal94: @iamareebjamal If the user is offline, he will not get the cache response automatically. The request header should only be modified if the user is offline. The stale results will only be offered if the new feed cannot be fetched. Besides this method just checks whether the WIFI/mobile network is turned on. The method you are thinking about is the checkNetworkConnection in the NetworkUtils IMO.

iamareebjamal: OK

iamareebjamal: Just add the cache syntax fix

rohanagarwal94:

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -99,7 +98,6 @@ public void handleVisibility() {
     @Override
     public void onDestroyView() {
         super.onDestroyView();
-        OpenEventApp.getEventBus().unregister(this);
```
iamareebjamal, line 111: Why is this removed? This will cause a memory leak

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
rohanagarwal94: There is no event bus registered for the fetching of feed event the reason being that rxJava was used to subscribe to feed observable.

iamareebjamal: OK

rohanagarwal94:
@arp95 @iamareebjamal @Shailesh351 @shreyasnbhat please review.

iamareebjamal (approved):

arp95 (approved):
LGTM.

Shailesh351 (approved):
Looks good. 

One suggestion it is not caching images so when offline it is showing big camera icons. Will it be better to  show image if it is loaded otherwise not visible ?

rohanagarwal94:
@shailesh351 I initially stored images in the picasso cache which you implemented. But @iamareebjamal pointed out that since the feed is bound to have new content frequently, the cache would fill up quickly.

iamareebjamal:
@rohanagarwal94 You can set appropriate placeholder image so that it does not look weird while loading, however that is another issue

rohanagarwal94:
@iamareebjamal the placeholder is already set to a camera icon.

iamareebjamal:
The problem as @Shailesh351 said was that it was not appropriate in this setting. As it appeared enlarged. See if you can an appropriate placeholder

rohanagarwal94:
@iamareebjamal let me create another issue for it. Discussing that here is futile IMO.
@mariobehling @the-dagger please review and merge.

PR #1751, opened 2017-06-29
Add progress bar to the livefeed

### LINKED ISSUE #1736: Add progress bar to the livefeed
**Actual Behaviour**

Currently empty screen is shown before the feed loads into the view.

**Expected Behaviour**

There should be a progress bar while the feed loads. @iamareebjamal @arp95 @Shailesh351 @shreyasnbhat @mahikaw should the circular progress bar be used or the normal progress bar shown at the start of the app?

**Screenshots of the issue**

screenshot_2017-06-25-20-27-27-793_org fossasia openevent

Yes.

rohanagarwal94:
Fixes #1736

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.fragments;
 
+import android.app.ProgressDialog;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
@@ -38,6 +39,7 @@ public class FeedFragment extends BaseFragment {
     private FeedAdapter feedAdapter;
     private List<FeedItem> feedItems;
     private SharedPreferences sharedPreferences;
+    private ProgressDialog downloadProgressDialog;
 
     @BindView(R.id.feed_swipe_refresh) SwipeRefreshLayout swipeRefreshLayout;
     @BindView(R.id.feed_recycler_view) RecyclerView feedRecyclerView;
@@ -55,6 +57,11 @@ public class FeedFragment extends BaseFragment {
         feedAdapter = new FeedAdapter(getContext(), (FeedAdapter.AdapterCallback)getActivity(), feedItems);
         feedRecyclerView.setAdapter(feedAdapter);
         sharedPreferences = PreferenceManager.getDefaultSharedPreferences(getContext());
+
+        setupProgressBar();
+
+        if(NetworkUtils.haveNetworkConnection(getContext()))
+            showProgressBar(true);
 
         downloadFeed();
 
@@ -80,9 +87,11 @@ public class FeedFragment extends BaseFragment {
                             .getString(R.string.refresh_failed), Snackbar.LENGTH_LONG)
                             .setAction(R.string.retry_download, view -> refresh()).show();
                     Timber.d("Refresh not done");
+                    showProgressBar(false);
                 }, () -> {
                     swipeRefreshLayout.setRefreshing(false);
                     Timber.d("Refresh done");
+                    showProgressBar(false);
                 });
     }
 
@@ -155,6 +164,24 @@ public class FeedFragment extends BaseFragment {
             }
         });
 
+    }
+
+    private void showProgressBar(boolean show) {
+        if(show)
+            downloadProgressDialog.show();
+        else
+            downloadProgressDialog.dismiss();
+    }
+
+    private void setupProgressBar() {
+        downloadProgressDialog = new ProgressDialog(getContext());
+        downloadProgressDialog.setIndeterminate(true);
+        downloadProgressDialog.setProgressPercentFormat(null);
+        downloadProgressDialog.setProgressNumberFormat(null);
+        downloadProgressDialog.setProgressStyle(ProgressDialog.STYLE_HORIZONTAL);
+        downloadProgressDialog.setCancelable(false);
+        String shownMessage = String.format(getString(R.string.downloading_format), getString(R.string.menu_feed));
+        downloadProgressDialog.setMessage(shownMessage);
     }
 
     @Override


rohanagarwal94:
@iamareebjamal @arp95 @shreyasnbhat @Shailesh351 please review.

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -49,6 +51,11 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
 
         View view = super.onCreateView(inflater, container, savedInstanceState);
 
+        downloadCompleteHandler = new DownloadCompleteHandler(getContext(), R.string.menu_feed);
```
iamareebjamal: The DownloadCompleteHandler is made to show dialog of completion for the Event API, not sure why being used in Feed?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
rohanagarwal94: Used its progress bar to avoid code redundancy. Want me to setup another progress bar here?

iamareebjamal: Yes, because both classes are completely different. One listens to Event Info download, and your use case is just showing an intermediate progress dialog I guess?

rohanagarwal94: yes thats it.

rohanagarwal94:

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -56,6 +58,11 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         feedRecyclerView.setAdapter(feedAdapter);
         sharedPreferences = PreferenceManager.getDefaultSharedPreferences(getContext());
 
+        setupProgressBar();
+
+        if(NetworkUtils.haveNetworkConnection(getContext()))
+        showProgressBar(true);
```
iamareebjamal: Indent this correctly please

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -80,9 +87,12 @@ private void downloadFeed() {
                             .getString(R.string.refresh_failed), Snackbar.LENGTH_LONG)
                             .setAction(R.string.retry_download, view -> refresh()).show();
                     Timber.d("Refresh not done");
+                    showProgressBar(false);
                 }, () -> {
+                    downloadProgressDialog.setProgress(100);
```
iamareebjamal: Are you incrementing the progress or it just turns from 0 to 100. If latter, then intermediate progress bar is preferred

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
rohanagarwal94: You mean indeterminate right?

iamareebjamal: Yeah, sorry :stuck_out_tongue:

rohanagarwal94:
@iamareebjamal updated.

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -157,6 +166,24 @@ public void networkUnavailable() {
...
+            downloadProgressDialog.show();
+        else
+            downloadProgressDialog.dismiss();
+    }
+
+    private void setupProgressBar() {
+        downloadProgressDialog = new ProgressDialog(getContext());
+        downloadProgressDialog.setIndeterminate(true);
+        downloadProgressDialog.setProgressPercentFormat(null);
+        downloadProgressDialog.setProgressNumberFormat(null);
```
iamareebjamal, line 180: Are these needed?
```
downloadProgressDialog.setProgressPercentFormat(null);
downloadProgressDialog.setProgressNumberFormat(null);
```

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -157,6 +166,24 @@ public void networkUnavailable() {
...
+        downloadProgressDialog.setProgressStyle(ProgressDialog.STYLE_HORIZONTAL);
+        downloadProgressDialog.setCancelable(false);
+        String shownMessage = String.format(getString(R.string.downloading_format), getString(R.string.menu_feed));
+        downloadProgressDialog.setMessage(shownMessage);
```
arp95, line 184: isnt it better to use a progress bar from a separate source or something. becasue we are using it in home screen as well.

iamareebjamal: From a separate source?

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
rohanagarwal94: Oh yes. By default the horizontal progress bar shows the percentage and the number in the progress. Hence it always stays 0/100 since its a binary task. Hence had to explicitly remove it.

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
arp95: i meant using a separate java file for these lines as we might want to use a progress bar for other activites as well. just a thought.

rohanagarwal94: That has been discussed in the threads above. As @iamareebjamal pointed out, the downloadCompleteHandler class was meant to be handling event download events and therefore it seemed it was not a good practice to use just the progress bar from that class when the work performed by both is poles apart.

arp95: sorry for not following the thread. @iamareebjamal is it possible to make a generic class which can handle both scenarios or should we stick to this for now?

iamareebjamal: Both scenarios are different as one is listening to progress and incrementing on changes and other one is intermediate. This one is not getting reused anywhere in the app, or else we would have created a single class. There is no code sharing between these use cases

iamareebjamal (approved):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
arp95: got it. in future if there is a scenario for example adding twitter news then we should go for a separate class then.

arp95 (approved):
LGTM.

Shailesh351 (approved):

Issue #1757, opened 2017-07-01
The download progress bar gets stuck on app startup


rohanagarwal94:
**Actual Behaviour**

The download progress bar gets stuck at 85%.

**Expected Behaviour**

The download should be complete.

**Screenshots of the issue**

![screenshot_2017-07-01-14-31-33-301_org fossasia openevent](https://user-images.githubusercontent.com/13241256/27761406-e7260b86-5e78-11e7-9a92-c194ce565f7d.png)

**Would you like to work on the issue?**

Yes

**Target Milestone to solve the issue**

[milestone 8](https://github.com/fossasia/open-event-android/milestones/31)

rohanagarwal94:
@the-dagger please close this too.

PR #1759, opened 2017-07-01
fix : Realm access from incorrect thread.

### LINKED ISSUE #1758: App crash in the livefeed
**Actual Behaviour**

The app is crashing in the live feed after the latest series of merges.

**Expected Behaviour**

The live feed should work fine.

**Steps to reproduce the issue**

1. Open the app.
2. The progress bar seems to be be stuck at 85%.
3. Minimise the app.
4. Open the live feed tab from the Nav Drawer.

**LogCat for the issue**

screenshot from 2017-07-01 16-12-38

Yes

rohanagarwal94:
Fixes #1758 and #1757

Issue logs discovered after debugging only - 

screenshot from 2017-07-01 16-48-56

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
+++ b/android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
@@ -472,6 +472,7 @@ public class RealmDataRepository {
     //Session Types Section
 
     private void saveSessionTypesInRealm(List<SessionType> sessionTypes) {
+        Realm realm = Realm.getDefaultInstance();
         realm.executeTransaction(realm1 -> realm1.insertOrUpdate(sessionTypes));
         realm.close();
     }


rohanagarwal94:
@iamareebjamal @arp95 @Shailesh351 @shreyasnbhat please review.

Shailesh351 (approved):
@rohanagarwal94 I think issue regarding app carsh in live feed is different.

rohanagarwal94:
@shailesh351 even I thought the same. But to my surprise as and when the downloader issue got resolved, the live feed too didnt crash.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
```diff
@@ -472,6 +472,7 @@ public Completable saveLocations(final List<Microlocation> locations) {
     //Session Types Section
 
     private void saveSessionTypesInRealm(List<SessionType> sessionTypes) {
+        Realm realm = Realm.getDefaultInstance();
         realm.executeTransaction(realm1 -> realm1.insertOrUpdate(sessionTypes));
```
iamareebjamal, line 476: The solution is to not create a new instance. See other method bodies on how it is done

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
rohanagarwal94: @iamareebjamal I have seen other method bodies before coming to the solution. Please check it from your side.

iamareebjamal: OK, thread safety in Realm has really taken every bit of it over.

iamareebjamal (approved):

PR #1761, opened 2017-07-01
fix: added support for new attributes in Session model

### LINKED ISSUE #1756: adding support for new attributes in Session model
**Actual Behaviour**

Currently the old model doesnt support created-at, deleted-at, submitted-at and is-mail-sent attributes of new model.

**Expected Behaviour**

Adding support for them.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

@iamareebjamal please add it.
https://github.com/fossasia/open-event-android/milestone/31

arp95:
Fixes #1756 

Changes: 

Added support for new attributes in Session model.

Screenshots for the change: 
-

@iamareebjamal @Shailesh351 @shreyasnbhat @rohanagarwal94 
please review.


android/app/src/main/java/org/fossasia/openevent/data/Session.java
```diff
@@ -40,6 +40,30 @@
     @Index
     private String startDate;
     private boolean isBookmarked;
+    private String createdAt;
+    private String deletedAt;
+    private String submittedAt;
+    private boolean isMailSent;
+
+    @JsonSetter("created-at")
```
iamareebjamal: Prefer `JsonProperty` if non-conflicting attributes

iamareebjamal (changes requested):

arp95:

arp95:
Guys please review.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
```diff
@@ -17,31 +18,18 @@
     @Override
     public void migrate(DynamicRealm dynamicRealm, long oldVersion, long newVersion) {
 
-        // Add schema increments and change methods when updating models
-
-        /* Example migration :
-
         // DynamicRealm exposes an editable schema
-        RealmSchema schema = realm.getSchema();
+        RealmSchema schema = dynamicRealm.getSchema();
```
iamareebjamal: Realm schema version was not changed at the time of declaration. See the link I posted

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
arp95: one confusion, is the constant variable DB_VERSION should be incremented?

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
```diff
@@ -1,48 +1,31 @@
...
-    public void migrate(DynamicRealm dynamicRealm, long oldVersion, long newVersion) {
-
-        // Add schema increments and change methods when updating models
-
-        /* Example migration :
+    public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+        //updating schema version
+        RealmConfiguration config = new RealmConfiguration.Builder()
+                .schemaVersion(2)
+                .build();
```
iamareebjamal: This is unneeded. Look at OpenEventApplcation class, the config is used there. So I guess only DB_VERSION needs to be incremented

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
arp95: should the condition check be oldVersion == 0 or oldVersion == DB_VERSION ?
but if we keep the latter then i think as we update the models in the later stages we wont need the chnages to the realm db at each stage.

arp95:
@iamareebjamal 
made changes. please see.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
```diff
@@ -2,10 +2,11 @@
 
 import io.realm.DynamicRealm;
 import io.realm.RealmMigration;
+import io.realm.RealmSchema;
 
 public class RealmDatabaseMigration implements RealmMigration {
 
-    public static final long DB_VERSION = 0;
+    public static long DB_VERSION = 0;
```
iamareebjamal: This needs to be final and only changed in place, not during migration

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
arp95: outside the if statement? 
but if the value changes why do we create it as final?
sorry not getting your point.

iamareebjamal: The value 0 needs to be bumped at declaration only. It will not be reassigned, that's why final

0 is to be replaced by 1

arp95:
please check if this is the way.

iamareebjamal (approved):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
```diff
@@ -14,35 +15,20 @@
...
-
-        // Migrate to version 1: Add a new class.
-        if (oldVersion == 0) {
-            schema.create("Person")
-                .addField("name", String.class)
-                .addField("age", int.class);
-            oldVersion++;
-         }
+        RealmSchema schema = dynamicRealm.getSchema();
+        if(oldVersion == DB_VERSION) {
```
Shailesh351: I think condition will be false and if block will not run. Because DB_VERSION is now 1 and oldVersion was 0. 

@arp95 @iamareebjamal Please check again. Correct me if i am wrong

iamareebjamal: Yeah, here DB_VERSION should be replaced with 0

iamareebjamal (changes requested):

arp95:
sorry guys. changing it in a while.

arp95:
@iamareebjamal @Shailesh351 
please review now.

iamareebjamal (approved):

arp95:
@mariobehling @the-dagger 
please review and merge.

rohanagarwal94 (approved):

Shailesh351 (approved):

PR #1767, opened 2017-07-03
fix: modified Event.class to support new api

### LINKED ISSUE #1763: adding support for new attributes in Event model
**Actual Behaviour**

Currently the old model doesnt support certain new attributes of new model.

**Expected Behaviour**

Adding support for them.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes.

@iamareebjamal please add.
https://github.com/fossasia/open-event-android/milestone/32

arp95:
Fixes #1763 

Changes: 

made changes to the Event Model.

Screenshots for the change: 
-

@iamareebjamal @Shailesh351 @rohanagarwal94 
can you guys review this. it currently doesnt have the relationships as i am confused on what relationships to include in the android model. it has relationship for tickets also which i think is for orga app, so what relationships are important here?


iamareebjamal:
the copyright, callforpapers, licensedetails, etc

arp95:
@iamareebjamal
there are some attribuyes like "can-pay-by-paypal" and some more. are these meant for orga app and should we include them in the android part?

shreyasnbhat:
@arp95 I don't think an attribute like can-pay-by-paypal makes sense for the Android Project

android/app/src/main/java/org/fossasia/openevent/data/Event.java
```diff
@@ -23,356 +14,435 @@
...
+    @JsonProperty("can-pay-by-bank")
+    private boolean canPayByBank;
+    @JsonProperty("created-at")
+    private String createdAt;
+    @JsonProperty("bank-details")
+    private String bankDetails;
+    @JsonProperty("cheque-details")
+    private String chequeDetails;
 
-    @JsonSetter("thumbnail")
```
Shailesh351: Why removed JsonSetter annotations?

iamareebjamal:
You can leave the unnecessary attributes from class

arp95:
sure then.

android/app/src/main/java/org/fossasia/openevent/data/Event.java
arp95: i am using only those which the new api supports as its live now.
also i have made a separate issue for removing all the old api attribute names.

iamareebjamal: But this will remove support of the old json format...

arp95: i have opened an issue removing support for old api right. As the new api is up i am making he models as per new api.

iamareebjamal: You may not support the old API at all but you have to support the json export format. That was the entire point of testing in opev-android-sdk

Shailesh351: @arp95 Still we should not remove support for old api. What if someone has exported event as zip in old api format and want to generate app from it?

arp95: you are right. but my whole point of talking to niranjan was that will samples also change or not. when he said yes i thought we ought to change the models completely.
sorry if i got it wrong.

niranjan94: Okay. So I think there has been a misunderstanding. As far as the app is concerned, we need to support only the new API and the new json export format. (both the api and export format will share the same attribute names). 

We do not need to support the old api, or the old attributes. We don't have users for us to spend time on being backward compatible. 

If anyone does want to use a zip exporter in the old api format, we can maybe direct them to the conversion script that was made by Arpit I believe. That's it. No support for old api or attributes required. 

And hypothetically, even if we did require  support for old format, we should be normalizing the data in the generator and not making the app support both formats. 

Lemme know if any clarifications are needed :)

arp95: so guys apart from removing the pay attributes which relationships i should use?
the api has tracks, microlocations, sessions nd all, i seem a bit confused in that.

Shailesh351: I think it is not required.

Shailesh351 (changes requested):
Please refactor modified method names everywhere it is used 
example: getStartTime is changed to getStartsAt
In about fragment we are using getStartTime. So it needs to be changed also.

Shailesh351:
@arp95 Also change sample event file according to new Open Event Universal Format. I was just testing your PR and noticed it

arp95:
doing it.

iamareebjamal (approved):

iamareebjamal (changes requested):
Correct event sample file and test online please

iamareebjamal:
@arp95  I meant test offline and also check why the build is failing

arp95:
@iamareebjamal @Shailesh351 
please review now.

arp95:
looking at the build fail. please wait.

arp95:
@iamareebjamal 
i seem to not find "Version" in the new api. not using that seems to cause a lot of errors. 
this will make me change the code also so what do you suggest using in place of "version" ?

niranjan94:
@arp95 like I (and @shubham-padia) had posted in the android gitter channel, Use Etags instead of the version object for conditional network requests and cache validation/invalidation.

arp95:
@niranjan94 
yes i saw that. but neither in the api or wiki i find an implementation for it? can you please provide the link on how to use it?

niranjan94:
@arp95 that was also posted in the channel by @shubham-padia 

---

**What is an ETag?**
An entity tag (ETag) is an HTTP header used for Web cache validation and conditional requests from browsers for resources.

**Why is it needed?**
- Clients can make use of this and request complete data if and only if the data has changed else use their local cache.
- This can be used to ensure concurrency in the case of multiple clients trying to modify the same data at the same time.
- To know more refer: http://fideloper.com/api-etag-conditional-get

To know how to utilize it in our API, please refer the wiki here: https://github.com/fossasia/open-event-orga-server/wiki/Etags-implementation-in-the-API

arp95:
thanks. i just saw the link. sorry for not seeing that.
also @niranjan94 speakers-call are needed over here?
and also the speaker model is finalised?

shubham-padia:
@arp95 for concurrency control you can have a look at this link:
http://fideloper.com/etags-and-optimistic-concurrency-control

niranjan94:
@arp95 call for speakers has been internally renamed as speakers call ... (In UI you should still be showing it as call for speakers)

- Speaker model is incomplete. WIP. Ref issue: https://github.com/fossasia/open-event-orga-server/issues/3727 

@arp95 like I said in the introductory call, please follow issue/PRs on other related projects too. Especially the server. So that you know what is done and what isn't.

arp95:
@niranjan94 sure!

iamareebjamal:
@arp95 Just remove the versioning logic in this PR

Rest will be handled organically

arp95:
@iamareebjamal @Shailesh351 
guys have made changes. just commented the version logic and i know it doesnt make sense doing like this but etag will take time and hence i am opening another issue to solve seperately if thats fine.

iamareebjamal:
Don't just comment out, remove it as it will not be reused later. And of course, adding E-Tag support is not in the scope of this PR :)

arp95:
sure then.

arp95:
@iamareebjamal @Shailesh351 @rohanagarwal94 @shreyasnbhat 
please review.

android/app/src/main/java/org/fossasia/openevent/data/Event.java
```diff
@@ -21,358 +16,311 @@
...
-    private CallForPapers callForPapers;
-    private String locationName;
+    private String identifier;
     private String name;
-    private Copyright copyright;
-    private String privacy;
-    private String placeholderUrl;
-    @Relationship("social-links")
-    private RealmList<SocialLink> socialLinks;
+    private Double latitude;
```
iamareebjamal: Use primitives please

android/app/src/main/java/org/fossasia/openevent/data/extras/SpeakersCall.java
```diff
@@ -0,0 +1,52 @@
+package org.fossasia.openevent.data.extras;
+
+import com.github.jasminb.jsonapi.IntegerIdHandler;
+import com.github.jasminb.jsonapi.annotations.Id;
+import com.github.jasminb.jsonapi.annotations.Type;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+@Type("speakers-call")
```
iamareebjamal: Add a blank line between import statement

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
```diff
@@ -6,18 +6,19 @@
...
-            schema.get("Session")
+        if(oldVersion == 1) {
+            schema.get("Event")
                     .addField("created-at", String.class)
                     .addField("deleted-at", String.class)
-                    .addField("submitted-at", String.class)
-                    .addField("is-mail-sent", Boolean.class);
+                    .addField("original-image-url", String.class)
+                    .addField("large-image-url", String.class)
+                    .addField("icon-image-url", String.class);
```
iamareebjamal: Why removing old migration? It won't be removed. Please read realm migrations carefully. The comment example I added here also showed how to handle 2 version migrations. Both migrations will be present here. One for oldVersion == 0 and one for oldVersion == 1

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDatabaseMigration.java
iamareebjamal: And if renaming fields, migration is not needed. Just don't rename the fields in model, only the jsonproperty and getters and setters

android/app/src/main/java/org/fossasia/openevent/data/Event.java
arp95: i am using the default ones we used earlier. will change it to double if you say.

iamareebjamal: OK, let it be then

arp95:
@iamareebjamal @Shailesh351 @rohanagarwal94 @shreyasnbhat 
please review.

arp95:
guys please review.

Shailesh351 (approved):

Shailesh351:
@arp95 Please squash commits

arp95:
@Shailesh351 
i am currently outside. can you point out the error so that i can do without running itself?

the-dagger:
@arp95 
https://travis-ci.org/fossasia/open-event-android/builds/250666330#L2081

arp95:
@the-dagger thanks!
please review and merge!

android/app/src/main/java/org/fossasia/openevent/api/processor/EventListResponseProcessor.java
```diff
@@ -26,82 +25,19 @@ private void save(Event event) {
...
-                download.downloadMicrolocations();
-                Timber.d("Downloading Microlocations");
-
-                counterRequests++;
-            }
-
-            if (storedVersion.getSessionTypesVer() != version.getSessionTypesVer()) {
-                download.downloadSessionTypes();
-                Timber.d("Downloading Session Types");
+	counterRequests = 7;
```
iamareebjamal: Fix the indentation please

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/api/processor/EventListResponseProcessor.java
arp95: @iamareebjamal i am outside currently.
@Shailesh351 please add a tab to this while fixing or otherwise i will be able to send a pr by night.

arp95:
@the-dagger @iamareebjamal @Shailesh351 
please review and merge.

iamareebjamal (dismissed):

android/app/src/main/assets/event
```diff
@@ -1,109 +1,52 @@
...
-    "start_time": "2017-03-17T09:00:00+08:00",
-    "ticket_url": "https://eventyay.com/e/45da88b7/",
-    "topic": "Science & Technology",
-    "thumbnail": "https://storage.googleapis.com/eventyay.com/events/6/thumbnail/R0IzYXZ6bT/1481017202.91.jpg",
+    "identifier": "e4a55e79",
+    "name": "FOSSASIA 2016",
+    "latitude": 1.346,
+    "longitude": 103.686,
+    "location-name": "15 Science Centre Rd, Singapore 609081",
+    "starts-at": "2017-06-05T14:42:19+05:00",
```
Shailesh351: @arp95 Please update file. Just attribute names needs to be changed. Changing values will break sample. 
Here event date is changing so it will not show any sessions in Schedule Fragment because sessions date-time are being different. Also add social links which you have removed.

Shailesh351 (changes requested):

android/app/src/main/java/org/fossasia/openevent/data/Event.java
```diff
@@ -21,358 +16,313 @@
 public class Event extends RealmObject {
```
Shailesh351: Please Add `id` field and annotations to support JSONAPI.

android/app/src/main/assets/event
arp95: i didnt get you? i just added new attributes and removed the old ones.

Shailesh351: Before start time was : 2017-03-17T09:00:00+08:00
Now it is : 2017-06-05T14:42:19+05:00

But in sessions time is 2017-03-17 so sessions will not be listed in schedule fragment

arp95:
@iamareebjamal @Shailesh351 
please review.

Shailesh351 (approved):

arp95:
@mariobehling @the-dagger 
please review and merge.

iamareebjamal (approved):

arp95:
@the-dagger 
please merge.

PR #1749, opened 2017-06-29
fix : Improve map fragment

Shailesh351:
Fixes #1743 

Changes: 
- Fixed padding before autocomplete search view
- Changed autocomplete textview to SearchView.SearchAutoComplete to make UI consistent
- Showed marker info on clicking on suggestion
- Added null check for event and location
- Removed sharing option from fDroid version
- Clicking map icon from sesssion detail activity and location activity directly focus on corresponding location and shows marker info.

Screenshots for the change: 

videotogif_2017 06 29_18 48 32

PR #1777, opened 2017-07-06
feat : Implement marker clustering in map fragment

### LINKED ISSUE #1730: Implement marker clustering in map fragment
**Actual Behaviour**

Due to large number of microlocations in events like [Mozilla All Hands](https://github.com/fossasia/open-event/blob/master/sample/MozillaAllHands17/microlocations) , the map UI gets clumsy.

**Expected Behaviour**

The markers should cluster itself to on zoom out and decluster on zoom in.

**Screenshots of the issue**

Something like this.

mapactivity2

Yes after the priority issues are solved.

[milestone 7](https://github.com/fossasia/open-event-android/milestones/30)

rohanagarwal94:
Fixes #1730

Due to markers present at single location in the API, the markers will not decluster. Therefore attaching a screenshot with sample data as marker positions.
screenshot_2017-07-07-02-50-10-594_org fossasia openevent


android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
```diff
@@ -182,40 +191,58 @@ private void showLocationsOnMap() {
...
-                marker.setIcon(vectorToBitmap(getContext(), R.drawable.map_marker, R.color.color_primary));
-                return false;
-            });
-            mMap.setOnMapClickListener(latLng -> locationMarker.setIcon(
-                    vectorToBitmap(getContext(), R.drawable.map_marker, R.color.dark_grey)));
-        }
-        return locationMarker;
-    }
+//    private Marker handleMarkerEvents(LatLng location, String locationTitle) {
+//        if (mMap != null) {
```
arp95: is this comment just for testing purposes for now?

rohanagarwal94: @arp95 yes.

rohanagarwal94:
@iamareebjamal @arp95 @shreyasnbhat @Shailesh351 @mahikaw please review.

arp95 (approved):
LGTM.

android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.android.gms.maps.model.MarkerOptions;
+import com.google.maps.android.clustering.ClusterManager;
+import com.google.maps.android.clustering.view.DefaultClusterRenderer;
+
+import org.fossasia.openevent.R;
+
+/**
+ * Created by rohanagarwal94 on 4/7/17.
+ */
+public class ClusterRenderer extends DefaultClusterRenderer<MicrolocationClusterWrapper> {
```
iamareebjamal: Instead of creating a concrete class, this could have been accomplished using anonymous inner class. If the class is not to be reused, it's better to use anon classes

android/app/src/googleplay/java/org/fossasia/openevent/utils/ImageUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.android.gms.maps.model.BitmapDescriptor;
+import com.google.android.gms.maps.model.BitmapDescriptorFactory;
+
+/**
+ * Created by rohanagarwal94 on 6/7/17.
+ */
+public class ImageUtils {
+
+    public static BitmapDescriptor vectorToBitmap(Context context, @DrawableRes int id, int color) {
```
iamareebjamal: If there is a method for `setIcon(Drawable)`, then this won't be needed, please check

android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.google.android.gms.maps.model.LatLng;
+import com.google.maps.android.clustering.ClusterItem;
+
+import org.fossasia.openevent.data.Microlocation;
+
+/**
+ * Created by rohanagarwal94 on 6/7/17.
+ */
+public class MicrolocationClusterWrapper implements ClusterItem {
```
iamareebjamal: This can be done by a function instead

```Java
static ClusterItem getClusterItem(Microlocation microlocation) {
    return new ClusterItem {
        @Override 
        public LatLng getPosition() {
            return new LatLng(microlocation.getLatitude(), microlocation.getLongitude());
        }
    };
}

**Review by iamareebjamal — CHANGES_REQUESTED** [review]

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ImageUtils.java (hunk 3 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ImageUtils.java]: Thing is markers cant directly use vector drawables which was the purpose of creating it in the first place. - https://stackoverflow.com/a/44140446

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java (hunk 2 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: That would clutter the maps fragment class IMO. Anon class would have been better it was a class with lesser LOC.

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java (hunk 4 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java]: Cant. The clustermanager needs the parameter model to implement Cluster item . [see this](https://github.com/rohanagarwal94/open-event-android/blob/4641cffc4504bc374d4de7667f3dd5e41de70aad/android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java#L56).

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java (hunk 2 above)
Comment (iamareebjamal) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: Then add it in some confined package, utils is common package for the whole app

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java (hunk 4 above)
Comment (iamareebjamal) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java]: And the function I wrote above returns just that

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java (hunk 2 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: @iamareebjamal it is added in the google play flavour only. Please check again.

Comment (iamareebjamal) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: I'm talking about ClusterRenderer being in the utils package. It is not a app wide utility class so it doesn't belong there

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java (hunk 4 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java]: Also the microlocations class cannot implement cluster item since it shows error of realm access from incorrect thread.

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java (hunk 2 above)
Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: But this utils package is created in the google play package inside the source, just like the map fragment in the fragments package inside the google play package while the other fragments in the main package inside the fragments package. Please suggest the name of the package if it doesn't fit right.

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java (hunk 4 above)
Comment (iamareebjamal) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/MicrolocationClusterWrapper.java]: Ok, then move it to some specific package, utils package should be for app scope components

File: android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java (hunk 2 above)
Comment (iamareebjamal) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: The point is not about the build flavor, it's just about the architecture of your components. You have put them in utils where they are only used in 1 class. They are not utils for the whole app. They should be either along with fragment they are created for, or in a separate package. 

google play module is not a package, it's a build flavor which gets merged in the main module when the app is built

Comment (rohanagarwal94) [reply android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java]: @iamareebjamal Noted. Does mapUtils as a package work?

File: android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java [hunk 5]
```diff
@@ -180,42 +176,41 @@ private void showLocationsOnMap() {
         });
     }
 
-    private void focucOnMarker(Marker marker){
-        marker.showInfoWindow();
```
Shailesh351, line 184: Why are you removing it? 
it reverts on of the feature of https://github.com/fossasia/open-event-android/pull/1749 which is Showing marker info on clicking on suggestion from search

rohanagarwal94: The thing is cluster manager does not in any way return the reference of the markers created from the location items it is passed to. The markers created inside the class gets stored in a cache and gets rendered as per the view of the cluster. Therefore it never returned the full list of markers hence crashing the app. It is unfortunately something that remains to be implemented by it. For the time being what I have done is just focus the camera on the particular marker since that only requires the location coordinates. Hope I am clear.

android/app/src/googleplay/java/org/fossasia/openevent/utils/ClusterRenderer.java
iamareebjamal: package name should be all in lowercase, use utils/map

rohanagarwal94:
@iamareebjamal updated.

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
```diff
@@ -102,6 +101,12 @@ public void onMapReady(GoogleMap map) {
         if(map != null){
             mMap = map;
             mMap.getUiSettings().setMapToolbarEnabled(true);
+            clusterManager = new ClusterManager<>(getActivity(), mMap);
+            clusterRenderer=new ClusterRenderer(getActivity(), mMap, clusterManager);
```
iamareebjamal: Spaces around `=`

iamareebjamal (changes requested):

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
rohanagarwal94: Done.

iamareebjamal (dismissed):

rohanagarwal94:
@mariobehling @the-dagger please review.

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
Shailesh351: @rohanagarwal94 For fixing it you can add Marker property in MicrolocationClusterWrapper.
And set marker in onClusterItemRendered() method of ClusterRenderer.

And in map fragment use this marker to show info window. Use cancelableCallBack in animateCamera method to show info window I think you have to use handler with post Delayed method.

rohanagarwal94: @shailesh351 tried this before. The thing is in that case, the search only works if the marker is visible. If it is clustered, then the app crashes since the cache is cleared in that case.

Shailesh351: @rohanagarwal94 See https://github.com/Shailesh351/open-event-android/commit/bae48921b798970c01852a7f64c891d2f4e4811e It is working for me.

In my opinion clustering markers is not a good idea because in the event all the location will be nearby.

rohanagarwal94: Will pull your changes and check if it is. As far as the use is concerned, it totally boils down to the fact that whether the event has a lot of places to go. So when the organizer plans to include cafeteria, washrooms, and other markers it will clutter the map. See the above screenshot as an example.

rohanagarwal94: @shailesh351 thanks for helping out.

rohanagarwal94:
@iamareebjamal @Shailesh351 @shreyasnbhat @arp95 @mahikaw please review again.

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
```diff
@@ -52,12 +48,14 @@
     final private String SEARCH = "searchText";
 
     private GoogleMap mMap;
-    private Marker locationMarker;
+    private Marker locationMarker = null;
```
iamareebjamal: Why explicit assignment?

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
```diff
@@ -180,42 +177,65 @@ private void showLocationsOnMap() {
...
+                Handler handler = new Handler();
+                handler.postDelayed(() -> {
+                    Marker marker = stringMicrolocationClusterWrapperMap.get(locationName).getMarker();
+                    if (marker != null) {
+                        locationMarker.setIcon(ImageUtils.vectorToBitmap(getContext(), R.drawable.map_marker, R.color.dark_grey));
+                        locationMarker = marker;
+                        marker.showInfoWindow();
+                        marker.setIcon(ImageUtils.vectorToBitmap(getContext(), R.drawable.map_marker, R.color.color_primary));
+                    }
+                }, 500);
```
iamareebjamal: Why postDelayed?

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class ClusterRenderer extends DefaultClusterRenderer<MicrolocationClusterWrapper> {
+
+    private Context context;
+
+    public ClusterRenderer(Context context, GoogleMap map,
+                             ClusterManager<MicrolocationClusterWrapper> clusterManager) {
+        super(context, map, clusterManager);
+        this.context=context;
+    }
```
iamareebjamal: If you are holding a reference to context, make sure it is cleared onDestroy

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,39 @@
...
+ * Created by rohanagarwal94 on 4/7/17.
+ */
+public class ClusterRenderer extends DefaultClusterRenderer<MicrolocationClusterWrapper> {
+
+    private Context context;
+
+    public ClusterRenderer(Context context, GoogleMap map,
+                             ClusterManager<MicrolocationClusterWrapper> clusterManager) {
+        super(context, map, clusterManager);
+        this.context=context;
```
iamareebjamal: Format properly please

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ImageUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.android.gms.maps.model.BitmapDescriptorFactory;
+
+/**
+ * Created by rohanagarwal94 on 6/7/17.
+ */
+public class ImageUtils {
+
+    public static BitmapDescriptor vectorToBitmap(Context context, @DrawableRes int id, int color) {
+        Drawable vectorDrawable = ResourcesCompat.getDrawable(context.getResources(), id, null);
+        assert vectorDrawable != null;
```
iamareebjamal: assert?

iamareebjamal (changes requested):

android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java
rohanagarwal94: The marker reference is only possible after it is rendered which happens only after the zoom. Hence postDelayed was necessary to show the info window after the rendering.

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
rohanagarwal94: Nope, not holding a reference.

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ImageUtils.java
rohanagarwal94: to make sure the vector drawable is not null. Otherwise would have to check it at every implementation of the method. Any better way in your mind?

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
iamareebjamal: It was not a question :smile: You are holding a reference in this ClusterRenderer, is it lifecycle based component or not. If we don't know the answer, we must make the context in it null when the fragment is destroyed

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ImageUtils.java
iamareebjamal: Is there a case it'd be null?

rohanagarwal94: Cant think of it as such. So remove it then?

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
rohanagarwal94: oh yeah sorry.

rohanagarwal94:

Shailesh351 (changes requested):
@rohanagarwal94 App is crashing in this case.
- Click any on marker
- Zoom out till markers clusters
- Then zoom in till markers appears again
- Click on marker

Error : java.lang.IllegalArgumentException: Unmanaged descriptor on setIcon method of marker

rohanagarwal94:
@shailesh351 @iamareebjamal @arp95 @mahikaw @shreyasnbhat please review now.

rohanagarwal94:
@Shailesh351 @iamareebjamal @arp95 @mahikaw @shreyasnbhat please review now. Have to update branch again and again :smile:

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+    public boolean onClusterItemClick(MicrolocationClusterWrapper item) {
+        if (microlocationClusterWrapper != null) {
+            getMarker(microlocationClusterWrapper).setIcon(ImageUtils.vectorToBitmap(context, R.drawable.map_marker, R.color.dark_grey));
+        }
+        microlocationClusterWrapper = item;
+        getMarker(item).setIcon(ImageUtils.vectorToBitmap(context, R.drawable.map_marker, R.color.color_primary));
+        return false;
+    }
+
+    public void focusOnMarkers(Map<String, MicrolocationClusterWrapper> stringMicrolocationClusterWrapperMap, String locationName) {
```
Shailesh351: Only use MicrolocationClusterWrapper here no need to use whole Map<String, MicrolocationClusterWrapper>

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+                    if (microlocationClusterWrapper != null)
+                        getMarker(microlocationClusterWrapper).setIcon(ImageUtils.vectorToBitmap(context, R.drawable.map_marker, R.color.dark_grey));
+                    microlocationClusterWrapper = item;
+                    marker.showInfoWindow();
+                    marker.setIcon(ImageUtils.vectorToBitmap(context, R.drawable.map_marker, R.color.color_primary));
+                }, 100);
+            }
+
+            @Override
+            public void onCancel() {
```
Shailesh351, line 85: Add comment here like "// do nothing" or something appropriate. 
As a best practice we always try to reduce codacy issue. Every time if there is codacy issue than try to resolve it. Thanks :smile:

rohanagarwal94:
@shailesh351 updated.

Shailesh351 (approved):
Just remove unused imports

rohanagarwal94:
@iamareebjamal please review.

mahikaw (approved):

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    @Override
+    protected void onClusterItemRendered(final MicrolocationClusterWrapper clusterItem, Marker marker) {
+        super.onClusterItemRendered(clusterItem, marker);
+        clusterItem.setMarker(marker);
+   }
+
+    public void destroyContext() {
```
iamareebjamal: Make it `detach` or `cleanUp` and make context, map and clusterManager null. The practce is to not keep strong reference to objects passed from view in classes which can stay longer in memory (Listeners)

iamareebjamal (changes requested):

rohanagarwal94:
@iamareebjamal updated. Clustermanager does not requires to be null since I have not passed around its reference in the cluster renderer. It was required for parent's constructor.

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+    private Context context;
+    private GoogleMap map;
+    private MicrolocationClusterWrapper microlocationClusterWrapper;
+
+    public ClusterRenderer(Context context, GoogleMap map,
+                             ClusterManager<MicrolocationClusterWrapper> clusterManager) {
+        super(context, map, clusterManager);
+        this.context = context;
+        this.map = map;
+        setOnClusterItemClickListener(this);
```
iamareebjamal, line 30: Now, the strong reference to this class is passed to a listener, you must set it null on detach

iamareebjamal (changes requested):

android/app/src/googleplay/java/org/fossasia/openevent/utils/map/ClusterRenderer.java
rohanagarwal94: Got it. Updated.

iamareebjamal (approved):

rohanagarwal94:
@mariobehling @the-dagger please review.

PR #1788, opened 2017-07-08
feat : Add JSONAPI support

### LINKED ISSUE #1680: Implement support for json-api
The Open Event Server is moving towards a new API as per the [json-api specification](http://jsonapi.org/). Hence all the client apps need to support the same. 

You can checkout some of the Android client libraries that you can use to easily consume the API at http://jsonapi.org/implementations/#client-libraries-android. 

Please note that the zip format is not changing and only the API format is and the android app needs to support both. (Incase the user uploads a zip and generates an app.)

Shailesh351:
Fixes #1680 

Changes: 
- Added JSONAPI support
- Added null checks in fb feed implementation because it was crashing app while testing.


android/app/src/main/assets/config.json
```diff
@@ -1,5 +1,5 @@
 {
   "Email": "dev@fossasia.org",
   "App_Name": "Open Event",
-  "Api_Link": "https://eventyay.com/api/v1/events/6/"
+  "Api_Link": "https://api.eventyay.com/v1/events/69/"
```
niranjan94: This is still a test server. The database may get flushed anytime. Don't depend on a specific event please.

niranjan94 (changes requested):

android/app/src/main/assets/config.json
Shailesh351: @niranjan64 it is just sample.

niranjan94: @Shailesh351 but you seem to be referring to a specific event by ID in the the tests too. Please correct me if I'm wrong.

niranjan94 (changes requested):

android/app/src/main/assets/config.json
Shailesh351: For testing we need sample data so that we can confirm app is working fine fetching response from server. When app is generated using api endpoint then this api_link will change and app will use that accordingly.

niranjan94: @Shailesh351 I _am_ actually talking about testing ... Like I said ... the server is not dependable. The db can get flushed. And if it does, your tests and CI will fail here.

niranjan94: You should be using local [mock servers](https://apiblueprint.org/tools.html#mock+servers) on the CI environment.

Shailesh351:
@niranjan94  I have one doubt.
Currently base api url for particular event will be `/v1/events/{event_id}/` and event response can be fetched using `/v1/events/{event_id}` 

Retfrofit base url will be `/v1/events/{event_id}/`. For making request for event endpoint can't be null i.e. @Get("")

So currently I am extracting event_id from api_link and using it like @Get("../{id}?include=social_links").

if base url is like `https://abc.xyz/` or `https://abc.def/{id}` then it will create problem in extracting id.

So can we have one more endpoint for event `/v1/events/{event_id}/event`. Will it be as per JSONAPI spec ?

Shailesh351:
@niranjan94 also because of it i have to add sample event_id in test.
`APIClient.getOpenEventAPI().getEvent("69")`

niranjan94:
No. We cannot have an endpoint at `/v1/events/{event_id}/event`. That is not as per json-api spec and doesn't make any sense either. 

Event URL: `<domain-name>/v1/events/<event-id>`

The domain name and the event ID are the only parts that's gonna change right ? So I donno where a base URL like `https://<domain-name>/{id}` is gonna come from.

Shailesh351:
@niranjan94 in some sample in https://github.com/fossasia/open-event raw github url is used as api url.

niranjan94:
> in some sample in https://github.com/fossasia/open-event raw github url is used as api url.

@Shailesh351  if you'd like to do that, that's something the client ... (In this case the android app) should take care of. 

But I would recommend to run samples in offline mode.

android/app/src/main/assets/config.json
niranjan94: https://github.com/Aconex/drakov this is good. Tried it out. Also supports API Blueprint.

Shailesh351: @niranjan94 Sure thanks :smile:

niranjan94: @Shailesh351 this could be done in a separate issue if'd like since it's kinda out of the scope of this PR. (Be sure to create one ... I'll assign it to you)

niranjan94:
@Shailesh351 why is this still pending with no activity ? What is the status of this ? Ideally #1827 should have been implemented after this.

Shailesh351:
@niranjan94 Sorry for the delay. I was busy in https://github.com/fossasia/open-event-android/pull/1827. 

The problem it that in the app we need a sample which is live hosted on server because it is needed to test all the other features like tracks, sessions  etc. and https://api.eventyay.com/ is down. https://open-event-api.herokuapp.com is not having any complete sample.

I will resolve conflicts & complete it by today

niranjan94:
@Shailesh351 you cannot depend on a single instance. This is not a valid reason. I have told this before. Don't depend on either of the test instance. You can host your own instance if you want with one click on heroku. 

> open-event-api.herokuapp.com is not having any complete sample.

Then add a complete sample. The API is there. use it. Please refrain from giving such excuses.  This is really important and we're halfway through GSoC. yet, we don't even have a jsonapi base in the app yet.

android/app/src/androidTest/java/org/fossasia/openevent/APITest.java
```diff
@@ -76,7 +76,7 @@ public void onFailure(Call<List<Sponsor>> call, Throwable t) {
 
     public void testEventAPIResponse() throws Exception {
```
Shailesh351, line 77: @iamareebjamal @niranjan94 @rohanagarwal94 @shreyasnbhat @mahikaw I have one doubt. Does Android Instrumentation Test runs on Travic CI ? I searched on google but didn't found any appropriate answer.

iamareebjamal: They can run but they are not enabled currently

Shailesh351: Okay. I think it can be done in unit test. I will change it in separate PR :+1:

iamareebjamal: No, unit tests should not access network at all. It becomes an instrumentation test or E2E test then

Shailesh351:
@iamareebjamal @rohanagarwal94 @shreyasnbhat @arp95 @mahikaw Please review

android/app/src/main/java/org/fossasia/openevent/api/network/OpenEventAPI.java
```diff
@@ -20,26 +21,25 @@
...
-    @GET("sessions")
-    Call<List<Session>> getSessions(@Query("order_by") String orderBy);
+    @GET("sessions?include=microlocation,track&fields[microlocation]=name&fields[track]=name")
+    Call<List<Session>> getSessions();
 
-    //TODO:Correct event api url to server's
-    @GET("event")
-    Call<Event> getEvents();
+    @GET("../{id}?include=social_links,speakers_call,event_copyright")
+    Call<Event> getEvent(@Path("id") String eventId);
```
iamareebjamal: Make the id `int`

iamareebjamal (changes requested):
Is speakers API already included in the app?

Shailesh351:
@iamareebjamal  Yes speakers API is already included in the App (https://github.com/fossasia/open-event-android/pull/1817)

Shailesh351:
@iamareebjamal @shreyasnbhat @rohanagarwal94 @arp95 @mahikaw Please review :smile: 
PR https://github.com/fossasia/open-event-android/pull/1827 depends on this PR and issue https://github.com/fossasia/open-event-android/issues/1844 depends on https://github.com/fossasia/open-event-android/pull/1827.

iamareebjamal (approved):

mahikaw (approved):

PR #1805, opened 2017-07-10
feat : Add lombok to data classes

### LINKED ISSUE #1804: Add lombok to reduce boilerplate code in project
**Actual Behaviour**

Currently models in project have setters and getters, toString and other general POJO methods which are hard to maintain and increase complexity of the class.

**Expected Behaviour**

Use [Lombok](https://projectlombok.org/) which uses annotation processing to generate these methods.

Yes

Shailesh351:
Fixes #1804 

Changes: 
- Add [lombok](https://projectlombok.org/) for code generation


Shailesh351:
@arp95 @iamareebjamal @rohanagarwal94 @shreyasnbhat Please review

android/app/lombok.config
```diff
@@ -0,0 +1,2 @@
+lombok.anyConstructor.suppressConstructorProperties = true
+lombok.addGeneratedAnnotation = false
```
iamareebjamal: Add `lombok.getter.noIsPrefix = true` or Jackson will have problem getting properties with kebab naming strategy

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -306,7 +306,7 @@ public void onDrawerStateChanged(int newState) {
 
     private void setNavHeader(Event event) {
         String logo = event.getLogoUrl();
-        if (!logo.isEmpty()) {
+        if (logo != null && !logo.isEmpty()) {
```
iamareebjamal: TextUtils.isEmpty

android/app/src/main/java/org/fossasia/openevent/data/Microlocation.java
```diff
@@ -24,69 +30,10 @@
     private int floor;
     private String room;
 
-    public Microlocation() {
-    }
+    public Microlocation() {}
```
iamareebjamal, line 33: Use `@NoArgsConstructor`

android/app/src/main/java/org/fossasia/openevent/data/facebook/Comments.java
```diff
@@ -5,16 +5,10 @@
  */
 import java.util.ArrayList;
 
+import lombok.Data;
+
+@Data
 public class Comments {
 
     private ArrayList<CommentItem> data;
```
iamareebjamal, line 13: Someone please open the issue to change this class to just List<CommentItem> wherever this is called

android/app/src/main/java/org/fossasia/openevent/data/facebook/Feed.java
```diff
@@ -2,16 +2,10 @@
 
 import java.util.ArrayList;
 
+import lombok.Data;
+
+@Data
 public class Feed {
 
     private ArrayList<FeedItem> data;
```
iamareebjamal, line 10: Same as above

iamareebjamal (changes requested):
Also check local json deserialization without JSON ignore properties because boolean fields get missed sometimes

Also, someone please write tests (another PR) for local json parsing, they can reuse the code from Opev Android SDK project

android/app/src/main/java/org/fossasia/openevent/data/Microlocation.java
Shailesh351: I have already tried it but it gives error.
Error:Class "Microlocation" must declare a public constructor with no arguments if it contains custom constructors.

Microlocation class has one constructor with two arguments (int id, String name) i think because of it it is giving error.

android/app/lombok.config
Shailesh351: @iamareebjamal I don't think it will create any problem. We have declared property according to jsonapi and kebabspace. 

Setting `lombok.getter.noIsPrefix = true` will just rename getters and setters for boolean type.
if true then it uses get prefix and if false is
example,
`isBookmarked` is property of Session model. If `noIsPrefix` is `true` then getter for it will be `getIsBookmarked` else `isBookmarked`

Please tell which problem you trying to say.

Shailesh351: @iamareebjamal Got it Thanks :+1:

iamareebjamal: Did you get the point or should I explain? :smile:

android/app/src/main/java/org/fossasia/openevent/data/Microlocation.java
iamareebjamal: NoArgsConstructor should work even if there is a custom constructor, but if it's not working, then it's OK

android/app/lombok.config
Shailesh351: @iamareebjamal Thanks :smile:  i tried removing JSON ignore properties then tested it in both case with `noIsPrefix` true and false. And got the point

iamareebjamal: :+1:

iamareebjamal (approved):

rohanagarwal94 (approved):

shreyasnbhat (approved):

mahikaw (approved):

arp95 (commented):
LGTM.

PR #1806, opened 2017-07-10
feat: added logs option in the apk generator

### LINKED ISSUE #1798: Adding option for showing logs in apk-generator
**Actual Behaviour**

Currently the apk generator doesnt show logs incase the apk has some problem in it.

**Expected Behaviour**

Adding logs option so that user can view them in case the apk upload fails

-

arp95:
Fixes #1798 partially

Changes: 

Two changes i am focusing now:

1. UI change where i am adding logs option and adding properties to email field like autofocus and autocomplete

2. The code for displaying of logs on testing

Till now 1. is done.

Screenshots for the change: 
generator

@niranjan94 @iamareebjamal @shreyasnbhat @Shailesh351 
please keep on reviewing as its hard to make changes later.


shreyasnbhat:
@arp95 You could move the Logs below the button IMO.

arp95:
@niranjan94 @iamareebjamal 
should i add only the latest log message?

niranjan94:
@arp95 a better option would be a real-time or near real-time display of the log messages as and when they occur. (Like in the web application generator).

apk-generator/v2/app/static/js/main.js
```diff
@@ -55,6 +55,10 @@ $dataSourceRadio.change(
     }
 );
 
+$('#aLog').click(function(e) {
+    $('#buildLog').toggle();
```
niranjan94: Don't select the element each time `#aLog` is clicked. Select once. Store and re-use.

apk-generator/v2/app/templates/index.html
```diff
@@ -25,7 +25,7 @@
     <form class="col-md-offset-4 col-xs-offset-2 col-xs-8 col-md-4 form-group generator_form" method="post" id="form"
           name="form" role="form" enctype="multipart/form-data" autocomplete="off">
         <label for="email">Email</label>
-        <input type="email" class="form-control" id="email" name="email" required>
+        <input type="email" class="form-control" id="email" name="email" required autofocus autocomplete="on">
```
niranjan94: autocomplete="on" is not required. Just remove `autocomplete="off"` from form tag.

apk-generator/v2/app/templates/index.html
```diff
@@ -68,6 +69,16 @@
                 <span id="status-message" data-original="Building your customized android app"></span>
             </div>
 
+	    <div class="col-md-offset-3 col-xs-offset-1 col-xs-10 col-md-6" style="margin-top: -15px;">
+                <center>
+                    <div id="status" style = "margin-bottom:10px;"></div>
+                    <a id = "aLog" style = "cursor:pointer;"> -Logs- </a>
```
niranjan94: Give a better id. `aLog` doesn't make sense. And don't put space around `=`

arp95:
@niranjan94 
can you provide me a link on that. currently we are adding logger.add(info) which can be used for info messages while the self.update_status() can be regarded be as success and failure messages right?

so i have to show them both right?

apk-generator/v2/app/templates/index.html
```diff
@@ -68,6 +69,16 @@
...
+                    <pre id = "buildLog" style = "display:none;">
```
niranjan94: Use `hyphen-case` for IDs and class names

niranjan94 (changes requested):

apk-generator/v2/app/templates/index.html
```diff
@@ -68,6 +68,16 @@
                 <span id="status-message" data-original="Building your customized android app"></span>
             </div>
 
+	    <div class="col-md-offset-3 col-xs-offset-1 col-xs-10 col-md-6" style="margin-top: -10px;">
+   	        <center>
+                    <div id="status" style="margin-bottom:10px;"></div>
+                    <a id="log-message" style="cursor:pointer;"> -Logs- </a>
+		    <pre id = "build-log" style = "display:none;">
+		    </pre>
```
arp95: just saw that i missed space in this part. will send the correct version once the entore pr is over.

arp95:
should i make a separate ajax request to get the error logs. one think @niranjan94 already pointed out where making jsonify( state=state, info=str(info)) will give the error message as the status request. but we want them to display on logs click right? so a separate request has to be made?

iamareebjamal:
@arp95 I recommend you see the implementation in web app generator and take inspiration from it. What say? :smile:

arp95:
i know you are right. but i faced some problems getting he backend code. really sorry for that.

iamareebjamal:
~~That's fine. You can ask the guys on gitter, or me and we'll help you. Code resuse will help development time a lot and in future may be we'll be able to create an abstract wrapper for these generators encapsulating common functionalities~~

Did not realize it was using the different programming stack. Sorry

arp95:
no problem. it is taking time understanding the backend code. please give me time. in the mean time i am also solving other issues as well.

arp95:
@niranjan94 
is this the correct way of sending receiving messages?

apk-generator/v2/app/generator/generator.py
```diff
@@ -327,24 +329,37 @@ def run_command(self, command):
...
         return rc
 
+    @socketio.on('message')
+    def handle_message(message):
+        send(message)
+
     def generate_status_updates(self, output_line):
         if 'Starting process \'Gradle build daemon\'' in output_line:
+            handle_message('Starting gradle builder')
             self.update_status('Starting gradle builder', skip_log=True)
```
iamareebjamal: I have not looked into other files, but if there is no reference of update_status without the handle_message call, why not include it in the function itself?

iamareebjamal (changes requested):

apk-generator/v2/app/generator/generator.py
```diff
@@ -327,24 +329,37 @@ def run_command(self, command):
         rc = process.poll()
         return rc
 
+    @socketio.on('message')
```
niranjan94: This decorator is not needed. You have misunderstood how it works.

apk-generator/v2/app/generator/generator.py
niranjan94: I agree with @iamareebjamal

apk-generator/v2/app/static/js/main.js
```diff
@@ -17,15 +17,24 @@ var $fileProgressHolder = $("#file-progress"),
...
     pollingWorker = null,
-    downloadUrl = null;
+    downloadUrl = null,
+    socket = null;
+
+socket = io.connect('http://' + document.domain + ':' + location.port);
+
+socket.on('message', function(message) {
+    console.log("[x] Received    : " + message);
+    $buildLog.text('message')
```
niranjan94: You need to append. Not replace.

apk-generator/v2/app/static/js/main.js
```diff
@@ -17,15 +17,24 @@ var $fileProgressHolder = $("#file-progress"),
...
     taskId = null,
     pollingWorker = null,
-    downloadUrl = null;
+    downloadUrl = null,
+    socket = null;
+
+socket = io.connect('http://' + document.domain + ':' + location.port);
+
+socket.on('message', function(message) {
+    console.log("[x] Received    : " + message);
```
niranjan94: Remove all log statements at the end.

niranjan94 (commented):
Since you're anyway using sockets, use sockets for the current existing build status too. 

And also., check how it it works with parallel builds... I think the status messages will be mixed up between the two.

apk-generator/v2/app/static/js/main.js
arp95: yes thios is just to get update from you. will send everything in place once i get it right.

apk-generator/v2/app/generator/generator.py
arp95: this helps me to get reference in the js code. thats why used a decorator. am i wrong?

apk-generator/v2/app/generator/generator.py
arp95: you mean self.update_status() in the handle_message function right? @iamareebjamal 
will se how it works.

iamareebjamal: Opposite of that. handle_message in update_status

arp95:
@niranjan94 @iamareebjamal 
after reading about sockets i came about this conclusion.
if i use sockets, by using send() methods it wasnt a useful way because we arent connected to the client yet. Then i went through the emit() methods with the help of which i could receive the message in the client side. but the problem now was the socketio had no solution to which client it had to pass as the client wasnt setup. Then on the internet they suggested using the __init__.py socketio instance and send message through that but again it would broadcast the message and this could miz up the status messages and log messages.

So right now i am coming to  a conclusion of not using sockets as they are making things messier. i am thinking of connecting to the self.update_Status() in the client and then stroring the mesages over there to display as logs. the reason being i just need to display the messages in the generate_status_messages() as logs having success and failed. fine?

niranjan94:
@arp95 you really have to read more on how Socket.IO actually works. 

> So right now i am coming to a conclusion of not using sockets as they are making things messier. i am thinking of connecting to the self.update_Status() in the client and then stroring the mesages over there to display as logs. the reason being i just need to display the messages in the generate_status_messages() as logs having success and failed. fine?

Not fine. The whole point of logs, is to get a live log of what exactly is happening. (Just like how it is _already_ there in the webapp.). Things aren't getting messier. 

What you should be doing is, use namespaces to ensure the messages you send goto specific namespaces. After the generation starts, the client already gets a unique ID from the server. Connect to a namespace with that ID as the reference. Then in the back, send all status updates to the correct namespace, (since you already have the ID with you)

niranjan94:
You don't need to worry if a client is connected or not. You just send statuses. If a client is connected, it will get it. Else nothing will happen.

arp95:
@niranjan94 i am broadcasting the server message to all clients. can you please check the code if this is the way to go?
i know you said using namespaces to specific clients but isnt this better as we dont have many clients?

apk-generator/v2/app/generator/generator.py
```diff
@@ -327,6 +329,9 @@ def run_command(self, command):
         rc = process.poll()
         return rc
 
+    def handle_message(self, message):
+        socketio.emit('message', message)
```
niranjan94: Docs clearly say emit is for named events. We're just listening to the unnamed `message` event. So this should be `send` not `emit`. 

And what's the use of broadcasting to all clients ? I don't want to get the status updates of the app that you are generating (Given that both of us are using the generator at the same time). So, use namespaces and broadcast the status to the correct namespace so that only the clients connected to that specific namespace get the messages

niranjan94 (changes requested):

apk-generator/v2/app/generator/generator.py
arp95: is the self.identifier to be used as namespace?

niranjan94: Yep :)

arp95:
@niranjan94 
now i have used namespace, used the send() method which is sending the message from server as you can see in the logs.
but the clients are not receiving them. have tried various ways to put the socket.on fucntion but what is causing the client not receiving the message?

apk-generator/v2/app/static/js/main.js
```diff
@@ -245,4 +255,8 @@ $form.submit(function (e) {
         .catch(function () {
             showError();
         });
+    
+    socket.on('message', function(message) {
```
niranjan94: Socket is not defined in this scope. You need to bind it to the socket after it gets connected.

niranjan94 (changes requested):
The main problem here is the way you're expecting SocketIO to work. You should note that, the generation process is handled by a celery worker.

A celery worker runs decoupled from the the web server. But socket IO needs a web server instance to send/receive messages. You won't get that from within a celery task. 

You can have a very simple work around for this. Have a small proxy route. Call it, maybe `/send-message`. From within the celery task, make a simple POST request to that endpoint with the namespace you want to send to and the message you want to send. And in that route, do the usual `send` and it will work perfectly. (Tried it locally)

(Apart from this, there is a js error I have mentioned inline.)  

> **PS:** Apologies for the very late review. I was away from keyboard the whole day and couldn't properly review the code on mobile.

apk-generator/v2/app/static/js/main.js
arp95: you mean defining it within where socket is connected?
also apart from this it was working fine? 
now i should just present a success or fail in front of it right?

niranjan94: Yes ... where socket is connected.

> also apart from this it was working fine?

After what I said above also was done.

> now i should just present a success or fail in front of it right?

Lets do that once the basic implementation is done. Don't worry about that now.

arp95:
@niranjan94 
as you can see i tried establishing a post request so as to send and receive messages.
but it is not happening right now. what is going wrong can you please check?

apk-generator/v2/app/generator/generator.py
```diff
@@ -327,6 +331,13 @@ def run_command(self, command):
         rc = process.poll()
         return rc
 
+    def handle_message(self, message):
+        if message is not None:
+            request_json = {'identifier': self.identifier, 'message': message}
+            logger.info("Sending POST Request with: " + json.dumps(request_json))
+            request_url = 'http://localhost:8080' + url_for('api.send_notif')
```
niranjan94: Ensure port is not hard coded.

apk-generator/v2/app/static/js/main.js
```diff
@@ -117,6 +123,17 @@ function hideProgress() {
     $fileProgressHolder.hide();
 }
 
+function connectSocket() {
+  socket = io.connect(location.protocol + '//' + document.domain + ':' +
+      location.port + '/' + identifier);
+  console.log("Socket Connected");
+
+  socket.on('logs-message', function(message) {
```
niranjan94: Should be `message` not `logs-message`

apk-generator/v2/app/generator/generator.py
```diff
@@ -310,6 +313,7 @@ def update_status(self, state, exception=None, message=None, skip_log=False):
                 self.task_handle.update_state(
                     state=state, meta=meta
                 )
+        self.handle_message(message)
```
niranjan94: `message` is always `None` until the app generation is complete. You probably want to use `state` here. See comments in #1821

niranjan94 (changes requested):
@arp95 from the looks of it, you haven't even initialised Socket IO. 

Also, see comments inline.

apk-generator/v2/app/generator/generator.py
arp95: yes i did that for local purposes right now. will do that.

niranjan94:
With the modifications I mentioned above, the code works.

arp95:
@niranjan94 
please review. finally working.

logsshow

apk-generator/v2/app/templates/index.html
```diff
@@ -69,7 +69,17 @@
...
                 	</label>
             	</li>	
-	</ul>	
+	</ul>
+
+        <div class="col-md-offset-3 col-xs-offset-1 col-xs-10 col-md-6" style="margin-top: -5px;">
+   	        <center>
+                    <div id="status" style="margin-bottom:10px;"></div>
+                    <a id="log-message" style="cursor:pointer;"> -Logs- </a>
+		    <pre id="build-log" style="display:none;">
```
niranjan94: Put `</pre>` on the same line. 
Like `<pre id="build-log" style="display:none;"></pre>` ... Because, in a `pre` tag, all space matters and will be visible.

apk-generator/v2/app/templates/index.html
```diff
@@ -69,7 +69,17 @@
                     	<input type="radio" name="build-type" value="fdroid" required/> FDroid
                 	</label>
             	</li>	
-	</ul>	
+	</ul>
+
+        <div class="col-md-offset-3 col-xs-offset-1 col-xs-10 col-md-6" style="margin-top: -5px;">
+   	        <center>
+                    <div id="status" style="margin-bottom:10px;"></div>
```
niranjan94: There seems to be some spacing issue here. (Indentation)

arp95:
also updated the repetitive showing of logs. now only distinct logs are shown.
only thing remaining is the changing of status messages to the socket connection along with the logs connection.
right? @niranjan94

apk-generator/v2/app/views/api.py
```diff
@@ -12,6 +11,15 @@
...
 
+@api.route('/app/send-notif', methods=['POST', ])
+def send_notif():
+    request_json = request.get_json(force=True)
+    identifier = request_json['identifier']
+    namespace = '/' + identifier
+    message = request_json['message']
+    from app import socketio
+    socketio.send(data=message, namespace=namespace)
+    return jsonify(state = "SUCCESS")
```
niranjan94: No space around `=`

apk-generator/v2/app/static/js/main.js
```diff
@@ -117,6 +124,18 @@ function hideProgress() {
     $fileProgressHolder.hide();
 }
 
+function connectSocket() {
+  socket = io.connect(location.protocol + '//' + document.domain + ':' +
+      location.port + '/' + identifier);
+
+  socket.on('message', function(message) {
+      if(!arr.contains(message)) {
```
niranjan94: The backend should be handling this. Not the frontend.

apk-generator/v2/requirements.txt
```diff
@@ -8,4 +8,5 @@ android-asset-resizer
 pillow>=3.4.2,<3.5
 Flask-HTMLmin>=1.2,<2
 validators>=0.11.2,<0.12
-bleach>=1.5.0,<2
\ No newline at end of file
+bleach>=1.5.0,<2
+flask-socketio>=2.9.0
```
niranjan94, line 12: `eventlet` is also needed. Since we'll be using that as the worker with gunicorn.

apk-generator/v2/config.py
```diff
@@ -22,6 +22,7 @@ class Config:
     CELERY_BROKER_URL = os.environ.get('REDIS_URL', 'redis://localhost:6379/0')
     CELERY_RESULT_BACKEND = CELERY_BROKER_URL
     SECRET_KEY = os.environ.get('SECRET_KEY', 'JQgeEYk9b4VjxhJP')
+    PORT = os.environ.get('PORT', '8080')
```
niranjan94, line 25: Like i had said ... Also check for an env variable `ANDROID_GENERATOR_SERVICE_PORT`. 

Priority should be 
1. `PORT` env variable (user override)
2. `ANDROID_GENERATOR_SERVICE_PORT` env variable (auto-set for kubernetes deployments)
3. `8080` (default)

apk-generator/v2/app/views/api.py
```diff
@@ -12,6 +11,15 @@
 
 TASK_RESULTS = {}
 
+@api.route('/app/send-notif', methods=['POST', ])
```
niranjan94: Lets call this route,  `/_internal/socket/message` so in future we can deny access to all `/_internal/*` routes to the public.

niranjan94 (changes requested):
Comments added inline. 

Also the following modifications need to be made. 

-Update `apk-generator/v2/README.md` and `kubernetes/run.sh` with the correct gunicorn command to run using the eventlet worker

arp95:
@niranjan94 
can you check this?
now only 2 things remain:

1: for the repetition thing we have to do it in the client side. because if we do it in the backend then we might miss the status message in the client and then the message wont repeat from the server.

2. the download button doesnt work. can you tell me why where exactly the download url is? as if i am setting the download url to message from socket it gives a 404 error.

arp95:
@niranjan94 please review.

apk-generator/v2/requirements.txt
niranjan94: Not added ...

niranjan94 (changes requested):
@arp95 while sending from the backend, instead of sending just a plain string, send a json string that represents an object like 

```json
{
  "state": "success",
  "message": "<download-url>"
}
```

```json
{
  "state": "pending",
  "message": "<the log line or status message>"
}
```

```json
{ 
   "state": "failure",
   "message": "<the exception message>"
}
```

Parse this in the client side using `JSON.parse` and do what you have to do depending on the state

arp95:
was this the reason download url was incorrect?

apk-generator/v2/app/static/js/main.js
```diff
@@ -117,6 +123,46 @@ function hideProgress() {
     $fileProgressHolder.hide();
 }
 
+function connectSocketLogs() {
+  var socketLogs = io.connect(location.protocol + '//' + document.domain + ':' +
+      location.port + '/' + identifier + '/logs');
+
+  socketLogs.on('message', function(message) {
+	if (message === "SUCCESS") {
```
niranjan94: This. The code inside this wouldnt' even be executed right ? How will download url be equal to `SUCCESS`

niranjan94 (changes requested):

niranjan94:
> for the repetition thing we have to do it in the client side. because if we do it in the backend then we might miss the status message in the client and then the message wont repeat from the server

That's not a valid reason. As long as a client is connected to the server, all status messages will go through. We shouldn't be sending duplicates. So ensure the server sends only one stream of status messages. No duplicates please.

arp95:
@niranjan94
sorry i was having viral for a few days.
can we please solve the repititve log issue in spearate pr as it is creating a problem where all unique messages from server arent received by the client which does causes delay.

mariobehling:
Closing this due to inactivity.

PR #1809, opened 2017-07-11
feat : Improve social links implementation in About Fragment

### LINKED ISSUE #1792: Issue in Social link button in about fragment
**Actual Behaviour**

Currently social link for button is not implemented properly in about fragment. It is using index which is not good. it makes social button for particular event. If event changes than link is not mapped properly with button. I mean clicking on github button may lead to facebook page.
`socialLinks.get(2).getLink()`

**Expected Behaviour**

social links implementation should be dynamic.

Yes

Shailesh351:
Fixes #1792 

Changes:
- Remove Image buttons from about fragment and add recyclerview which displays social link buttons according to input data.


Shailesh351:
@arp95 @iamareebjamal @rohanagarwal94 @shreyasnbhat @mahikaw Please review.

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+        imageView.setOnClickListener(view -> {
+            if (socialLink != null && !Utils.isEmpty(socialLink.getLink())) {
+                setUpCustomTab(socialLink.getLink());
+            }
+        });
+    }
+
+    private void setUpCustomTab(String url) {
+
+        Uri uri = Uri.parse(url);
```
shreyasnbhat: Couldn't this function be added in a Utils class?

shreyasnbhat (changes requested):

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+    private SocialLink socialLink;
+    private Context context;
+
+    public SocialLinkViewHolder(View itemView, Context context) {
+        super(itemView);
+        ButterKnife.bind(this, itemView);
+
+        this.context = context;
+        imageView.setOnClickListener(view -> {
+            if (socialLink != null && !Utils.isEmpty(socialLink.getLink())) {
```
iamareebjamal, line 36: Immediately return if socialLink is null, without even setting the click listener

Better yet, assume SocialLink to be not null add a @NonNull annotation where SocialLink is getting attached, null social link being attached does not make sense and should crash the app so that we can address it as bug

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public void bindSocialLink(SocialLink socialLink) {
+        this.socialLink = socialLink;
+        setImageDrawable();
+    }
+
+    private void setImageDrawable() {
+        String name = this.socialLink.getName();
+        String link = this.socialLink.getLink();
```
iamareebjamal: Remove `this` and pass name and link as @NonNull parameters

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+
+        if (name == null || link == null) {
+            showView(false);
+            return;
+        }
```
iamareebjamal: Rather check using `TextUtils.isEmpty`

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private void showView(boolean show) {
+        if (show) {
+            imageView.setVisibility(View.VISIBLE);
+            layout.setVisibility(View.VISIBLE);
+        } else {
+            imageView.setVisibility(View.GONE);
+            layout.setVisibility(View.GONE);
+        }
+    }
```
iamareebjamal, line 69: Isn't imageView inside layout?

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
```diff
@@ -0,0 +1,113 @@
...
+                break;
+            case ConstantStrings.SOCIAL_LINK_LINKEDIN:
+                imageView.setImageDrawable(getDrawable(R.drawable.ic_linkedin_24dp));
+                break;
+            case ConstantStrings.SOCIAL_LINK_YOUTUBE:
+                imageView.setImageDrawable(getDrawable(R.drawable.ic_youtube_24dp));
+                break;
+            case ConstantStrings.SOCIAL_LINK_GOOGLE:
+                imageView.setImageDrawable(getDrawable(R.drawable.ic_google_plus_24dp));
+                break;
```
iamareebjamal: This code is duplicated in other classes, right?

Rather make a util method which returns a drawable ID when passed a name

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
Shailesh351: Yes it is but problem is that if i don't setVisibility of imageview to `GONE` then it still occupies space in recyclerview because of padding.

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
Shailesh351: @iamareebjamal I think it won't work because when initializing ViewHolder socialLink will always been null and it will not set onClickListener for any ViewHolder.

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
iamareebjamal: OK :+1:

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SocialLinkViewHolder.java
iamareebjamal: OK makes sense :+1:

iamareebjamal (approved):

shreyasnbhat (approved):

arp95 (approved):

PR #1812, opened 2017-07-12
feat : Add tests for JSON parsing

### LINKED ISSUE #1807: Add test for local JSON parsing
**Actual Behaviour**

Currently there is no test for local JSON parsing

**Expected Behaviour**

Add test for local JSON parsing

Yes

Shailesh351:
Fixes #1807 

Changes: 
- Add test for JSON parsing
- Removed  JSON property ignore annotations from method because it is redundant as object mapper is configured for ignoring it. https://github.com/fossasia/open-event-android/blob/development/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java#L671
Also for tests to actually work we should not add JSON ignore property annotation in model


Shailesh351:
@iamareebjamal @arp95 @rohanagarwal94 @shreyasnbhat Please review.

Also need some help here. How can i access `assets` in Test ? So that i can actually test sample event jsons.

shreyasnbhat:
Check this out.
https://stackoverflow.com/questions/26804651/how-to-access-files-from-assets-folder-during-tests-execution

iamareebjamal:
Using roboelectric will be an overkill and it'll take time. What you can do is copy the json in `tests/resources` and use standard class loader to load these

arp95:
these tests are only for samples created by me and mahika?
because in reality we an event organiser would enter the required info in the frontend and then he/she will export the json which i think would always be correct and in case of any errors it would be their job to check.
right?

Shailesh351:
@arp95 No it is not only for samples. I have added json strings according to Open Event Universal Format. Current model are able to parse it successfully. Suppose someone plays with Models in future then this tests will ensures that they don't break JSON parsing. 
In short these tests will ensure that models are able to parse Open Event Universal Format JSONs.

Shailesh351:
@iamareebjamal Is there any other way to directly use(Shared) Main source sets assets in Test. Because if i copy jsons to `tests/resources` then it will make sample inconsistent i mean we will need to maintain both resources.

arp95:
yes @Shailesh351.
also maintaining both resources shouldnt be a prob because if the format changes we ultimately have to change the test for that too

mahikaw:
@arp95 Please tag me by my username @-mahikaw, otherwise I don't get notifications about mentions. :)

Shailesh351:
@iamareebjamal @arp95 @rohanagarwal94 @shreyasnbhat Please review.

iamareebjamal (changes requested):
Great! But is there a requirement for manually written strings, they'll make tests hard to maintain

Shailesh351:
I thought same. Only Testing local json will be sufficient. Removing it. :+1:

Shailesh351:
@iamareebjamal @arp95 @mahikaw @rohanagarwal94 @shreyasnbhat Please review.

iamareebjamal (changes requested):
There are some tests required to see if attributes and relationships are correctly mapped, please add those

Shailesh351:
I think ObjectMapper will handle it. Can you please elaborate if i have missed something ?

iamareebjamal:
We have to test that the booleans are mapped correctly (due to lombok and naming strategy) and our mapper is handling relationships correctly

This is because right now it is working fine, but we don't want future additions to break the app.

android/app/src/test/java/org/fossasia/openevent/utils/JSONDeserializationTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static junit.framework.Assert.assertTrue;
+
+public class JSONDeserializationTest {
+
+    private ObjectMapper objectMapper;
+
+    @Before
+    public void setUp() {
+        objectMapper = new ObjectMapper();
```
iamareebjamal: Instead of this use the same mapper the main code is using and just set the failing on properties as true. This will make sure that the app is working correctly. This is just testing your code implementation

iamareebjamal (changes requested):

Shailesh351:
@iamareebjamal Updated

android/app/src/test/resources/assets/tracks
```diff
@@ -0,0 +1,36 @@
+[
```
iamareebjamal: I thought it could be done without copying JSON?

Shailesh351:
To test that Boolean attributes are mapped properly and relationships are  handled properly I have to add sample with some data. Because using fossasia sample from the assets will created problem on getting size of object in realtionship which is too large like track-sessions, speaker-sessions.

iamareebjamal:
But this will create a problem of maintaining test data too. What exactly is the problem with too large of a size?

Shailesh351:
@iamareebjamal Problem is that if we are testing sample speakers from assets for boolean and relationship then firstly we need to count how many speakers object it have. 
`assertEquals(1, speakers.size());`
And for boolean field like `is-featured` some speaker can have true value and some false. So it will create problem for `assertTrue(speakers.get(0).getIsFeatured());`. 

Regarding creating sample data for test it is very small containing only one or two objects so maintaining it will not create problem. If json format changes than we will just need to update fields.

iamareebjamal:
We don't have to verify size. We will just see statically that which session has which property and assert it. The thing is that I know that you will take care of updating the tests but what if in future someone comes and none of the maintainers remembers that test JSON have to be updated too. Tests will not fail but the app will have inconsistent data.

Don't assert size, just verify that the list is not empty and specific fields bound to be error prone

And even if we assert the size, it will be a good practice as it will keep the tests up to date with JSON and fail them if anyone tries to change JSON without updating them

iamareebjamal (approved):

rohanagarwal94 (approved):
Great :+1:

shreyasnbhat (approved):

PR #1815, opened 2017-07-13
fix : App crash on using bookmark widget

### LINKED ISSUE #1814: App crash on creating a bookmark widget
**Actual Behaviour**

The app crashes. The bookmark widget does not show the updated bookmarks from the app in the widget.

**Expected Behaviour**

The widget should work fine.

**Steps to reproduce it**
1. Add the bookmark widget in your phone.

**LogCat for the issue**

screenshot from 2017-07-13 20-22-41

Yes..

[milestone 11](https://github.com/fossasia/open-event-android/milestone/33)

rohanagarwal94:
Fixes #1814
Realm access from incorrect thread.

@iamareebjamal @arp95 @Shailesh351 @shreyasnbhat @mahikaw please review.


arp95 (approved):
LGTM.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
```diff
@@ -336,6 +336,7 @@ public Session getSession(String title) {
     }
 
     public RealmResults<Session> getBookMarkedSessionsSync() {
+        Realm realm = Realm.getDefaultInstance();
```
iamareebjamal: Close realm sessions whenever you create them

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
rohanagarwal94: Closing realm session does not result in updation of the bookmarks in the widget somehow.

iamareebjamal: Because the updates in the widget are dispatched by another mechanism and this is a synchronous query(non-updating). 

You should not rely on Realm to do it because leaking realm instance can lead to multiple databases being created. App data size can shoot up to 40 MB

rohanagarwal94:
@iamareebjamal please review. Found this the only plausible solution. Could the constructor be anyway changed in the session?

iamareebjamal:
You really don't need this. Just call `copyFromRealm` on the list of sessions in the widget service or provider

Secondly, this won't be needed because it is not the right way to do it, but still, you don't need to create this constructor, lombok will do that for you. Read lombok docs for further info in future.

rohanagarwal94:
@iamareebjamal the problem of incorrect thread then persists as per your suggestion.

iamareebjamal:
The problem is that you are expecting realm to update boomark widget, which is not correct thing to do. You can create a realm instance in the getBookmarkedSessionsSync method and close it before returning the list. It won't throw exception then

rohanagarwal94:
@iamareebjamal I tried that too, it returned empty. I tried converting it to the list in the getBookmarkedSessionsSync and it worked out fine to some extent. The major problem in that case was the randomness in updation of bookmarks which @shreyasnbhat talked about while adding/deleting them. That is why as a last resort I had to turn to cloning.

iamareebjamal:
This may be because at some places, the widget is being updated, and in other places it is not. Verify that each place is calling widget service to update

rohanagarwal94:
@iamareebjamal please review.

rohanagarwal94:
@shailesh351 @mahikaw @iamareebjamal @shreyasnbhat please review.

iamareebjamal (approved):

Shailesh351 (approved):

shreyasnbhat (approved):

PR #1823, opened 2017-07-15
feat:Added datewise organization of bookmarks in homescreen

### LINKED ISSUE #1810: Add date-wise split for bookmarks in homescreen
**Actual Behaviour**
Doesn't exist

**Expected Behaviour**
Could be made similar to the Google I/O App.

Yes

Milestone 10

shreyasnbhat:
Fixes #1810 

Changes

1. Modified GlobalSearchAdapter to use in the Homescreen.
2. GlobalSearchAdapter can be renamed as GenericAdapter.
3. Added date-wise splitting of bookmarks.

Screenshots for the change: 
ac5d8cf5-dafe-4cd2-a49a-66f00d2c31fe


shreyasnbhat:
@iamareebjamal @Shailesh351 @arp95 @rohanagarwal94 Please review.
I would need some help in what to use from DateUtils to convert this date string to the one with the day eg. Mon , 17 July , 2017

iamareebjamal:
@shreyasnbhat Create a new Date format string and its SimpleDateFormatter, add it in the switch case mapper, and use the getDateStringWithDefault with that String parameter

shreyasnbhat:
@iamareebjamal @arp95 @Shailesh351   What shall I do about the 2 headers? Shall I remove the Bookmark Header entirely? As there are 2 headers back to back for exmple. Bookmarks and then 17 Mar.

Shailesh351:
@shreyasnbhat You can keep Bookmarks header and for dates header make it gray as we have in DayScheduleFragment.

I would also suggest to use Sticky headers.

shreyasnbhat:
@Shailesh we can add sticky headers later. It would have to be implemented in the entire home screen for a consistent UI. 
For that to work we would have to get all the event-info items as ViewHolders and would involve major refactoring of code.

shreyasnbhat:
@iamareebjamal @Shailesh351 @arp95 Updated.
Screeshot
39a73269-57fc-46db-9b3d-f7bfc602e972

shreyasnbhat:
@Shailesh351 @iamareebjamal @arp95 Please review.

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -74,11 +76,12 @@
 
     private String searchText = "";
     private SearchView searchView;
+    final ArrayList<String> dateList = new ArrayList<>();
 
-    private DayScheduleAdapter bookMarksListAdapter;
+    private GlobalSearchAdapter bookMarksListAdapter;
```
Shailesh351: @shreyasnbhat Adding bookmarks in Global Search results is not a good idea. Bookmarks are specific to particular user and global search should give results from all over the app.

Shailesh351 (changes requested):
Instead of adding bookmarks in Global search adapter.
Create new BookmarksAdapter and implement stickeadersRecycleView. Return bookmarked session date as header id.
Look at other adapters for reference. It will simply take 6-7 line of code to add sticky headers in recyclerview

shreyasnbhat:
@Shailesh351 The global search adapter does give results from multiple sources. Bookmarks can very well be a source. I proposed that we rename Global search adapter to something like Generic Adapter as there would be un-necessary code duplication for no reason if I made another adapter called Bookmarks adapter. If I want to go ahead with adding sessions/bookmarked sessions in the global search this would be the best way to go. And for the addition of sticky headers I can do it but the question is that whether the UI will consistent with white background headers being sticky and the other red headers not being sticky. Additionally the red headers are all static in nature. Sticky headers just creates inconsistent UI here. What happens when we reach the end of the bookmarks? We would probably have a header with the text 19th Mar with no bookmarks below it which is wierd. 
In a gist I strongly feel that we shouldn't add sticky headers here. @iamareebjamal @arp95 please suggest.

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
shreyasnbhat: Global Search is meant for the user only. GlobalSearchAdapter can be renamed into GenericAdapter

shreyasnbhat:
@the-dagger @iamareebjamal @arp95 Please review and merge. This is a precursor to #1826

android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
```diff
@@ -90,9 +98,19 @@ public Object getItem(int position) {
                 resultHolder = new LocationViewHolder(location, context);
                 break;
             case DIVIDER:
-                View header = inflater.inflate(R.layout.search_result_type_header_format, parent, false);
+                View header;
+                if(context instanceof SearchActivity){
+                    header = inflater.inflate(R.layout.search_result_type_header_format, parent, false);
+                }
+                else{
```
iamareebjamal: else in previous line please

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -212,14 +215,44 @@ private void handleVisibility() {
         }
     }
 
+    private void loadEventDates() {
+
```
iamareebjamal: Remove extra line

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -74,11 +76,12 @@
 
     private String searchText = "";
     private SearchView searchView;
+    final ArrayList<String> dateList = new ArrayList<>();
```
iamareebjamal: private

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -74,11 +76,12 @@
...
 
-    private DayScheduleAdapter bookMarksListAdapter;
+    private GlobalSearchAdapter bookMarksListAdapter;
     private SocialLinksListAdapter socialLinksListAdapter;
     private RealmResults<Session> bookmarksResult;
-    private List<Session> mSessions = new ArrayList<>();
+    private List<Object> mSessions = new ArrayList<>();
```
iamareebjamal, line 84: Why object?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
shreyasnbhat: Global Search Adapter is used and as we also have the Date Headers. Hence, we need to use Object.

iamareebjamal: OK

shreyasnbhat:
@iamareebjamal  Updated.

shreyasnbhat:
@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -212,14 +215,43 @@ private void handleVisibility() {
...
+            for (String eventDate : dateList) {
+                boolean headerCheck = false;
+                for(Session bookmarkedSession : bookmarked){
+                    if(bookmarkedSession.getStartDate().equals(eventDate)){
+                        if(!headerCheck){
+                            String headerDate = "Invalid";
+                            try {
+                                headerDate = DateUtils.formatDay(eventDate);
+                            }
+                            catch (ParseException e){
```
iamareebjamal: Catch in previous line

iamareebjamal (changes requested):
Just this one change more

iamareebjamal (dismissed):

shreyasnbhat:
@arp95 @rohanagarwal94 @mahikaw Please review.

android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
```diff
@@ -90,9 +98,18 @@ public Object getItem(int position) {
                 resultHolder = new LocationViewHolder(location, context);
                 break;
             case DIVIDER:
-                View header = inflater.inflate(R.layout.search_result_type_header_format, parent, false);
+                View header;
+                if(context instanceof SearchActivity){
```
rohanagarwal94: Add space here before `{`

rohanagarwal94 (changes requested):

shreyasnbhat:
@rohanagarwal94 Updated.

shreyasnbhat:
@iamareebjamal @rohanagarwal94 @arp95 Updated guys. Please review. All minor changes done.

iamareebjamal (approved):

shreyasnbhat:
@rohanagarwal94 @arp95 Please review.

rohanagarwal94 (approved):

PR #1829, opened 2017-07-18
fix: Incorrect session details display from widget list item click

### LINKED ISSUE #1825: Incorrect redirect to session details on clicking a list item in the widget
**Actual Behaviour**

If the app is offline and the user clicks a different bookmarked session from the bookmark widget, he gets directed to the same session details of the first clicked list item.

**Expected Behaviour**

The user should be directed to the corresponding session details activity of the corresponding clicked list item.

**Steps to reproduce it**

1. Get the bookmark widget.
2. Close the app.
3. Click the first item on the widget list view.
4. Press the home button.
5. Click any other item in the list view of the widget.

Yes.

[milestone 11](https://github.com/fossasia/open-event-android/milestone/34)

rohanagarwal94:
Fixes #1825

Changes:
Also fixes the multiple stacking of activities when clicked on multiple list items.

@iamareebjamal @Shailesh351 @arp95 @shreyasnbhat @mahikaw please review.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
@@ -435,6 +435,10 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
 
         final SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
 
+        id = getIntent().getIntExtra(ConstantStrings.ID, 0);
+        title = getIntent().getStringExtra(ConstantStrings.SESSION);
+        trackName = getIntent().getStringExtra(ConstantStrings.TRACK);
+
         sessionById = realmRepo.getSession(id);
         sessionById.addChangeListener((RealmChangeListener<Session>) loadedSession -> {
             if(!loadedSession.isValid())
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -435,6 +435,10 @@ protected void onResume() {
 
         final SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
 
+        id = getIntent().getIntExtra(ConstantStrings.ID, 0);
```
Shailesh351, line 438: Getting this value from Extras creating problem in onCreate() method?
We are already  getting id in the onCreate() method. 
https://github.com/rohanagarwal94/open-event-android/blob/ec9f4ab95d166fddcd31b559b61b6adc376079ac/android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java#L145

rohanagarwal94: The problem is when the app is stopped, so does the lifecycle. Therefore when a list item is clicked, the onCreate method is never called and therefore the sessiondetailsactivity always showed the first clicked item details on any subsequent clicks. So it was necessary to fetch them from the intent in the onresume itself.

Shailesh351 (approved):

iamareebjamal (approved):
This is a good fix, but we need to remove the duplicate code in these classes. There shouldn't be same method calls in onCreate and onResume. Probably move everything in onResume. 

Probably in future

shreyasnbhat:
This was a neat fix. Nice work.

PR #1833, opened 2017-07-20
feat : Add SharedPreferences Util

### LINKED ISSUE #1832: Create Util for Shared Preferences to use it from all over the app
**Actual Behaviour**

Currently shared preference is used in different different activities and classes which increases complexity. 

**Expected Behaviour**

Create Util class for shared preference which will be used to save data in shared preferences all over the app. 

Yes

[Milestone 11](https://github.com/fossasia/open-event-android/milestone/34)

Shailesh351:
Fixes #1832 

Changes: Add SharedPreferencesUtil to use it from all over the App.


Shailesh351:
@iamareebjamal @rohanagarwal94 @arp95 @shreyasnbhat @mahikaw Please review

android/app/src/fdroid/java/org/fossasia/openevent/fragments/OSMapFragment.java
```diff
@@ -142,7 +138,7 @@ private void showEvent() {
             return;
 
         try {
-            int id = sharedPreferences.getInt(ConstantStrings.SESSION_MAP_ID,-1);
+            int id = SharedPreferencesUtil.getInt(ConstantStrings.SESSION_MAP_ID,-1);
```
iamareebjamal: Space after comma

android/app/src/main/java/org/fossasia/openevent/utils/SharedPreferencesUtil.java
```diff
@@ -0,0 +1,54 @@
+package org.fossasia.openevent.utils;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import org.fossasia.openevent.OpenEventApp;
+
+public class SharedPreferencesUtil {
+    private static SharedPreferences sharedPreferences;
+    private static SharedPreferences.Editor editor;
```
iamareebjamal, line 10: Storing context holders in static variables is a memory leak. Create a static method of `getInstance()` and return an instance of this object

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/utils/SharedPreferencesUtil.java
Shailesh351: @iamareebjamal I'm not storing context here. I am making static SharedPreference object which is returned by context.getSharedPreferences() method. It should not create a memory leak. Right ?

iamareebjamal: SharedPreference holds context as it is dependent on it, but here it is application context so the leak does not happen

iamareebjamal (approved):

rohanagarwal94 (approved):

shreyasnbhat (approved):

PR #1834, opened 2017-07-20
feat: Add theming to app generator

### LINKED ISSUE #1552: Ways letting user customise app on the genrator
**Actual Behaviour**

As discussed in this issue #1544 , we need a good way to let users customise the app and get it according to their needs.

**Expected Behaviour**

Work on different customise options.

**Steps to reproduce it**

-

**LogCat for the issue**

-

**Screenshots of the issue**

-

Yes but would like help from others as well.

iamareebjamal:
Fixes #1552 

Changes: 
- Add color pickers for primary, primary dark and accent color
- Send color data to backend
- Replace preset colors in colors.xml

Screenshots for the change: 
screenshot_20170720_173624
screenshot_20170720_173704
photo6305183553452877833


iamareebjamal:
@niranjan94 @arp95 @Shailesh351 @mahikaw @rohanagarwal94 @shreyasnbhat Please review

shreyasnbhat:
@iamareebjamal This is really cool.

apk-generator/v2/app/generator/generator.py
```diff
@@ -155,6 +156,12 @@ def generate(self, should_notify=True):
         replace(self.get_path("app/src/main/res/layout/nav_header.xml"), 'twitter', 'background')
         replace(self.get_path("app/build.gradle"), '"org.fossasia.openevent"', '"%s"' % self.app_package_name)
 
+        if self.theme_colors:
+            default_colors = colors()
+            self.update_status('Setting theme colors')
+            for key in default_colors.keys():
+                replace(self.get_path("app/src/main/res/values/color.xml"), default_colors[key], self.theme_colors[key])
```
niranjan94: Instead of searching for color and replacing, wouldn't it be better to parse the xml and change colors by key

niranjan94 (changes requested):

apk-generator/v2/app/generator/generator.py
iamareebjamal: Similar approach was used for replacing app name, that's why I used this. Will change to XML parsing

iamareebjamal: Should I use minidom library or xml.etree.cElementTree builtin ?

niranjan94: Yep. XML parsing is more reliable. Use the built in library. Don't think we need any complex parsing yet.

iamareebjamal:
@nirankan94 please review now

niranjan94 (approved):
Good work! :)

PR #1860, opened 2017-08-05
[WIP] feat : Refactor Adapter Using FastAdapter

### LINKED ISSUE #1859: Refactor the adapter of the app
**Actual Behaviour**

Current adapter does not support many features and crude implementations of custom features has made it error prone over time.

**Expected Behaviour**

Refactor the adapter of the app. Use FastAdapter to handle complex tasks for the app, like efficient sorting, filtering, animations, header views etc.

Yes

[Milestone 14](https://github.com/fossasia/open-event-android/milestone/37)

Shailesh351:
Fixes #1859 

Changes: 
- User FastAdapter for listing, filtering etc
- Added DividerItemDecoration in recyclerview instead of view element in xml.
- Created HeaderViewHolder

- [x] TracksAdapter
- [x] LocationsAdapter
- [ ] SponsorsAdapter
- [ ] SessionsAdapter
- [ ] SpeakerAdapter
- [ ] ScheduleAdapter
- [ ] FeedAdapter
- [ ] Others


android/app/build.gradle
```diff
@@ -107,6 +107,10 @@ dependencies {
     compile 'com.squareup:otto:1.3.8'
     compile 'com.jakewharton.timber:timber:4.5.1'
     compile 'com.jakewharton:butterknife:8.5.1'
+    compile('com.mikepenz:fastadapter:2.6.2@aar') {
+        transitive = true
+    }
+    compile 'com.mikepenz:fastadapter-commons:2.6.0@aar'
```
iamareebjamal: You can also add realm support module for FastAdapter

Shailesh351:
@iamareebjamal Using FastAdapter with Realm is making code base more complex and difficult to implement.

The issue i faced is `Realm access from Incorrect thread` (While filtering). To solve this i have to use copyFromRealm method to add data for Adapter and it causes delay in starting. Also it creates problem for   bookmarking as data used in recyclerview is copyFromRealm it is need to reload recyclerview.
Should we continue with current implementation ?

iamareebjamal:
Sorry for late reply. Yeah it's OK to continue with current implementation

Shailesh351:
Closing it. Using FastAdapter with Realm is making code base more complex and difficult to implement.

@iamareebjamal We planned to create common open-event-android-sdk which needs common database in the App. May be we can use FastAdapter in future after migrating to common database.

PR #1871, opened 2017-08-12
fix : Speaker TextDrawable when thumnail is null

### LINKED ISSUE #1870: Bug in speaker TextDrawable when thumbnail is null
**Actual Behaviour**

TextDrawable as speaker image placeholder has issue when thumbnail is null and internet connection is not there. It shows random drawable for it.
 
**Expected Behaviour**

It should work fine(Showing own name latters)

**Screenshot for the issue**
Check `FOSSASIA Team` drawable
screenshot-1502538259347

Yes

[Milestone 14](https://github.com/fossasia/open-event-android/milestone/37)

Shailesh351:
Fixes #1870 

**Changes:**
- Removed thumbnail null check while loading image so that Picasso can handle it by itself.( if thumbnail is null it will show drawable passed in the **error()** method)


iamareebjamal (approved):

Shailesh351:
@mariobehling Please merge

PR #1879, opened 2017-08-16
feat: Add threetenabp and refactor DateConverter

Shailesh351:
Fixes #1875 

Changes: 
- Add ThreeTenABP
- Modify Tests according to it
- Modify Date/Time implementation according to it.

PR #1882, opened 2017-08-17
fix : Upcoming/Ongoing issue

### LINKED ISSUE #1846: Bug in deciding upcoming for different time zones
**Actual Behaviour**

The current implementation of the feature doesn't take care of different time zones.

**Expected Behaviour**

The feature should take into account different time zones as well. 

Yes.

https://github.com/fossasia/open-event-android/milestone/35

Shailesh351:
Fixes #1846

Changes:
- Use default shared preference to get preferences from settings(Because settings preference uses default preference which is by default private)
- Add isUpcomingMethod

As we have added ThreeTenABP(https://github.com/fossasia/open-event-android/pull/1879)So All the Date conversions take care of TimeZone.


Shailesh351:
@iamareebjamal @heysadboy Please review

android/app/src/main/java/org/fossasia/openevent/utils/SharedPreferencesUtil.java
```diff
@@ -10,7 +10,7 @@
     private static SharedPreferences.Editor editor;
 
     static {
-        sharedPreferences = OpenEventApp.getAppContext().getSharedPreferences(ConstantStrings.FOSS_PREFS, Context.MODE_PRIVATE);
+        sharedPreferences = PreferenceManager.getDefaultSharedPreferences(OpenEventApp.getAppContext());
```
iamareebjamal, line 13: Why not FOSS_PREFS?

Shailesh351: All the settings preferences are stored in the default preferences. If we use different preference than we will not able to get values from default. Problem I have faced was at https://github.com/fossasia/open-event-android/blob/development/android/app/src/main/java/org/fossasia/openevent/OpenEventApp.java#L111 It always returns `false` if we use other preference than default preference

iamareebjamal (approved):

scandindian (approved):
Works great!

PR #1918, opened 2017-09-17
fix : Do not show stale notifications

### LINKED ISSUE #1917: Do not show stale notifications
**Actual Behaviour**

Notifications are shown for past sessions when a user bookmarks it

**Expected Behaviour**

Notifications should only be shown for upcoming events

**Steps to reproduce it**

Bookmark any past event

No

manav-mj:
Fixes #1917 

Changes: Added if statement to set alarm only when the event time is greater than the current time


android/app/src/main/java/org/fossasia/openevent/utils/NotificationUtil.java
```diff
@@ -37,12 +37,16 @@ public static Completable createNotification(Session session, Context context) {
...
-            Intent myIntent = new Intent(context, NotificationAlarmReceiver.class);
-            myIntent.putExtra(ConstantStrings.SESSION, session.getId());
-            PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 0, myIntent, PendingIntent.FLAG_UPDATE_CURRENT);
-
-            AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
-            alarmManager.set(AlarmManager.RTC, zonedDateTime.toInstant().toEpochMilli(), pendingIntent);
-        }).doOnComplete(() -> onSuccess(session)).doOnError(throwable -> onError(throwable, session));
+            ZonedDateTime currentZonedDateTime = ZonedDateTime.now();
+            // Checking if event time is greater than current time
+            if (zonedDateTime.toInstant().toEpochMilli() > currentZonedDateTime.toInstant().toEpochMilli()) {
```
iamareebjamal: There are methods like `isAfter` and `isBefore` 

Please use those :smile:

android/app/src/main/java/org/fossasia/openevent/utils/NotificationUtil.java
```diff
@@ -37,12 +37,16 @@ public static Completable createNotification(Session session, Context context) {
...
+                Intent myIntent = new Intent(context, NotificationAlarmReceiver.class);
+                myIntent.putExtra(ConstantStrings.SESSION, session.getId());
+                PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 0, myIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+
+                AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
+                alarmManager.set(AlarmManager.RTC, zonedDateTime.toInstant().toEpochMilli(), pendingIntent);
+            }else onError(new Throwable("Event time before current time"), session);
```
iamareebjamal: No need to throw error. Just log and don't set alarm

iamareebjamal: An example would be to just return after logging if the session is stale

iamareebjamal (changes requested):

manav-mj:
@iamareebjamal pls review again.

android/app/src/main/java/org/fossasia/openevent/utils/NotificationUtil.java
```diff
@@ -37,12 +37,15 @@ public static Completable createNotification(Session session, Context context) {
...
                 zonedDateTime.minusHours(10);
             }
-            Intent myIntent = new Intent(context, NotificationAlarmReceiver.class);
-            myIntent.putExtra(ConstantStrings.SESSION, session.getId());
-            PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 0, myIntent, PendingIntent.FLAG_UPDATE_CURRENT);
-
-            AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
-            alarmManager.set(AlarmManager.RTC, zonedDateTime.toInstant().toEpochMilli(), pendingIntent);
+            // Checking if the event time is before the current time
+            if (zonedDateTime.isBefore(ZonedDateTime.now())) {
```
iamareebjamal: It should be the opposite logic. If the event time is after current time, the notification should be shown

android/app/src/main/java/org/fossasia/openevent/utils/NotificationUtil.java
```diff
@@ -37,12 +37,15 @@ public static Completable createNotification(Session session, Context context) {
...
+                Intent myIntent = new Intent(context, NotificationAlarmReceiver.class);
+                myIntent.putExtra(ConstantStrings.SESSION, session.getId());
+                PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 0, myIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+
+                AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
+                alarmManager.set(AlarmManager.RTC, zonedDateTime.toInstant().toEpochMilli(), pendingIntent);
+            }else Timber.i("Event is finished");
```
iamareebjamal: Please add a space after curly bracket and wrap the logging statement in curly braces as well.

Also, use `Timber.d` instead as this is a debugging message, and for the message you may use

`Session is finished. Skipping showing notification` so that the developers are aware of the context :smile:

iamareebjamal (changes requested):
Also, please squash the commits into a single one :+1:

iamareebjamal (approved):

scandindian (approved):

PR #1952, opened 2017-10-14
fix: Fixed crashes caused by invalid realm objects.

### LINKED ISSUE #1951: Multiple nullpointerexceptions in app.
**Actual Behaviour**
`EditProfileActivity` and `OSMapFragment` have bug similar to the one mentioned in #1949 

**Expected Behaviour**
These crashes due to nullpointerexception must be fixed.

**Steps to reproduce it**
1. Uninstall and reinstall the app.
2. Dismiss the dialog which asks to download event schedule.
3.1 Open `OSMapFragment` and app will crash.
3.2 Open `EditProfileActivity` and press back key, the app will crash.

**LogCat for the issue**

Provide logs for the crash here
`E/AndroidRuntime: FATAL EXCEPTION: main
                                                                        Process: org.fossasia.openevent, PID: 20819
                                                                        java.lang.IllegalStateException: Object is no longer managed by Realm. Has it been deleted?
                                                                            at io.realm.internal.InvalidRow.getStubException(InvalidRow.java:183)
                                                                            at io.realm.internal.InvalidRow.isNull(InvalidRow.java:159)
                                                                            at io.realm.EventRealmProxy.realmGet$latitude(EventRealmProxy.java:331)
                                                                            at org.fossasia.openevent.data.Event.getLatitude(Event.java:31)
                                                                            at org.fossasia.openevent.fragments.OSMapFragment.showEvent(OSMapFragment.java:163)`

I will send a PR for this issue.

m-murad:
Fixes #1951

Changes: Fixed nullpointerexceptions crashes.


android/app/src/main/java/org/fossasia/openevent/activities/auth/EditProfileActivity.java
```diff
@@ -186,15 +186,16 @@ protected void onDestroy() {
 
     @Override
     public void onBackPressed() {
-        if (user == null || firstNameInput == null || lastNameInput == null) {
+
+        String firstName = firstNameInput.getText().toString().trim();
+        String lastName = lastNameInput.getText().toString().trim();
+
+        if (!user.isValid() || "".equals(firstName) || "".equals(lastName)) {
```
iamareebjamal: `TextUtils.isEmpty()` would look better here

iamareebjamal (changes requested):
Nice. Thanks

Just one recommendation for empty string checking

android/app/src/main/java/org/fossasia/openevent/activities/auth/EditProfileActivity.java
m-murad: Agreed.
Updating PR.

iamareebjamal (approved):

PR #1992, opened 2017-11-11
Fix #1989 fab moves up when snackBar messages are displayed

### LINKED ISSUE #1989: SnackBar hides floatingActionButton in schedule fragment
**Actual Behaviour**

If we add or remove a bookmark in schedule fragment,the snackBar hides the floatingActionButton

**Expected Behaviour**

The floatingActionButton should shift upwards when a snackBar message is displayed

**Steps to reproduce it**

Go to schedule fragment and add a bookmark

**Screenshots of the issue**

screenshot_20171110-234314

Yes

nikit19:
Fixes #1989 

Changes: fab moves up when snackBar messages are displayed

Screenshots for the change: 
ezgif com-video-to-gif 21

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
@@ -41,7 +41,6 @@ public class ScheduleFragment extends BaseFragment {
 
     @BindView(R.id.viewpager) ViewPager viewPager;
     @BindView(R.id.tabLayout) TabLayout scheduleTabLayout;
-    @BindView(R.id.schedule_fab_filter) FloatingActionButton fabFilter;
     @BindView(R.id.filter_text) TextView filtersText;
     @BindView(R.id.close_filter) ImageView closeFilterBarButton;
     @BindView(R.id.filter_bar) LinearLayout filterBar;
@@ -63,7 +62,8 @@ public class ScheduleFragment extends BaseFragment {
         setHasOptionsMenu(true);
 
         View view = super.onCreateView(inflater, container, savedInstanceState);
-
+        FloatingActionButton fabFilter = (FloatingActionButton) getActivity().findViewById(R.id.schedule_fab_filter);
+        fabFilter.setVisibility(View.VISIBLE);
         filterBar.setVisibility(View.GONE);
         OpenEventApp.getEventBus().register(true);
         compositeDisposable = new CompositeDisposable();
@@ -72,6 +72,8 @@ public class ScheduleFragment extends BaseFragment {
 
         setupViewPager(viewPager);
         scheduleTabLayout.setupWithViewPager(viewPager);
+
+        fabFilter.setOnClickListener(v -> filterSchedule());
 
         return view;
     }
@@ -164,7 +166,6 @@ public class ScheduleFragment extends BaseFragment {
         return super.onOptionsItemSelected(item);
     }
 
-    @OnClick (R.id.schedule_fab_filter)
     public void filterSchedule() {
         final AlertDialog.Builder dialogSort = new AlertDialog.Builder(getActivity())
                 .setTitle(R.string.dialog_filter_title)
@@ -211,6 +212,8 @@ public class ScheduleFragment extends BaseFragment {
             compositeDisposable.dispose();
         if(viewPager != null && onPageChangeListener != null)
             viewPager.removeOnPageChangeListener(onPageChangeListener);
+        FloatingActionButton fabFilter = (FloatingActionButton) getActivity().findViewById(R.id.schedule_fab_filter);
+        fabFilter.setVisibility(View.GONE);
     }
 }
 
--- a/android/app/src/main/res/layout/fragment_schedule.xml
+++ b/android/app/src/main/res/layout/fragment_schedule.xml
@@ -61,17 +61,5 @@
             android:background="@android:color/white"
             android:layout_below="@+id/filter_bar"/>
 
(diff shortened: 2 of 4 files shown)


android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -41,7 +41,6 @@
 
     @BindView(R.id.viewpager) ViewPager viewPager;
     @BindView(R.id.tabLayout) TabLayout scheduleTabLayout;
-    @BindView(R.id.schedule_fab_filter) FloatingActionButton fabFilter;
```
scandindian, line 44: Why are you removing its global declaration and initializing variable multiple times in separate functions?

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -63,7 +62,8 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         setHasOptionsMenu(true);
 
         View view = super.onCreateView(inflater, container, savedInstanceState);
-
+        FloatingActionButton fabFilter = (FloatingActionButton) getActivity().findViewById(R.id.schedule_fab_filter);
```
scandindian, line 65: Use BindView

nikit19:
@heysadboy I guess we can't use BindView when we are accessing a parent view from the fragment.
https://github.com/JakeWharton/butterknife/issues/578
and I get a null pointer exception if I declare it globally so had to declare it multiple times. Any suggestions?

android/app/src/main/res/layout/fragment_schedule.xml
```diff
@@ -61,17 +61,5 @@
...
-            android:id="@+id/schedule_fab_filter"
-            android:layout_width="@dimen/fab_width"
-            android:layout_height="@dimen/fab_height"
-            android:layout_alignParentBottom="true"
-            android:layout_alignParentEnd="true"
-            android:layout_alignParentRight="true"
-            android:layout_marginBottom="@dimen/layout_margin_large"
-            android:layout_marginEnd="@dimen/layout_margin_large"
-            android:layout_marginRight="@dimen/layout_margin_large"
-            app:srcCompat="@drawable/ic_filter_24dp"/>
```
iamareebjamal, line 74: All problems are solved if you don't move this from fragment to activity and use fragment view to show snackbar instead

iamareebjamal (changes requested):

android/app/src/main/res/layout/fragment_schedule.xml
nikit19: @iamareebjamal  The reason why I moved it to the activity is because coordinate layout was used there and the fab moves up automatically in that layout. I am not able to do the same thing in fragment because the snackBar messages are given from the SessionsListAdapter. Any suggestions ?

iamareebjamal: You can use CoordinatorLayout in Fragment itself and use that view to show snackbar

nikit19: ` View view = LayoutInflater.from(context).inflate(R.layout.fragment_schedule, null);
        CoordinatorLayout coordinatorLayout = (CoordinatorLayout)view.findViewById(R.id.coordinate_layout_fab_filter);`

@iamareebjamal After writing this in the DayScheduleViewHolder I pass this coordinatorLayout in the snackBar but the snackBar messages are not displayed.
Can you please tell me how should I solve this
Thank you.

iamareebjamal: Can you check if you have called `show()`?

Also, if you don't want, you can not add coordinatorlayout in fragment and use the one in activity by `getActivity().findViewById` in `onCreateView` method

nikit19: yes show() was called

@iamareebjamal this is exactly what I have done in the current PR and it works perfectly. Can you please see
Thank you

iamareebjamal: You have also moved the FAB from fragment to activity

nikit19: @iamareebjamal I need the coordinator layout in the DayScheduleViewHolder because the snackbar messages are given from there so I can't use `getActivity().findViewById`. I don't why even after inflating the layout correctly ,the snackBar messages are not displayed.

Is it a wrong practice to store the fragment's FAB in the activity?
Thank you

iamareebjamal: You can store the fab in Activity, but then show the snackbar using activity, and not fragment. So use the callback communication mechanism from fragment to activity

the-dagger:
@nikit19 Any updates?

nikit19:
@the-dagger the current PR works fine but when I tried not moving the fab from fragment to activity. I was not able to achieve the same results. As the snackbar messages were given from the DayScheduleViewHolder like this `Snackbar.make(slot_content, R.string.added_bookmark, Snackbar.LENGTH_SHORT).show();
`

iamareebjamal:
@nikit19 Status? We have a lot of PRs here which are almost complete and unmerged. Let's help clean the backlog up.

As mentioned above, you can store the FAB in the activity, but then please use a callback from fragment to activity to trigger the call.

We can have a discussion here regarding the implementation if there's any problem

nikit19:
I guess this is no longer required. Closing it

PR #1996, opened 2017-11-14
Fix #1986-Provide option to download data in the app.

### LINKED ISSUE #1986: No option to download data once user says no
**Actual Behaviour**

If the user selects no for downloading data at start up, there is no option to download it again.
Only option is to restart the app.

**Expected Behaviour**

There should be an option to download data in the overflow menu.

**Steps to reproduce it**

Open the app and select no for downloading data.

Yes

mayank8318:
Fixes #1986 

Changes: Add a menu item for downloading data

Screenshots for the change: 
23419125_1867275683301237_8488570078164418560_n

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
@@ -26,9 +26,6 @@ import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBar;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.Toolbar;
-import android.text.Html;
-import android.text.method.LinkMovementMethod;
-import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.MotionEvent;
@@ -36,7 +33,6 @@ import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.ImageView;
-import android.widget.TextView;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.squareup.otto.Subscribe;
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -417,7 +417,7 @@ private void downloadPageId() {
         }
     }
 
-    private void setupConnection() {
+    public void setupConnection() {
```
scandindian: This works but it's not a good practice. :-/

scandindian (approved):

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
mayank8318: Should I move it to an utils class and make it a static method? But it will have to be public still.

scandindian: Yes that'd be better .

mayank8318: @heysadboy I tried doing that but there is a long list of chained calls to different methods all of which are private to MainAcitivity and non-static. Some of them also affect the UI. I am facing problems transferring all this to a Utils class. How should I go about it?
Thanks :)

iamareebjamal: How about renaming this method to reflect that this means dowloading data? 😄

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
mayank8318: Yes that would be best. On it.

mayank8318:
@iamareebjamal Changed the name to ` downloadData() `

iamareebjamal (approved):

PR #2008, opened 2017-12-03
Fix #2005-Provide Ascending/Descending option in sorting.

### LINKED ISSUE #2005: Provide Ascending/Descending option in sorting
**Actual Behaviour**

Currently there is no option to select the order of sort (Ascending or Descending).

**Expected Behaviour**

There should be an option to select the order of sorting.

**Steps to reproduce it**

Open Schedule and click sort from the menu.

**Screenshots of the issue**

image

Yes

mayank8318:
Fixes #2005 

Changes: Add buttons to Sort dialog box for ascending / descending option.

Screenshots for the change: 
image

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
+++ b/android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
@@ -82,7 +82,7 @@ public class DayScheduleAdapter extends BaseRVAdapter<Session, DayScheduleViewHo
     public void filter(String constraint) {
         final String query = constraint.toLowerCase(Locale.getDefault());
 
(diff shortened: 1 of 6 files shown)


scandindian (changes requested):
I think it would look more minimalistic if we use icon button instead of text for ascending and descending.

mayank8318:
@heysadboy Didn't get you. You mean it should be a part of the `menu` ?

scandindian:
@mayank8318 use image button instead of text. Like ↑ for ascending. What's your view about it?

mayank8318:
@heysadboy Using Image Button in place of a text buttons in an alert dialog does't look intuitive for the user. Maybe we can use the arrows in the menu or something like in any normal website or sorting thing?
But I guess the text buttons looks better.

scandindian (approved):

iamareebjamal (approved):

PR #2029, opened 2017-12-21
Added LocationFragment ViewModel class

### LINKED ISSUE #2027: Create ViewModel for LocationsFragment
**Actual Behaviour**

There is no ```ViewModel``` for the ```LocationsFragment```

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes!

NA

**Other information**

Parent issue #2015

dr0pdb:
Fixes #2027 

Changes: Added ```ViewModel``` class for the ```LocationsFragment```.
Moved the database related code to the ```ViewModel```.


android/app/src/main/java/org/fossasia/openevent/viewmodels/LocationsFragmentViewModel.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.List;
+
+import io.realm.RealmResults;
+
+/**
+ * Created by srv_twry on 20/12/17.
+ */
+
+public class LocationsFragmentViewModel extends ViewModel{
```
iamareebjamal: Please remove the template comment before committing please 😄

iamareebjamal (changes requested):
More logic can be moved from the fragment so that most code is in the viewmodel instead of fragment. This includes state like the search query etc

dr0pdb:
@iamareebjamal  I have moved the search query state to the view model but that was the only thing which i was able to move from the fragment to the ViewModel.

1. RecyclerView code can't be moved.
2. Anything related to visibility of views i.e. ```HandleVisibility``` and ```SetVisibility``` can't be moved.
3. The code related to search view can't be moved (Most probably), so methods like ```onQueryTextChange``` can't be moved either.
4. The swipe to refresh layout code and methods can't be moved either.

Please point me in the right direction if i am wrong.

android/app/src/main/java/org/fossasia/openevent/fragments/LocationsFragment.java
```diff
@@ -188,7 +184,6 @@ public void onDestroyView() {
         Utils.unregisterIfUrlValid(this);
 
         // Remove listeners to fix memory leak
-        realmResults.removeAllChangeListeners();
```
iamareebjamal, line 191: Trigger removal of listeners in the viewmodel to avoid memory leak

android/build.gradle
```diff
@@ -21,6 +21,7 @@ allprojects {
         maven {
             url 'https://github.com/uPhyca/stetho-realm/raw/master/maven-repo'
         }
+        maven { url 'https://maven.google.com' }
```
iamareebjamal: This is already present as `google()` section

iamareebjamal (changes requested):
OK. Yeah it seems like it is the only movable code :)

dr0pdb:
@iamareebjamal 
1. Triggered removal of listeners in the ```onCleared``` of the view model
2. Removed redundant repository declaration. 

Please review.

iamareebjamal (approved):

PR #2037, opened 2017-12-22
fixes #2033 Improved the UI for SpeakerDetailsActivity

### LINKED ISSUE #2033: Improve the UI for the SpeakerDetailsActivity
**Actual Behaviour**

There is scope of improvement in the UI of SpeakerDetailsActivity

**Expected Behaviour**

The UI should be improved.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

screenshot_2017-12-21-17-56-21-77

Yes

NA

**Proposed changes**

1. The Speaker's bio & social media should each be in enclosed in a card view.
2. Header's indicating bio(About me) , contact(Contact me) and seminar's(Seminars) can be inserted before the respective card views.

dr0pdb:
Fixes #2033 

Changes: Added cardviews for Speaker's bio and social media.
Added title before fields like bio,social media and seminars.

Screenshots for the change: 

screenshot_2017-12-22-16-41-09-78


android/app/src/main/res/values/color.xml
```diff
@@ -22,4 +22,5 @@
     <color name="timestamp">#a0a3a7</color>
     <color name="link">#0a80d1</color>
     <color name="card_title">#ffffff</color>
+    <color name="card_title_background">#EF9A9A</color>
```
iamareebjamal: This is a non themable, non standard color. Please align the card textview inside the card itself and remove the bg color

iamareebjamal (changes requested):

dr0pdb:
@iamareebjamal 
Changes:
1. Removed the bg color
2. Alligned the card text view inside the card.

Screenshot:
screenshot_2017-12-22-17-50-27-88

Please review.

iamareebjamal:
I think we should let it be left aligned and not enclose `Sessions` cards inside another card. 

Also it is sessions rather than seminars

dr0pdb:
Changes: 
1. Title's are left alligned.
2. Sessions card is outside the recycler view cards.

Screenshot:
screenshot_2017-12-22-22-24-59-30

@iamareebjamal Please review.

iamareebjamal:
Sessions should not be in CardView, and please make title padding same as content

dr0pdb:
Changes: Fixed padding and sessions title no longer inside a card view.
Screenshot:
screenshot_2017-12-22-22-37-51-03

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/activities/SpeakerDetailsActivity.java
```diff
@@ -239,21 +247,34 @@ else if (!TextUtils.isEmpty(selectedSpeaker.getOrganisation()))
 
     private void hideEmptyURLButtons() {
 
+        int hidden = 0;
+
         if (TextUtils.isEmpty(selectedSpeaker.getLinkedin())) {
             linkedin.setVisibility(View.GONE);
+            hidden++;
```
the-dagger: Instead, please use a flag with a default value of false here which gets set to true if there is even a single visible URL.

the-dagger (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SpeakerDetailsActivity.java
dr0pdb: @the-dagger 
Using a flag would mean either:
1. I have to add else with each and every ```if``` to mark the flag as true.
2. I have to add an if statement in the end checking  if atleast one of the URL is visible and mark the flag as true in that case.

In my view using an int counter like i have done or going for the second option is good.

the-dagger: Set the flag to false by default and in each if block, set its value to true.
At the end, check if it is true to show or false to hide the card.

Makes sense?

dr0pdb: Sorry but it doesn't.
We are actually hiding things(not showing) in the ```if``` block. Let's say one block of ```if``` runs, then that means one of the social media info is **not** available.
If we need to hide the card, we need to be sure that all the five social media info's are not available i.e. all the five ```if``` blocks need to run. If any one of them doesn't run,then we are sure to have a social media block for the speaker.

the-dagger: Ok, I seem to have missed that part.
No worries, then make the visibility as GONE by default and modify the if blocks to set the visibility to VISIBLE in case that social media field exists.

Let's not hardcode the value to 5 as it's been done here :)

dr0pdb: Yeah that will do :)

the-dagger: 👍

srv-twry:
Changes: The social media links and the title card starts as "GONE" and are set to visible as suggested.
@the-dagger  Please review.

dr0pdb:
Changes: The social media links and the title card starts as "GONE" and are set to visible as suggested.
@the-dagger   Please review.

the-dagger (approved):

PR #2039, opened 2017-12-22
feat: Add viewmodel class for the SpeakersListFragment

### LINKED ISSUE #2038: Create ViewModel for the SpeakersListFragment
**Actual Behaviour**

ViewModel is absent for the SpeakersListFragment.

**Expected Behaviour**

It should be present

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes!

NA

**Other information**

Parent issue #2015

dr0pdb:
Fixes #2038 
Changes: Added ViewModel class for the SpeakersListFragment.
Migrated Database query and search state code to the ViewModel.
Very similar to #2029 

Works well in below conditions:
1. Tested by sorting according to different criterias(country,organisation).
2. Tested the functionality with search text view.

@iamareebjamal Please review.


android/app/src/main/java/org/fossasia/openevent/fragments/SpeakersListFragment.java
```diff
@@ -153,15 +146,14 @@ public void onDestroyView() {
         Utils.unregisterIfUrlValid(this);
 
         // Remove listeners to fix memory leak
-        realmResults.removeAllChangeListeners();
         if(swipeRefreshLayout != null) swipeRefreshLayout.setOnRefreshListener(null);
         if(searchView != null) searchView.setOnQueryTextListener(null);
     }
 
     @Override
     public void onSaveInstanceState(Bundle bundle) {
```
the-dagger: No need for this anymore, ViewModel will outlive the activity lifecycle so you'll always get the same data you fetched initially.

the-dagger (changes requested):

dr0pdb:
Changes: Removed redundant ```onSaveInstanceState()``` method.
@the-dagger Please review.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SpeakersListFragmentViewModel.java
```diff
@@ -0,0 +1,49 @@
...
+import io.realm.RealmResults;
+
+import static org.fossasia.openevent.utils.SortOrder.sortOrderSpeaker;
+
+public class SpeakersListFragmentViewModel extends ViewModel{
+
+    private MutableLiveData<List<Speaker>> speakersList;
+    private RealmDataRepository realmRepo;
+    private RealmResults<Speaker> realmResults;
+    private String searchText = "";
```
the-dagger, line 21: Do we need this variable?

dr0pdb: Yes, i think we need to store the value of the Search text here after the activity is no longer active.

the-dagger: But I can't see this variable being used anywhere while fetching the data inside ViewModel.
It only has a getter and setter.

dr0pdb: Sorry but i am getting slightly confused here. 
Here is the flow:
1. The activity gets created.(blank searchText variable is declared in the activity : line number 66)
2. The view model instance is created.(the blank searchText variable is created in the viewmodel : line number 21)
3. On line 85 in the activity we query the view model for the searchText.

Now let's say the screen is rotated, and the activity is recreated. Then again in the onCreate line number 85 , we will get the original search query back.

I can see the use of the variable here.

srv-twry: @the-dagger Infact i think we needed that ```onSavedInstanceState``` method also. Since there is no way of setting the ```searchText``` string in the ViewModel.

the-dagger: @srv-twry 
Looks like there is an issue with how the search behavior has been implemented here.
Instead of querying the data from the database for entered text, the existing data is being filtered.

>On line 85 in the activity we query the view model for the searchText.

Nope, take a look closely, the ViewModel always returns the entire list of speakers.

What you need to do here is, get rid of the query method in the adapter and implement the search the way it should be done.

You will need to create a new method inside your ViewModel that queries the database based on the string that was entered by the user and then returns the value back.

the-dagger: Or maybe open a new issue here for properly implementing the search behavior.
The current implementation looks a bit hacky to me.

I'll be approving this PR for now, but let's not leave this issue unattended.
Querying in the adapter is something that shouldn't really be done.
The data should be queried in the ViewModel and should be presented to the user from over there.

the-dagger (approved):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SpeakersListFragmentViewModel.java
dr0pdb: @the-dagger Yes.
The ViewModel always returns the entire list of speakers and then the actual filtering is done inside the adapter using the "filter" method
That's how it was implemented before and has been implemented everywhere else.  Hence i thought it should be left untouched.

the-dagger: No problem.
We can fix the implementation since we are anyways rewriting the Repository layer of the app.

dr0pdb: Infact i had a PR merged today which had the same issue. It was for LocationsFragment. There also the filter operation was done inside the adapter and hence i left it untouched.
Shall i open issues for each and every recyclerview because it's implemented in this way everywhere for eg in LocationsFragment, SpeakersListFragment, DayScheduleFragment, TracksFragment.
It's almost in each and every RecyclerView.

the-dagger: Open an issue for this PR and after that with every ViewModel related PR, you can do this modification in that PR itself.

PR #2048, opened 2017-12-23
Fix #1987-Give option to undo in Tracks section.

### LINKED ISSUE #1987: No option to undo Bookmark change in the Tracks section
**Actual Behaviour**

Currently the Snackbar in Tracks when a bookmark is added or removed does not have an UNDO action.

**Expected Behaviour**

There should be an UNDO action like the one in Bookmarks section.

**Steps to reproduce it**

Open Tracks
Select any option
Bookmark / Remove bookmark for any event.

**Screenshots of the issue**

image

image

Yes

mayank8318:
Fixes #1987 

Changes: Provide undo button in the snackbar.

Screenshots for the change: 
image

image

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
@@ -4,6 +4,8 @@ import android.app.SearchManager;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
+import android.support.design.widget.CoordinatorLayout;
+import android.support.design.widget.Snackbar;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
@@ -16,11 +18,14 @@ import android.widget.TextView;
 
 import org.fossasia.openevent.OpenEventApp;
 import org.fossasia.openevent.R;
+import org.fossasia.openevent.listeners.BookmarkStatus;
 import org.fossasia.openevent.adapters.GlobalSearchAdapter;
 import org.fossasia.openevent.data.Microlocation;
 import org.fossasia.openevent.data.Session;
 import org.fossasia.openevent.data.Speaker;
 import org.fossasia.openevent.data.Track;
+import org.fossasia.openevent.listeners.OnBookmarkSelectedListener;
+import org.fossasia.openevent.utils.SnackbarUtil;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -32,13 +37,12 @@ import io.realm.Realm;
 import io.realm.RealmResults;
 import timber.log.Timber;
 
-public class SearchActivity extends BaseActivity implements SearchView.OnQueryTextListener {
+public class SearchActivity extends BaseActivity implements SearchView.OnQueryTextListener, OnBookmarkSelectedListener {
 
     private GlobalSearchAdapter globalSearchAdapter;
     private List<Object> results = new ArrayList<>();
 
     private Realm realm = Realm.getDefaultInstance();
-
     private SearchView searchView;
     private String searchText;
 
@@ -48,6 +52,8 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
     protected RecyclerView searchRecyclerView;
     @BindView(R.id.txt_no_results)
     protected TextView noResultsView;
+    @BindView(R.id.main_content)
+    protected CoordinatorLayout coordinatorLayoutParent;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -60,6 +66,7 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
         handleVisibility();
 
         globalSearchAdapter = new GlobalSearchAdapter(results, this);
+        globalSearchAdapter.setOnBookmarkSelectedListener(this);
         searchRecyclerView.setAdapter(globalSearchAdapter);
         searchRecyclerView.addItemDecoration(new DividerItemDecoration(this, DividerItemDecoration.VERTICAL));
         searchRecyclerView.setLayoutManager(new LinearLayoutManager(this));
@@ -159,6 +166,7 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
 
         //Set listener to null to avoid memory leaks
         if (searchView != null) searchView.setOnQueryTextListener(null);
+        globalSearchAdapter.clearOnBookmarkSelectedListener();
     }
 
     public void addResultsFromTracks(String queryString) {
@@ -231,6 +239,13 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
             noResultsView.setVisibility(View.INVISIBLE);
             searchRecyclerView.setVisibility(View.VISIBLE);
         }
+    }
+
+    @Override
+    public void showSnackbar(BookmarkStatus bookmarkStatus) {
+        Snackbar snackbar = Snackbar.make(coordinatorLayoutParent, SnackbarUtil.getMessageResource(bookmarkStatus), Snackbar.LENGTH_LONG);
+        SnackbarUtil.setSnackbarAction(this, snackbar, bookmarkStatus)
+                .show();
     }
 }
 
--- a/android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
+++ b/android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
@@ -11,6 +11,7 @@ import org.fossasia.openevent.adapters.viewholders.DayScheduleViewHolder;
 import org.fossasia.openevent.adapters.viewholders.HeaderViewHolder;
 import org.fossasia.openevent.data.Session;
 import org.fossasia.openevent.dbutils.RealmDataRepository;
+import org.fossasia.openevent.listeners.OnBookmarkSelectedListener;
 import org.fossasia.openevent.utils.DateConverter;
 import org.fossasia.openevent.utils.SortOrder;
 import org.fossasia.openevent.utils.Utils;
@@ -31,6 +32,7 @@ public class DayScheduleAdapter extends BaseRVAdapter<Session, DayScheduleViewHo
 
     private Context context;
     private String eventDate;
+    private OnBookmarkSelectedListener onBookmarkSelectedListener;
 
     private RealmDataRepository realmRepo = RealmDataRepository.getDefaultInstance();
 
@@ -59,7 +61,7 @@ public class DayScheduleAdapter extends BaseRVAdapter<Session, DayScheduleViewHo
     public DayScheduleViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
         LayoutInflater layoutInflater = LayoutInflater.from(parent.getContext());
         View view = layoutInflater.inflate(R.layout.item_schedule, parent, false);
-        return new DayScheduleViewHolder(view,context);
+        return new DayScheduleViewHolder(view,context, onBookmarkSelectedListener);
     }
 
     @Override
@@ -136,4 +138,7 @@ public class DayScheduleAdapter extends BaseRVAdapter<Session, DayScheduleViewHo
         }
     }
 
+    public void setOnBookmarkSelectedListener(OnBookmarkSelectedListener onBookmarkSelectedListener) {
+        this.onBookmarkSelectedListener = onBookmarkSelectedListener;
+    }
 }
--- a/android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
+++ b/android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
@@ -18,6 +18,7 @@ import org.fossasia.openevent.data.Session;
 import org.fossasia.openevent.data.Speaker;
 import org.fossasia.openevent.data.Track;
 import org.fossasia.openevent.dbutils.RealmDataRepository;
+import org.fossasia.openevent.listeners.OnBookmarkSelectedListener;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -25,6 +26,7 @@ import java.util.List;
 public class GlobalSearchAdapter extends BaseRVAdapter<Object, RecyclerView.ViewHolder> {
 
     private Context context;
+    private OnBookmarkSelectedListener onBookmarkSelectedListener;
 
     private List<Object> filteredResultList = new ArrayList<>();
     private RealmDataRepository realmRepo = RealmDataRepository.getDefaultInstance();
@@ -109,7 +111,7 @@ public class GlobalSearchAdapter extends BaseRVAdapter<Object, RecyclerView.View
                 break;
             case SESSION:
                 View bookmark = inflater.inflate(R.layout.item_schedule, parent, false);
-                resultHolder = new DayScheduleViewHolder(bookmark, context);
+                resultHolder = new DayScheduleViewHolder(bookmark, context, onBookmarkSelectedListener);
                 break;
             default:
                 //If viewType doesn't match any of the above objects no view is created
@@ -153,4 +155,11 @@ public class GlobalSearchAdapter extends BaseRVAdapter<Object, RecyclerView.View
         }
     }
 
+    public void setOnBookmarkSelectedListener(OnBookmarkSelectedListener onBookmarkSelectedListener) {
+        this.onBookmarkSelectedListener = onBookmarkSelectedListener;
+    }
+
+    public void clearOnBookmarkSelectedListener() {
+        this.onBookmarkSelectedListener = null;
+    }
 }
--- a/android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
+++ b/android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
@@ -5,8 +5,6 @@ import android.content.Intent;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.os.Build;
-import android.support.design.widget.Snackbar;
-import android.support.v4.graphics.ColorUtils;
 import android.support.v7.widget.RecyclerView;
 import android.view.View;
 import android.widget.Button;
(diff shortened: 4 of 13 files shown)


dr0pdb (approved):

the-dagger (dismissed):
Let's implement the snacker properly here.
Ideally it shouldn't be obstructing the FAB as shown in the screenshot, instead use CoordinatorLayout as parent and pass it to your snacker's view.

Ref : https://stackoverflow.com/a/32777587/5471095

mayank8318:
@the-dagger Actually since the snackbar is used in the view holder, we cannot access the Coordinator Layout. @nikit19 is trying to solve it in PR #1992 .

iamareebjamal:
@mayank8318  Status? We have a lot of PRs here which are almost complete and unmerged. Let's help clean the backlog up.

The ViewHolder can let the adapter know when to show the snackbar and the adapter can tell the fragment or activity the same through callback interface

We can have discussions on implementation here if there's a problem

mayank8318:
@iamareebjamal Sorry for the delay I was waiting for the other PR to solve it since it is on this particular issue.
Need help here - 
I made an interface for callback in the `DayScheduleAdapter` but here the problem is I have to pass the callback between fragments. So I made the interface extend `Serializable` and passed the callback as an argument in a bundle. But I get null value for the object. What am I doing wrong here?

public static DayScheduleFragment getInstance(DayScheduleAdapter.DayScheduleCallback dayScheduleCallback)
    {
        DayScheduleFragment f = new DayScheduleFragment();
        Bundle args = new Bundle();
        args.putSerializable("callback", dayScheduleCallback);
        f.setArguments(args);
        return f;
    }

    @Override
    public void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        date = getArguments().getString(ConstantStrings.EVENT_DAY, "");
        dayScheduleCallback = (DayScheduleAdapter.DayScheduleCallback) getArguments().getSerializable("callback");
    }

iamareebjamal:
You don't need to send it in Bundle. Actually there's no pattern where an object to be used as a listener should be serialized and sent. Only data objects should be serialized. Secondly, when you serialize the object and deserialize it, the object reference is not the same, a new object is created, so the callback won't even work.

Also, if the interface implementer is the whole activity, then you are serializing the whole activity and then sending which is a huge memory leak

You can simply add a setter in fragment and pass the interface implementation there

mayank8318:
Changes : Adjust for FAB so that Snackbar doesn't overlap with it.
image

android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
```diff
@@ -29,18 +30,28 @@
...
     private String eventDate;
+    private DayScheduleCallback dayScheduleCallback;
 
     private RealmDataRepository realmRepo = RealmDataRepository.getDefaultInstance();
 
     private ArrayList<String> tracks = new ArrayList<>();
     private List<Session> copyOfSessions = new ArrayList<>();
 
-    public DayScheduleAdapter(List<Session> sessions, Context context) {
+    public interface DayScheduleCallback {
```
iamareebjamal: Please rename to a proper name, like OnBookmarkSelectedListener

iamareebjamal: Also this is used in places other than DayScheduleAdapter, so move to its own class

android/app/src/main/java/org/fossasia/openevent/adapters/DayScheduleAdapter.java
```diff
@@ -29,18 +30,28 @@
...
+        void showSnackbar(int toDisplayResource, int length, int actionCode, int sessionId, ImageButton slot_bookmark, int colorCode);
+    }
+
+    public DayScheduleAdapter(List<Session> sessions, Context context, DayScheduleCallback dayScheduleCallback) {
         super(sessions);
         this.copyOfSessions = new ArrayList<>(sessions);
         this.context = context;
+        this.dayScheduleCallback = dayScheduleCallback;
```
iamareebjamal: Remove from constructor and add a setter instead

android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
```diff
@@ -36,10 +37,11 @@
     private final int DIVIDER = 4;
     private final int SESSION = 5;
 
-    public GlobalSearchAdapter(List<Object> dataList, Context context) {
+    public GlobalSearchAdapter(List<Object> dataList, Context context, DayScheduleAdapter.DayScheduleCallback dayScheduleCallback) {
         super(dataList);
         this.context = context;
         this.filteredResultList = dataList;
+        this.dayScheduleCallback = dayScheduleCallback;
```
iamareebjamal: Same here

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
```diff
@@ -89,50 +94,34 @@ public void bindSession(RealmDataRepository realmRepo){
...
-                                    slot_bookmark.setImageResource(R.drawable.ic_bookmark_white_24dp);
-
-                                    WidgetUpdater.updateWidget(context);
-                                })
-                                .setActionTextColor(ColorUtils.setAlphaComponent(Views.getAccentColor(context), 220));
-                        snackbar.show();
-                    } else {
-                        Snackbar.make(slot_content, R.string.removed_bookmark, Snackbar.LENGTH_SHORT).show();
-                    }
+                    dayScheduleCallback.showSnackbar(R.string.removed_bookmark, Snackbar.LENGTH_LONG, CODE_UNDO_REMOVED, sessionId, slot_bookmark, storedColor);
```
iamareebjamal: Check for null before calling the callback

iamareebjamal: Please reduce the arguments amount to at most 2. We don't need to pass the snack bar length and string resource to be used. It should just have id and information that it was removed or added, which can be expressed through a boolean

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
```diff
@@ -89,50 +94,34 @@ public void bindSession(RealmDataRepository realmRepo){
...
+            slot_bookmark.setColorFilter(storedColor, PorterDuff.Mode.SRC_ATOP);
 
             final int sessionId = session.getId();
 
             slot_bookmark.setOnClickListener(v -> {
-                if(session.getIsBookmarked()) {
+                if (session.getIsBookmarked()) {
 
                     realmRepo.setBookmark(sessionId, false).subscribe();
                     slot_bookmark.setImageResource(R.drawable.ic_bookmark_border_white_24dp);
```
iamareebjamal: Also, please change this to camel case while you're at it

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -235,5 +250,30 @@ public void onDestroyView() {
...
+                slot_bookmark.setImageResource(R.drawable.ic_bookmark_white_24dp);
+                slot_bookmark.setColorFilter(colorCode, PorterDuff.Mode.SRC_ATOP);
+
+                WidgetUpdater.updateWidget(getContext());
+            })
+                    .setActionTextColor(ColorUtils.setAlphaComponent(Views.getAccentColor(getContext()), 220));
+        }
+
+        snackbar.show();
+    }
```
iamareebjamal: This code is almost same everywhere, shouldn't this be handled in just one place, the activity?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
mayank8318: @iamareebjamal  Sorry didn't get you. Do you mean like a helper function for making the Snackbar?

iamareebjamal: I mean if there is no change, and it only occurs in a single fragment hierarchy, why can't it be handled in a single place

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
mayank8318: Can't find that method and if I refactor it here it treats it as a variable "isBookmarked" which is part of Session Realm Object

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
mayank8318: @iamareebjamal We will still need the reference to slot_bookmark right? Should I pass the view holder itself as then I dont need to pass color, session etc

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
mayank8318: @iamareebjamal We need to pass the parentView in each fragment while making the Snackbar. So I can make a helper method where I pass the view from the fragments and the rest of the code is the same everywhere. Didnt know if this was good practice or not. Should I do it like this?
It is done similarly in Phimpe if I am not wrong.

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
iamareebjamal: Create an Object with these parameters and send it instead. Sounds good?

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
iamareebjamal: Yup, if the code depends on each fragment's layout, then create a helper method for all the similar code

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
mayank8318: Yep cool

mayank8318:

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
iamareebjamal: I'm talking about `slot_bookmark` variable

mayank8318: Done

android/app/src/main/java/org/fossasia/openevent/SnackbarHelper.java
```diff
@@ -0,0 +1,31 @@
+package org.fossasia.openevent;
+
+import android.widget.ImageButton;
+
+/**
+ * Created by mayank on 14/1/18.
+ */
+public class SnackbarHelper {
```
iamareebjamal: Please remove author comments and move to appropriate package

android/app/src/main/java/org/fossasia/openevent/SnackbarHelper.java
```diff
@@ -0,0 +1,31 @@
...
+        return imageButton;
+    }
+
+    public int getStoredColor() {
+        return storedColor;
+    }
+
+    public int getSessionId() {
+        return sessionId;
+    }
```
iamareebjamal: Instead of generated getters, use Lombok `@Data` annotation

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -232,5 +238,10 @@ public void handleVisibility() {
             searchRecyclerView.setVisibility(View.VISIBLE);
         }
     }
+
+    @Override
+    public void showSnackbar(int actionCode, SnackbarHelper snackbarHelper) {
```
iamareebjamal: Encapsulate `actionCode` inside object itself. Change int to enum, rename SnackbarHelper to `BookmarkStatus`

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/DayScheduleViewHolder.java
```diff
@@ -173,10 +169,13 @@ public void bindSession(RealmDataRepository realmRepo){
             Timber.d("This session has no track somehow : " + session + " " + sessionTrack);
         }
 
-        if(session.getMicrolocation() != null) {
+        if (session.getMicrolocation() != null)
+
+        {
```
iamareebjamal: Please correct the syntax of all above if and elses

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -134,6 +140,7 @@ public void onEventLoaded(EventLoadedEvent eventLoadedEvent) {
     private void setUpBookmarksRecyclerView() {
         bookmarksRecyclerView.setVisibility(View.VISIBLE);
         bookMarksListAdapter = new GlobalSearchAdapter(sessions, getContext());
+        bookMarksListAdapter.setOnBookmarkSelectedListener(this);
```
iamareebjamal, line 143: Remove the listeners on destroy in each fragment/activity

android/app/src/main/java/org/fossasia/openevent/listeners/OnBookmarkSelectedListener.java
```diff
@@ -0,0 +1,16 @@
+package org.fossasia.openevent.listeners;
+
+import org.fossasia.openevent.SnackbarHelper;
+
+/**
+ * Created by mayank on 13/1/18.
+ */
+
+public interface OnBookmarkSelectedListener {
```
iamareebjamal: Remove author comment

android/app/src/main/java/org/fossasia/openevent/utils/SnackbarUtil.java
```diff
@@ -0,0 +1,52 @@
...
+        else if (actionCode == CODE_ERROR)
+            toDisplayResource = R.string.error_create_notification;
+        else
+            toDisplayResource = R.string.added_bookmark;
+
+        Snackbar snackbar = Snackbar.make(view, toDisplayResource, Snackbar.LENGTH_LONG);
+        if (actionCode == CODE_UNDO_ADDED) {
+            snackbar = snackbar.setAction(R.string.undo, view1 -> {
+                RealmDataRepository.getDefaultInstance().setBookmark(snackbarHelper.getSessionId(), false).subscribe();
+                snackbarHelper.getImageButton().setImageResource(R.drawable.ic_bookmark_white_24dp);
```
iamareebjamal: We only need the snackbar stuff to be handled in this utility class, passing foreign views is bad. Handle what can be handled locally, there itself, create a common static method in `Views` if you need to, but only pass absolutely needed stuff to the listener. I think the only thing needed to pass is the action type and the id of the session.

iamareebjamal (changes requested):

mayank8318:
Can't find a solution for the Codacy issue. Any suggestions there?

android/app/src/main/java/org/fossasia/openevent/status/BookmarkStatus.java
```diff
@@ -0,0 +1,18 @@
+package org.fossasia.openevent.status;
```
iamareebjamal: A new package is not needed, just add it in the same package as its listener. The coupling between components should be reflected by the packaging

android/app/src/main/java/org/fossasia/openevent/status/BookmarkStatus.java
```diff
@@ -0,0 +1,18 @@
...
+import org.fossasia.openevent.listeners.OnBookmarkSelectedListener;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class BookmarkStatus {
+
+    private ImageButton imageButton;
```
iamareebjamal: Why can't we handle the image button locally, why do we need to send it in a listener? We should only propagate data, not views.

android/app/src/main/java/org/fossasia/openevent/status/BookmarkStatus.java
```diff
@@ -0,0 +1,18 @@
...
+    private int storedColor;
+    private int sessionId;
+    private OnBookmarkSelectedListener.Status actionCode;
```
iamareebjamal: Move the status out of listener, you may add it in this class instead. Listener should be pure interface

android/app/src/main/java/org/fossasia/openevent/utils/SnackbarUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.fossasia.openevent.listeners.OnBookmarkSelectedListener.Status.CODE_ERROR;
+import static org.fossasia.openevent.listeners.OnBookmarkSelectedListener.Status.CODE_UNDO_ADDED;
+import static org.fossasia.openevent.listeners.OnBookmarkSelectedListener.Status.CODE_UNDO_REMOVED;
+
+public class SnackbarUtil {
+    public static Snackbar setSnackbarAction(Context context, Snackbar snackbar, BookmarkStatus bookmarkStatus) {
+
+        if (bookmarkStatus.getActionCode() == CODE_UNDO_ADDED) {
+            snackbar = snackbar.setAction(R.string.undo, view1 -> {
```
iamareebjamal: You don't need to reassign `snackbar`. It is not an immutable type like String

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/status/BookmarkStatus.java
mayank8318: `slotBookmark` belongs to the viewholder, so I cannot access it outside without passing it. Or I will have to access the view holder somehow.

iamareebjamal: Why do you need to access it in the listener implementation?

mayank8318: I need to update the UI in the snackbar action for undo.

iamareebjamal: When you undo, the view is updated again, right? So it can be updated there

iamareebjamal (approved):
Great work :smile:

PR #2049, opened 2017-12-23
Added a cancel button in the network not available dialog

### LINKED ISSUE #2041: Add a cancel button in the Network not available dialog
**Actual Behaviour**

When opening the application without internet connection, a dialog appears with title "Network not available" with one button "Ok" which directs the user to device Settings. There is no option to cancel and browse the application without network.

**Expected Behaviour**

There should be a "Cancel" button.

**Steps to reproduce it**

Switch off mobile data and wifi.
Open the application.

**LogCat for the issue**

NA

**Screenshots of the issue**

screenshot_2017-12-22-23-14-16-22

Yes

NA

**Proposed changes**

The dialog should also contain a "Cancel" button to close the dialog. Since it is possible to browse the application without network, the user should be allowed to do so. Infact since the dialog is ```cancellable```, it would be much better if user has an option right in front of him/her. 

dr0pdb:
Fixes #2041 
Changes: Created a dialog function in the DialogFactory and used it to create a new dialog with ```OK``` and ```Cancel``` buttons.

Screenshots for the change: 
screenshot_2017-12-23-15-05-39-05

@the-dagger Please review.


android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -650,12 +650,22 @@ public void noInternet(NoInternetEvent event) {
...
-                        startActivity(setNetworkIntent);
+                    R.string.ok,
+                    R.string.cancel,
+                    (dialogInterface, button) -> {
+                        switch (button) {
+                            case DialogInterface.BUTTON_POSITIVE:
+                                Intent setNetworkIntent = new Intent(Settings.ACTION_SETTINGS);
+                                startActivity(setNetworkIntent);
+                                break;
+                            case DialogInterface.BUTTON_NEGATIVE:
```
the-dagger, line 664: Maybe dismiss the dialog here?

the-dagger (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
dr0pdb: @the-dagger Done with the change.

the-dagger (approved):

Issue #2042, opened 2017-12-22
Implement search query in the ViewModel for SpeakersListFragment


dr0pdb:
**Actual Behaviour**

Currently the search query in most of the fragments is implemented inside the adapter.

**Expected Behaviour**

The search query should be implemented inside the ViewModel for SpeakersListFragment

**Steps to reproduce it**

NA.

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes, but after i am done with the rest of ```ViewModel```  PR's. Can be assigned to someone else in the mean time.

**Target Milestone to solve the issue**

NA

**Other information**

Please check the discussion out in this PR #2039

iamareebjamal:
But this will mean unnecessary fetching from database for filtering data which is already present in memory @the-dagger

This will both be less performant and more resource intensive and slow. Filter method in adapter is explicitly used for this purpose 

Any advantage of this?

the-dagger:
@iamareebjamal I am really not sure if using a list and filtering on it is a good idea.
I mean we will be going ahead with ROOM and later on a Paged List anyways, combine that with large number of speakers in the app (~300), not sure if going ahead, filtering on the list would be good.

What do you think?

dr0pdb:
@the-dagger May i add something here ?

the-dagger:
Sure, definitely 👍

dr0pdb:
I think we can implement the query inside the ViewModel without querying the database more than once.
The complete list will be fetched from the database and will be stored inside the ```ViewModel``` and we can store another list that will store only the filtered values according to the search query and return it when required.

iamareebjamal:
@the-dagger Any interaction with database is always slower than operation in memory. Querying with LIKE statement is even slower and we are not implementing throttling, so on each keypress of user, 300 items are being searched again, and then modelled into POJOs, new objects are being created, list is being cleared and add items are inserted. This will also cause frequent GC calls to collect old redundant objects. Furthermore, with database filtering we cannot implement fuzzy search in future.

the-dagger:
@iamareebjamal Ok, alright.
But let's atleast move the logic for filtering away from the Adapter, instead put the filter in the ViewModel instead and keep the adapter here free from any such logic.

Sounds good?

iamareebjamal:
Yeah that would be great

PR #2055, opened 2017-12-23
Fix #2045 create ViewModel for tracksFragment

### LINKED ISSUE #2045: Create ViewModel for the TracksFragment
**Actual Behaviour**

There is no ViewModel for the TracksFragment

**Expected Behaviour**

There should be one

Yes

nikit19:
Fixes #2045 

Changes: Added viewModel for tracksFragment 

Screenshots for the change: 
screenshot_20171223-231839

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.fragments;
 
+import android.arch.lifecycle.ViewModelProviders;
 import android.os.Bundle;
 import android.support.design.widget.Snackbar;
 import android.support.v4.view.MenuItemCompat;
@@ -24,12 +25,12 @@ import org.fossasia.openevent.R;
 import org.fossasia.openevent.adapters.TracksListAdapter;
 import org.fossasia.openevent.api.DataDownloadManager;
 import org.fossasia.openevent.data.Track;
-import org.fossasia.openevent.dbutils.RealmDataRepository;
 import org.fossasia.openevent.events.RefreshUiEvent;
 import org.fossasia.openevent.events.TracksDownloadEvent;
 import org.fossasia.openevent.utils.NetworkUtils;
 import org.fossasia.openevent.utils.Utils;
 import org.fossasia.openevent.utils.Views;
+import org.fossasia.openevent.viewmodels.TracksFragmentViewModel;
 import org.fossasia.openevent.views.stickyheadersrecyclerview.StickyRecyclerHeadersDecoration;
 
 import java.lang.ref.WeakReference;
@@ -37,7 +38,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import butterknife.BindView;
-import io.realm.RealmResults;
 import timber.log.Timber;
 
 /**
@@ -45,8 +45,6 @@ import timber.log.Timber;
  * Date: 05-06-2015
  */
 public class TracksFragment extends BaseFragment implements SearchView.OnQueryTextListener {
-
-    final private String SEARCH = "searchText";
 
     private List<Track> tracks = new ArrayList<>();
     private TracksListAdapter tracksListAdapter;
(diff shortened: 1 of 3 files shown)


dr0pdb:
@nikit19 Please wait before a decision is taken on whether to implement the Search query filtering inside the adapter or inside the ViewModel. Please check these discussions out #2042 #2039 .
In case it's decided to implement it inside the ViewModel then you will have to implement it also. So it's better to wait in my opinion.

nikit19:
@srv-twry okay cool.

android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
```diff
@@ -132,16 +129,14 @@ public void onDestroyView() {
...
         if(swipeRefreshLayout != null) swipeRefreshLayout.setOnRefreshListener(null);
         if(searchView != null) searchView.setOnQueryTextListener(null);
     }
 
     @Override
     public void onSaveInstanceState(Bundle bundle) {
-        if (isAdded() && searchView != null) {
-            bundle.putString(SEARCH, searchText);
+        if (isAdded() && searchView != null && tracksFragmentViewModel != null) {
+        tracksFragmentViewModel.setSearchText(searchText);
```
dr0pdb: Please indent it properly.

dr0pdb (changes requested):
I downloaded your PR locally to test it. It seems like the Search functionality isn't working as expected.

**Actual behaviour**
Once you enter a search query the view is updated but when you delete that query the View isn't updated.

**Expected behaviour**
The View should be updated in both the cases.

**Steps to reproduce**

1. Enter a random text inside the search query so that all the tracks are filtered out.
2. Delete the search query now, the tracks should come back again but they don't.

**Screenshots**
screenshot_2017-12-25-16-02-50-94

screenshot_2017-12-25-16-02-57-18

Please fix it.

nikit19:
Hi @srv-twry thanks for informing. You are right. The search isn't working like it should work.
But any idea why is it not working properly because I didn't change any of the onQuery methods ?

android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
nikit19: :+1:

dr0pdb:
@nikit19 In a second thought , i feel it's not advantageous to actually find out the bug and fix it. You have to change the Search Query implementation anyways.
Now that the discussion at #2042 is finished, you can go ahead and move the filter logic inside the ViewModel only. If the issue about the Seach Query still persists after that, then we will try to find the bug :)

nikit19:
@srv-twry filter logic added and now the search functionality works as expected
ezgif com-video-to-gif 28

nikit19:
@iamareebjamal Please review

iamareebjamal:
@nikit19 Please update the PR to use RealmDataWrapper

nikit19:
@iamareebjamal I have updated the viewmodel code but I am always getting "no tracks yet "
I suspect it has something to with the line 52 in the tracksFragmentViewModel.java
as tracksList.getValue() is always null.
Can you please guide me what went wrong

nikit19:
@iamareebjamal updated the PR to use FilterableLiveRealmData
Now the tracksFragment is working as expected.
Please review

android/app/src/main/java/org/fossasia/openevent/viewmodels/TracksFragmentViewModel.java
```diff
@@ -0,0 +1,50 @@
...
+        if (!this.searchText.equals(searchText) || tracksList == null) {
+            setSearchText(searchText);
+            final String query = searchText.toLowerCase(Locale.getDefault());
+            Predicate<Track> predicate = track -> track.getName()
+                    .toLowerCase(Locale.getDefault())
+                    .contains(query);
+            filterableRealmLiveData.filter(predicate);
+            if (tracksList == null) {
+                tracksList = Transformations.map(filterableRealmLiveData, input -> input);
+            }
```
iamareebjamal: FilterableRealmLiveData is already LiveData of List right, why do we need to transform again?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/TracksFragmentViewModel.java
nikit19: Yes, it was redundant.Removed it, please review

iamareebjamal (approved):

PR #2058, opened 2017-12-25
Fix #2047 - Create View model for FeedFragment.

### LINKED ISSUE #2047: Create ViewModel for Feed Fragment
Create ViewModel for Feeds fragment.

Yes

mayank8318:
Fixes #2047 

Changes: Create View Model for Feed Fragment.

Screenshots for the change: -

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
@@ -1,6 +1,7 @@
 package org.fossasia.openevent.fragments;
 
 import android.app.ProgressDialog;
+import android.arch.lifecycle.ViewModelProviders;
 import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.design.widget.Snackbar;
@@ -14,24 +15,23 @@ import android.widget.TextView;
 
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.adapters.FeedAdapter;
-import org.fossasia.openevent.api.APIClient;
 import org.fossasia.openevent.data.facebook.FeedItem;
 import org.fossasia.openevent.modules.OnImageZoomListener;
 import org.fossasia.openevent.utils.ConstantStrings;
 import org.fossasia.openevent.utils.NetworkUtils;
 import org.fossasia.openevent.utils.SharedPreferencesUtil;
 import org.fossasia.openevent.utils.Views;
+import org.fossasia.openevent.viewmodels.FeedFragmentViewModel;
 
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.List;
 
 import butterknife.BindView;
-import io.reactivex.android.schedulers.AndroidSchedulers;
-import io.reactivex.disposables.CompositeDisposable;
-import io.reactivex.disposables.Disposable;
-import io.reactivex.schedulers.Schedulers;
 import timber.log.Timber;
+
+import static org.fossasia.openevent.utils.AuthUtil.INVALID;
+import static org.fossasia.openevent.utils.AuthUtil.VALID;
 
 /**
  * Created by rohanagarwal94 on 10/6/17.
@@ -39,13 +39,24 @@ import timber.log.Timber;
 public class FeedFragment extends BaseFragment {
 
     private FeedAdapter feedAdapter;
-    private List<FeedItem> feedItems;
     private ProgressDialog downloadProgressDialog;
-    private final CompositeDisposable compositeDisposable = new CompositeDisposable();
+    private FeedFragmentViewModel feedFragmentViewModel;
+    private List<FeedItem> feedItems;
+    private FeedAdapter.OpenCommentsDialogListener openCommentsDialogListener;
+    private OnImageZoomListener onImageZoomListener;
 
     @BindView(R.id.feed_swipe_refresh) SwipeRefreshLayout swipeRefreshLayout;
     @BindView(R.id.feed_recycler_view) RecyclerView feedRecyclerView;
     @BindView(R.id.txt_no_posts) TextView noFeedView;
+
+    public static FeedFragment getInstance(FeedAdapter.OpenCommentsDialogListener openCommentsDialogListener,
+                                           OnImageZoomListener onImageZoomListener) {
+        FeedFragment feedFragment = new FeedFragment();
+        feedFragment.openCommentsDialogListener = openCommentsDialogListener;
+        feedFragment.onImageZoomListener = onImageZoomListener;
+
+        return feedFragment;
+    }
 
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
@@ -54,15 +65,16 @@ public class FeedFragment extends BaseFragment {
         View view = super.onCreateView(inflater, container, savedInstanceState);
 
         feedItems = new ArrayList<>();
+        feedFragmentViewModel = ViewModelProviders.of(this).get(FeedFragmentViewModel.class);
         RecyclerView.LayoutManager mLayoutManager = new LinearLayoutManager(getContext());
         feedRecyclerView.setLayoutManager(mLayoutManager);
-        feedAdapter = new FeedAdapter(getContext(), (FeedAdapter.OpenCommentsDialogListener)getActivity(), feedItems);
-        feedAdapter.setOnImageZoomListener((OnImageZoomListener)getActivity());
+        feedAdapter = new FeedAdapter(getContext(), openCommentsDialogListener, feedItems);
+        feedAdapter.setOnImageZoomListener(onImageZoomListener);
         feedRecyclerView.setAdapter(feedAdapter);
 
         setupProgressBar();
 
-        if(NetworkUtils.haveNetworkConnection(getContext()))
+        if (NetworkUtils.haveNetworkConnection(getContext()))
             showProgressBar(true);
 
         downloadFeed();
@@ -79,29 +91,26 @@ public class FeedFragment extends BaseFragment {
             return;
         }
 
-        compositeDisposable.add(APIClient.getFacebookGraphAPI()
-                .getPosts(SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_ID, null),
-                        getContext().getResources().getString(R.string.fields),
-                        getContext().getResources().getString(R.string.facebook_access_token))
-                .subscribeOn(Schedulers.io())
-                .observeOn(AndroidSchedulers.mainThread())
-                .subscribe(feed -> {
-                    feedItems.clear();
-                    feedItems.addAll(feed.getData());
-                    feedAdapter.notifyDataSetChanged();
-                    handleVisibility();
-                }, throwable -> {
-                    Snackbar.make(swipeRefreshLayout, getActivity()
-                            .getString(R.string.refresh_failed), Snackbar.LENGTH_LONG)
-                            .setAction(R.string.retry_download, view -> refresh()).show();
-                    Timber.d("Refresh not done");
-                    showProgressBar(false);
-                    swipeRefreshLayout.setRefreshing(false);
-                }, () -> {
-                    Views.setSwipeRefreshLayout(swipeRefreshLayout, false);
-                    Timber.d("Refresh done");
-                    showProgressBar(false);
-                }));
+        feedFragmentViewModel.getPosts(getContext().getResources().getString(R.string.fields),
+                getContext().getResources().getString(R.string.facebook_access_token), SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_ID, null))
+                .observe(this, feedResponse -> {
+                    if (feedResponse.getResponse() == VALID) {
+                        feedItems.clear();
+                        feedItems.addAll(feedResponse.getFeed().getData());
+                        feedAdapter.notifyDataSetChanged();
+                        handleVisibility();
+                        Views.setSwipeRefreshLayout(swipeRefreshLayout, false);
+                        Timber.d("Refresh done");
+                        showProgressBar(false);
+                    } else if (feedResponse.getResponse() == INVALID) {
+                        Snackbar.make(swipeRefreshLayout, getActivity()
+                                .getString(R.string.refresh_failed), Snackbar.LENGTH_LONG)
+                                .setAction(R.string.retry_download, view -> refresh()).show();
+                        Timber.d("Refresh not done");
+                        showProgressBar(false);
+                        swipeRefreshLayout.setRefreshing(false);
+                    }
+                });
     }
 
     public void handleVisibility() {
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,51 @@
...
+                    feedLive.setValue(feed.getData());
+                }, throwable -> {
+                    feedFragment.showSnackbarRefreshFailed();
+                }, () -> {
+                    feedFragment.refreshDone();
+                });
+        return feedLive;
+    }
+
+    @Override
```
dr0pdb: This method is redundant because you are overriding just to call super. Isn't it ?

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Created by mayank on 25/12/17.
+ */
+
+public class FeedFragmentViewModel extends ViewModel {
+
+    private MutableLiveData<List<FeedItem>> feedLive;
+
+    public LiveData<List<FeedItem>> getFeed(Context context, FeedFragment feedFragment) {
```
dr0pdb: You cannot hold an instance of any ```View```, ```Fragment``` or ```ActivityContext``` inside the ```ViewModel``` since it would cause memory leaks.

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,51 @@
...
+import org.fossasia.openevent.utils.ConstantStrings;
+import org.fossasia.openevent.utils.SharedPreferencesUtil;
+
+import java.util.List;
+
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.schedulers.Schedulers;
+
+/**
+ * Created by mayank on 25/12/17.
```
dr0pdb: Please remove this template comment.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,35 @@
...
+public class FeedFragmentViewModel extends ViewModel {
+
+    private MutableLiveData<List<FeedItem>> feedLive;
+
+    public FeedFragmentViewModel() {
+        feedLive = new MutableLiveData<>();
+    }
+
+    public LiveData<List<FeedItem>> getFeedList() {
+        if (feedLive == null)
```
dr0pdb: Isn't this null check redundant ? Since we're already initialising it in the constructor, how can it be null after that ?

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/adapters/BaseRVAdapter.java
```diff
@@ -59,9 +59,12 @@ public void clear() {
         notifyDataSetChanged();
     }
 
+    /**
+     * Return size of dataList and 0 if it is null.
+     */
     @Override
     public int getItemCount() {
-        return dataList.size();
+        return dataList == null ? 0 : dataList.size();
```
iamareebjamal: This will introduce subtle bugs, the app should crash if null list is passed

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -80,8 +79,8 @@ private void downloadFeed() {
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe(feed -> {
-                    feedItems.clear();
-                    feedItems.addAll(feed.getData());
+                    feedFragmentViewModel.getFeedList().getValue().clear();
```
iamareebjamal: You shouldn't mutate the ViewModel state, the UI should replicate the data

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,32 @@
...
+import android.arch.lifecycle.LiveData;
+import android.arch.lifecycle.MutableLiveData;
+import android.arch.lifecycle.ViewModel;
+
+import org.fossasia.openevent.data.facebook.FeedItem;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FeedFragmentViewModel extends ViewModel {
```
iamareebjamal: This class is redundant. it is doing nothing of logic except holding the LiveData. Please move all possible logic from Fragment to here

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/adapters/BaseRVAdapter.java
mayank8318: Oh ok i will revert that.

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
mayank8318: Ok I will make methods in ViewModel to do this and call it here?

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
mayank8318: I tried moving the API code but it requires `Context` and it does UI changes in the lambda expression. So I will have to pass context and reference of FeedFragment as parameters which is against the ViewModel principles. Any advice as to what I can do here?

iamareebjamal:
@mayank8318 Status? We have a lot of PRs here which are almost complete and unmerged. Let's help clean the backlog up.

There are many things that can be moved to ViewModel without context. In fact, any code that doesn't require android functionality like UI updating or checking network status does not need to be in Fragment at all:

```
APIClient.getFacebookGraphAPI().getPageId(SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_NAME, null),
                            getResources().getString(R.string.facebook_access_token))
                            .subscribeOn(Schedulers.io())
                            .observeOn(AndroidSchedulers.mainThread())
                            .subscribe(facebookPageId -> {
                                String id = facebookPageId.getId();
                                SharedPreferencesUtil.putString(ConstantStrings.FACEBOOK_PAGE_ID, id);
                            });
```

Can be moved to ViewModel and requires 2 strings to be passed in the method

mayank8318:
Changes : Move more code to view model.

android/app/src/main/java/org/fossasia/openevent/adapters/BaseRVAdapter.java
```diff
@@ -59,6 +59,9 @@ public void clear() {
         notifyDataSetChanged();
     }
 
+    /**
+     * Return size of dataList.
+     */
```
iamareebjamal: Commenting is good, but explaining pretty obvious code is anti-pattern. Please remove :smile:

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.schedulers.Schedulers;
+
+public class FeedFragmentViewModel extends ViewModel {
+
+    private List<FeedItem> feedLive;
```
iamareebjamal: This list and all the operations of add and clear should remain in the Fragment itself

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void clearFeedList() {
+        feedLive.clear();
+    }
+
+    public void addAllFeed(List<FeedItem> feed) {
+        feedLive.addAll(feed);
+    }
+
+    public void updateFBPageID(Resources resources) {
```
iamareebjamal: No passing of Android specific classes, please. Calculate SharedPref in fragment and pass the result here

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+        if (SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_ID, null) == null)
+            APIClient.getFacebookGraphAPI().getPageId(SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_NAME, null),
+                    resources.getString(R.string.facebook_access_token))
```
iamareebjamal: Here as well

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(facebookPageId -> {
+                        String id = facebookPageId.getId();
+                        SharedPreferencesUtil.putString(ConstantStrings.FACEBOOK_PAGE_ID, id);
+                    });
```
iamareebjamal: Convert this Observable to LiveData and return it to fragment and allow Fragment to handle its result

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    protected void onCleared() {
+        super.onCleared();
+        feedLive = null;
+    }
```
iamareebjamal: No point in setting list to null

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -85,8 +84,8 @@ private void downloadFeed() {
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe(feed -> {
```
iamareebjamal: This observable code should be in View Model

iamareebjamal (changes requested):

mayank8318:
@iamareebjamal Please review:)

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+public class FeedFragmentViewModel extends ViewModel {
+
+    private MutableLiveData<FacebookPageId> facebookPageIdMutableLiveData;
+
+    public FeedFragmentViewModel() {
+        facebookPageIdMutableLiveData = new MutableLiveData<>();
+    }
+
+    public LiveData<FacebookPageId> updateFBPageID(String fbAccessToken) {
+        if (SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_ID, null) == null)
```
iamareebjamal: Don't use SharedPreferenceUtil here, calculate this in fragment and pass it here. It makes it easier to test the class

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+            APIClient.getFacebookGraphAPI().getPageId(SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_NAME, null),
+                    fbAccessToken)
```
iamareebjamal: Here, sending LiveData instead of Observable is no good because you are not disposing the observable properly, It will cause memory leak.

Convert Observable to Flowable and then to LiveData through
`LiveDataReactiveStreams.fromPublisher`

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public Observable<Feed> getPosts(String fields, String fbAccessToken) {
+        return APIClient.getFacebookGraphAPI()
+                .getPosts(SharedPreferencesUtil.getString(ConstantStrings.FACEBOOK_PAGE_ID, null),
+                        fields,
+                        fbAccessToken)
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread());
+    }
```
iamareebjamal: Same here, convert to LiveData before returning

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
mayank8318: Done

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
mayank8318: Done

android/app/src/main/java/org/fossasia/openevent/viewmodels/FeedFragmentViewModel.java
mayank8318: @iamareebjamal Can't implement the `throwable` part for LiveData. I can wrap the LiveData in a custom class with a status code though not sure how effective that will be.

mayank8318:

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -54,15 +54,16 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         View view = super.onCreateView(inflater, container, savedInstanceState);
 
         feedItems = new ArrayList<>();
+        feedFragmentViewModel = ViewModelProviders.of(this).get(FeedFragmentViewModel.class);
         RecyclerView.LayoutManager mLayoutManager = new LinearLayoutManager(getContext());
         feedRecyclerView.setLayoutManager(mLayoutManager);
-        feedAdapter = new FeedAdapter(getContext(), (FeedAdapter.OpenCommentsDialogListener)getActivity(), feedItems);
-        feedAdapter.setOnImageZoomListener((OnImageZoomListener)getActivity());
+        feedAdapter = new FeedAdapter(getContext(), (FeedAdapter.OpenCommentsDialogListener) getActivity(), feedItems);
+        feedAdapter.setOnImageZoomListener((OnImageZoomListener) getActivity());
```
iamareebjamal: Don't cast. Either require the listener during construction or via setter, don't cast

iamareebjamal (changes requested):

mayank8318:
@iamareebjamal Is this what you meant?

iamareebjamal (approved):

PR #2059, opened 2017-12-25
Added ViewModel for the ScheduleFragment

### LINKED ISSUE #2032: Create ViewModel for the ScheduleFragment
**Actual Behaviour**

There is no ViewModel for the ScheduleFragment

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes!

NA

**Other information**

Parent issue #2015

dr0pdb:
Fixes #2032 

Changes: Added ViewModel for the ScheduleFragment


android/app/src/main/java/org/fossasia/openevent/viewmodels/ScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,51 @@
...
+                    eventDatesStringLiveData.setValue(eventDatesString);
+        });
+        return eventDatesStringLiveData;
+    }
+
+    public LiveData<List<Track>> getTracks() {
+        MutableLiveData<List<Track>> tracksLiveData = new MutableLiveData<>();
+        realmRepo.getTracks().addChangeListener((tracks, orderedCollectionChangeSet) -> {
+            tracksLiveData.setValue(tracks);
+        });
```
iamareebjamal: I didn't realize this in other PRs but this approach is wrong. this attaches the listener each time and also creates a new live data as well, the live data and listener should be created once and stored as a reference in the ViewModel and just returned to the view in each call. The onCleared method should be clearing the stored reference instead of getting a new reference and cleared. Please fix this in this PR and also raise issues for previous one as well

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -92,26 +95,25 @@ protected int getLayoutResource() {
...
+                try {
+                    adapter.addFragment(new DayScheduleFragment(),
+                            DateConverter.formatDay(date), date);
+                    adapter.notifyDataSetChanged();
+                } catch (DateTimeParseException pe) {
+                    Timber.e(pe);
+                    Timber.e("Invalid date %s in database", date);
+                }
+            }
+        });
```
iamareebjamal: Logic like this should be in ViewModel

iamareebjamal (changes requested):

dr0pdb:
Changes: Moved more logic from the Fragment to the ViewModel. Now using global LiveData and Listeners. 

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -92,26 +91,20 @@ protected int getLayoutResource() {
...
+        scheduleFragmentViewModel.getEventDatesString().observe(this, eventDatesString -> {
+            int eventDays = eventDatesString.size();
+            for (int i = 0; i < eventDays; i++) {
+                String date = eventDatesString.get(i).first;
+                String formattedDate = eventDatesString.get(i).second;
+                adapter.addFragment(new DayScheduleFragment(),
+                        formattedDate, date);
+                adapter.notifyDataSetChanged();
+            }
+        });
```
iamareebjamal: Move the logic of calculation of the loop inside ViewModel, the view should simply receive `eventDatesString.get(i)`

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -92,26 +91,20 @@ protected int getLayoutResource() {
...
 
-        realmRepo.getTracks().addChangeListener((tracks, orderedCollectionChangeSet) -> {
+        scheduleFragmentViewModel.getTracks().observe(this,tracksList -> {
             this.tracks.clear();
-            this.tracks.addAll(tracks);
+            this.tracks.addAll(tracksList);
             tracksNames = new String[this.tracks.size()];
             isTrackSelected = new boolean[this.tracks.size()];
```
iamareebjamal: These can be moved to ViewModel as well

android/app/src/main/java/org/fossasia/openevent/viewmodels/ScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,79 @@
...
+            });
+        }
+        return tracksLiveData;
+    }
+
+    @Override
+    protected void onCleared() {
+        if (eventDatesRealmResults != null) {
+            eventDatesRealmResults.removeAllChangeListeners();
+            eventDatesRealmResults = null;
```
iamareebjamal: No need to set item to null. we are not passing the instance anywhere, and even if we were, then setting it to null wouldn't have made any difference. Java is copy by value. If this pattern was possible, this would have been such an elegant solution to memory leak that we wouldn't have needed architecture components :stuck_out_tongue_winking_eye:

android/app/src/main/java/org/fossasia/openevent/viewmodels/ScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,79 @@
...
+    public ScheduleFragmentViewModel() {
+        realmRepo = RealmDataRepository.getDefaultInstance();
+    }
+
+    public LiveData<List<Pair<String,String>>> getEventDatesString() {
+        if (eventDatesStringLiveData == null) {
+            eventDatesStringLiveData = new MutableLiveData<>();
+            eventDatesRealmResults = realmRepo.getEventDates();
+            eventDatesRealmResults.addChangeListener((eventDates, orderedCollectionChangeSet) -> {
+                        List<Pair<String,String>> eventDatesString = new ArrayList<>();
```
iamareebjamal: Indentation is off

android/app/src/main/java/org/fossasia/openevent/viewmodels/ScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class ScheduleFragmentViewModel extends ViewModel {
+
+    private RealmDataRepository realmRepo;
+    private MutableLiveData<List<Pair<String,String>>> eventDatesStringLiveData;
+    private MutableLiveData<List<Track>> tracksLiveData;
+    private RealmResults<Track> tracksRealmResults;
+    private RealmResults<EventDates> eventDatesRealmResults;
+
+
```
iamareebjamal: Avoid extra spaces

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
dr0pdb: @iamareebjamal The ```adapter.addFragment``` method needs both the ```date``` as well as the ```formattedDate``` variable that's why i have to pass them both as a ```List``` of ```Pair```. Perhaps i should name the lambda parameter as ```eventDatesStringPair``` instead of ```eventDatesString```. Please explain a bit more about how i can move the logic of calculation to ViewModel.

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
dr0pdb: They are arrays. I don't think they have an ```addAll``` method. Hence i think we will still need a ```for``` loop.

iamareebjamal: I meant these variables should be moved to ViewModel

```
tracksNames = new String[this.tracks.size()];
isTrackSelected = new boolean[this.tracks.size()];
```

But let it be, the implementation should be changed and new issue should be created for changing the filter functionality

dr0pdb: Alright, I will create an issue for the filter functionality

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
iamareebjamal: Move the loop in ViewModel and then in the for loop, create formattedDate and date, then create a pair and pass it here

So in view, it will only look like this:

```Java
scheduleFragmentViewModel.getEventDatesString().observe(datePair -> {
    adapter.addFragment(new DayScheduleFragment(),	datePair.first, datePair.second);
    adapter.notifyDataSetChanged();
});
```

dr0pdb:
Changes: Moved the ```for``` loop to the ViewModel, Fixed indentation and removed redundant null assignments.

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -92,26 +91,14 @@ protected int getLayoutResource() {
...
-                            adapter.addFragment(new DayScheduleFragment(),
-                                    DateConverter.formatDay(date), date);
-                            adapter.notifyDataSetChanged();
-                        } catch (DateTimeParseException pe) {
-                            Timber.e(pe);
-                            Timber.e("Invalid date %s in database", date);
-                        }
-                    }
-                });
+        scheduleFragmentViewModel.getEventDateString().observe(this,datePair -> {
```
iamareebjamal: Add space after comma

android/app/src/main/java/org/fossasia/openevent/viewmodels/ScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,93 @@
...
+    public LiveData<Pair<String,String>> getEventDateString() {
+        if (eventDateStringLivePair == null) {
+            eventDateStringLivePair = new MutableLiveData<>();
+        }
+        if (eventDateStringData == null) {
+            eventDateStringData = new ArrayList<>();
+            eventDatesRealmResults = realmRepo.getEventDates();
+            addEventDatesRealmResultsChangeListener();
+        }else {
```
iamareebjamal: Add space after }

iamareebjamal (changes requested):

dr0pdb:
Changes: Removed formatting related errors.

@iamareebjamal Please review.

iamareebjamal (approved):

PR #2064, opened 2017-12-26
fix: Store LiveData & Listeners as global variables in SpeakerListFragmentViewModel

### LINKED ISSUE #2061: Store LiveData and onChangeListeners as global variables in SpeakerListFragment
**Actual Behaviour**

The LiveData and the onChangeListeners for the SpeakerListFragment ViewModel is created everytime the method is called.

**Expected Behaviour**

It should be stored as a global variable and returned if asked for.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes

NA

For more information see #2059

dr0pdb:
Fixes #2061 

Changes: LiveData and onChangeListeners for Realm are now stored as global variables in SpeakerListFragmentViewModel. Instances are cleared in ```onCleared```.

@iamareebjamal Please review


android/app/src/main/java/org/fossasia/openevent/viewmodels/SpeakersListFragmentViewModel.java
```diff
@@ -19,17 +19,22 @@
...
 
-    public LiveData<List<Speaker>> getSpeakers() {
-        speakersList = new MutableLiveData<>();
-        realmResults = realmRepo.getSpeakers(sortOrderSpeaker());
-        realmResults.addChangeListener((speakers, orderedCollectionChangeSet) -> {
-            speakersList.setValue(speakers);
-        });
+    public LiveData<List<Speaker>> getSpeakers(int sortType) {
+        if (speakersList == null || sortType != speakersListSortType) {
+            speakersList = new MutableLiveData<>();
```
iamareebjamal: new List should only be created if it is null. Make data holders as immutable as possible

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SpeakersListFragmentViewModel.java
dr0pdb: The issue is with the implementation of the Realm Repository. It sends the list of speakers according to the sort order passed to it. If the sort order is changed , then the LiveData containing the previous list becomes useless.

dr0pdb: @iamareebjamal An alternative would be that we can store all the three type of List as LiveData in the ViewModel and return the appropriate one according to the sort order. But this would mean more memory consumption.

iamareebjamal: You should clear instead of creating new

dr0pdb:
Changes: Creating new LiveData only when it's null. 
@iamareebjamal Please review.

iamareebjamal (approved):

PR #2069, opened 2017-12-28
Fix #2062 - Use default RealmLiveData wrapper.

### LINKED ISSUE #2062: Use default RealmLiveData wrapper
https://academy.realm.io/posts/android-architecture-components-and-realm/

Instead of managing it in ViewModel, use the official method

No

mayank8318:
Fixes #2062 

Changes:
Step 1: Implement LiveRealmData for LocationsFragment.
Step 2: Implement LiveRealmData for AboutFragment.
Step 3: Implement LiveRealmData for SpeakerListFragment.
@iamareebjamal Please review and suggest changes. I will repeat that for other view models.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
+++ b/android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
@@ -22,6 +22,7 @@ import io.reactivex.schedulers.Schedulers;
 import io.realm.Case;
 import io.realm.Realm;
 import io.realm.RealmList;
+import io.realm.RealmObject;
 import io.realm.RealmResults;
 import io.realm.Sort;
 import timber.log.Timber;
@@ -561,6 +562,20 @@ public class RealmDataRepository {
 
     public RealmResults<EventDates> getEventDatesSync(){
         return realm.where(EventDates.class).findAll();
+    }
+
+    /**
+     * Convert RealmResults to LiveReamlData
+     */
+    public static <K extends RealmObject> LiveRealmData<K> asLiveData(RealmResults<K> data) {
+        return new LiveRealmData<>(data);
+    }
+
+    /**
+     * Convert RealmObject to LiveReamlDataObject
+     */
+    public static <K extends RealmObject> LiveRealmDataObject<K> asLiveDataForObject(K data) {
+        return new LiveRealmDataObject<>(data);
     }
 
     /**
--- a/android/app/src/main/java/org/fossasia/openevent/dbutils/LiveRealmData.java
+++ b/android/app/src/main/java/org/fossasia/openevent/dbutils/LiveRealmData.java
@@ -1,0 +1,28 @@
+package org.fossasia.openevent.dbutils;
+
+import android.arch.lifecycle.LiveData;
+
+import io.realm.RealmChangeListener;
+import io.realm.RealmModel;
+import io.realm.RealmResults;
+
+public class LiveRealmData<T extends RealmModel> extends LiveData<RealmResults<T>> {
+
+    private RealmResults<T> results;
+    private final RealmChangeListener<RealmResults<T>> listener =
+            results -> setValue(results);
+
+    public LiveRealmData(RealmResults<T> realmResults) {
+        results = realmResults;
+    }
+
+    @Override
+    protected void onActive() {
+        results.addChangeListener(listener);
+    }
+
+    @Override
+    protected void onInactive() {
+        results.removeChangeListener(listener);
+    }
+}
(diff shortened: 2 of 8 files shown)


mayank8318:
App is running in my mobile but Travis says "Compilation failed" ?

iamareebjamal:
```
<===----------> 30% EXECUTING [2m 9s]> :app:compileGoogleplayDebugJavaWithJavac<===----------> 30% EXECUTING [2m 10s]<===----------> 30% EXECUTING [2m 11s]<===----------> 30% EXECUTING [2m 12s]e/travis/build/fossasia/open-event-android/android/app/src/googleplay/java/org/fossasia/openevent/fragments/MapsFragment.java:87: error: cannot find symbol
                .addChangeListener((microlocations, orderedCollectionChangeSet) -> {
                ^
  symbol:   method addChangeListener((microloca[...]; } })
  location: class LiveRealmData<Microlocation>
```

iamareebjamal:
Switch to Google Play variant and compile

mayank8318:
@iamareebjamal Please check :)

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
```diff
@@ -563,6 +563,13 @@ public Completable saveEventDates(List<EventDates> eventDates) {
         return realm.where(EventDates.class).findAll();
     }
 
+    /**
+     * Convert RealmResults to LiveReamlData
+     */
+    public static LiveRealmData<Microlocation> asLiveData(RealmResults<Microlocation> data) {
+        return new LiveRealmData<>(data);
+    }
```
iamareebjamal: You can create a generic method instead

iamareebjamal (changes requested):

mayank8318:
Changes : Use generic method for asLiveData( ).

iamareebjamal (dismissed):

mayank8318:
@iamareebjamal I will repeat it for other view models which are ready.

mayank8318:
Changes : Add wrapper for AboutViewModel.

iamareebjamal:
@mayank8318 Can you try this scenario:

- Go to any page with bookmarked items
- Press on an item to get detail view of that item
- Toggle the bookmark of that item
- Return to the previous screen and see if the state changed there as well

mayank8318:
@iamareebjamal State changes in the previous screen but bookmarks are not visible in the about fragment.
dateList is causing the problem I think.
Solved that by making an ASYNC call but now getBookmarkedSessions() is returning empty result.

mayank8318:
Changes : Make async call for date list.
Bookmarks work when I change to another app and open Open-Event again (when it resumes focus).
@iamareebjamal Help needed! Can't find why this is happening.

iamareebjamal:
I'll look into it and get back to you :smile: Thanks for notifying

iamareebjamal:
That's because everything is asynchronous without chaining. The subscription to sessions is done after the dates are loaded, and hence before the bookmarks can be loaded, their LiveData is returned, and then the transform is done on the already subscribed data, meaning no actual event is sent to the view.

It is like - 
return instanceA;

instanceA.subscribe

After few seconds, instanceA is replaced with instanceB containing the actual bookmark events, but it's of no use because view is already subscribed to instanceA. That's why it is extremely important for shared objects to be strictly immutable. I will recommend you to temporarily make the LiveData fields final and then check out the errors to know what I am talking about. Unfortunately, we cannot make the fields final as we need to map them to RealmLiveData lazily but effectively, there should be one and only one assignment statement for a LiveData

iamareebjamal:
We could have synchronized the events easily with RxJava but it is not as easy with LiveData, so we can have a discussion and come up with a solution because there are a lot of interdependent asynchronous querying going on here. The important point here is to share a single object ALWAYS with the view and then responsibly send events in order. 

Event is loaded -> Send to view, attach date listener
Date is loaded -> Send to view, attach session listener
Sessions are loaded -> Send to view

This way it will work

mayank8318:
Yes got it. DateList code is creating problem since I changed it to an Async call. Using the old code for `dateList` makes sense in my view since it doesn't have a change listener anyways and I tried it too ;)
@iamareebjamal Should I push my changes?

iamareebjamal:
Yes @mayank8318

mayank8318:
Changes : Changed dateList loading back to previous version.

iamareebjamal (dismissed):

mayank8318:
Changes : Add LiveRealmData wrapper for SpeakerListFragmentViewModel.

iamareebjamal (approved):

iamareebjamal:
Great, let's convert other ViewModels in separate PRs

PR #2070, opened 2017-12-28
fix: Search query logic moved to ViewModel in LocationsFragment

### LINKED ISSUE #2043: Implement the search query in ViewModel for LocationFragment
**Actual Behaviour**

Currently the search query in most of the fragments is implemented inside the adapter.

**Expected Behaviour**

The search query should be implemented inside the ViewModel for LocationFragment

**Steps to reproduce it**

NA.

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes, but after i am done with the rest of ```ViewModel``` PR's. Can be assigned to someone else in the meantime.

NA

**Other information**

Please check the discussion out in this PR #2039

dr0pdb:
Fixes #2043 

Changes: Moved the search query logic to the ViewModel from the Adapter for LocationsFragment.


android/app/src/main/java/org/fossasia/openevent/viewmodels/LocationsFragmentViewModel.java
```diff
@@ -23,15 +27,42 @@ public LocationsFragmentViewModel() {
...
+    }
+
+    private void getFilteredLocations() {
+        final String query = searchText.toLowerCase(Locale.getDefault());
+
+        List<Microlocation> filteredLocationsList = Observable.fromIterable(locations.getValue())
+            .filter(microlocation -> microlocation.getName()
+                .toLowerCase(Locale.getDefault())
+                .contains(query))
+            .toList().blockingGet();
```
iamareebjamal: Great work. But don't do blocking get, you can subscribe and do the rest further

iamareebjamal (changes requested):

dr0pdb:
Changes: Instead of using blockingGet, now using subscribe.

@iamareebjamal Please review.

iamareebjamal:
@srv-twry Please resolve conflicts and update the state of PR to current standard. Thanks :smile:

dr0pdb:
@iamareebjamal I'm kind of stuck in this. Can you please guide ?
The problem is since the MutableLiveData has been replaced by Realm wrapper which is then mapped to a LiveData, i am not able to apply filtering on it. The contents of a LiveData cannot be accessed , it gives NPE. How do i get the contents(ArrayList) of the LiveData which wraps an ArrayList and apply filter on it and return another LiveData/MutableLiveData ?

iamareebjamal:
One thing can be to store the result and add listener on data changes in view model itself. But it uses state so it is not a good solution. Another solution is to use https://developer.android.com/reference/android/arch/lifecycle/LiveDataReactiveStreams.html

And convert live data to flowable and filter then back again in livedata

dr0pdb:
@iamareebjamal Thanks, but to convert from LiveData to Flowable, we need the ```LifecycleOwner``` as a parameter. How can i get it inside the ViewModel ? I can pass it from the Fragment in the method call but wouldn't that lead to memory leaks ?

iamareebjamal:
OK, so instead I think we have to use filter on realm result and then convert it to livedata, so hold instance to realm result instead of livedata and filter and wrap in realm livedata every time you return

dr0pdb:
@iamareebjamal I don't think using Realm Wrapper is possible for this kind of implementation. The problem is every time we ask for filtering, the ```Transformations.map``` function creates a new LiveData after filtering from the RealmResult stored in the ViewModel , hence all the observers are lost. Because of this i am getting an empty screen inside the Fragment.
I checked it using ```filteredLocations.hasActiveObservers()``` and it is always false.
The filtering is done correctly and i logged the filtered RealmResult's size, which for no search query comes out to be 16 as expected.
Since there is no way to change the contents of a LiveData without creating a new instance, i feel we have to use MutableLiveData here. 
I have pushed the commit so that you can have a look and may be suggest something if it's an implementation error made by me.

iamareebjamal:
So this is not working correctly, right?

dr0pdb:
yeah, it isn't with the wrapper

iamareebjamal:
OK, so then use mutable livedata then :+1:

I want you to try one more thing though, try making a FilterableRealmLiveData class, and encapsulate what you will do in the ViewModel. it will help us not repeat the code

It will extend MutableLiveData and have a method called `filter(String)`

dr0pdb:
I am stuck at a problem. This is how i am trying to implement the generic class ```FilterableRealmLiveData``` 

<code>

    public class FilterableRealmLiveData<V extends RealmModel> extends MutableLiveData<List<V>> {

        public void filter(RealmResults<V> realmResults, String filterKey) {
            final String query = filterKey.toLowerCase(Locale.getDefault());
            Observable.fromIterable(realmResults)
                    .filter(singleItem -> singleItem.getName()
                            .toLowerCase(Locale.getDefault())
                            .contains(query))
                    .toList().subscribe(filteredList -> {
                       setValue(filteredList);
                        Timber.d("Filtering done total results %d", filteredList.size());
                        if (filteredList.isEmpty()) {
                            Timber.e("No results published. There is an error in query. Check " + 
    getClass().getName() + " filter!");
                        }
                    });
        }
    }
</code>

It isn't compiling with the following message: 

> Error:(20, 45) error: cannot find symbol method getName()
where V is a type-variable:
V extends RealmModel declared in class FilterableRealmLiveData

How do i make sure that it knows that we will only call filter with those Type ```V``` for which ```getName()``` is defined. Is it possible ?

iamareebjamal:
@srv-twry Two suggestions:

- Firstly, don't pass the realm data in the filter method, this class should be exactly like wrapper class, except that it will have a new filter method
- Second, leave the filtering work to caller, now we want `getName` in generic class, tomorrow we'll want `getId`. The key is to ask a predicate in argument of the method `Predicate<V>` and then just pass it to the filter method of the observable. Let the class calling filter implement how it wants to filter the data, we just want to encapsulate the way we are getting, filtering and disposing the data

dr0pdb:
@iamareebjamal I finally got it to work with this
<code>

    public class FilterableRealmLiveData<V> extends MutableLiveData<List<V>> {  
         public void filter(Predicate<V> predicate, Observable<V> observable) {
             observable.filter(predicate::test).toList().subscribe(this::setValue);
         }
    }

</code>

But i am getting the error that ```Predicate``` needs min API level 24 which is too high in my opinion. 
What can we do now ?

iamareebjamal:
Why the Observable needs to be passed to the method?

Use the rx.Predicate

dr0pdb:
>  Why the Observable needs to be passed to the method?

Either i need to pass the realm data or the observable created from the realm data. This is my proposed workflow:
Store the unfiltered data in the ViewModel
Whenever asked for filtering, i use the unfiltered data, create an observable from it and filter out results according to query and return a new List without changing the unfiltered data.
So i need to pass in the Observable to the filter method.

Please review.

iamareebjamal:
Store the unfiltered data in the wrapper itself. Then you don't need to create Observable from outside the class. Think of it like how you can encapsulate maximum functionality inside the wrapper. If the process is common for all types, it can be encapsulated inside the wrapper.

Storing the unfiltered data and creating an observable from it is repeatable and doesn't depend on a particular use case of wrapper -> it is same everywhere -> it can be encapsulated

iamareebjamal:
```Java
class Wrapper<T> extends ... {
    private final unfilteredData;

   constructor(same as other wrappers) {
       initialize the unfiltered data
   }

   filter(Predicate) {
       Create observable from unfiltered data
       filter using predicate
       set value
       add disposable to composite disposable
   }

   onCleared() {
        dispose each connection
   }

}
```

dr0pdb:
Changes: Added unfiltered data inside the ViewModel only. 
@iamareebjamal  Please review.

android/app/src/main/java/org/fossasia/openevent/utils/FilterableRealmLiveData.java
```diff
@@ -0,0 +1,44 @@
...
+            setValue(filteredList);
+            Timber.d("Filtering done total results %d", filteredList.size());
+            if (filteredList.isEmpty()) {
+                Timber.e("No results published. There is an error in query. Check " + getClass().getName() + " filter!");
+            }
+        }));
+    }
+
+    public void disposeCompositeDisposable() {
+        if (compositeDisposable != null && !compositeDisposable.isDisposed()) {
```
iamareebjamal: `compositeDisposable != null` should never be null

Make it final

android/app/src/main/java/org/fossasia/openevent/utils/FilterableRealmLiveData.java
```diff
@@ -0,0 +1,44 @@
...
+ * Allows filtering of the LiveData components using ReactiveX predicate.
+ * Should only be used with LiveData wrapping a List.
+ * @param <V>: The data type of the List.
+ */
+public class FilterableRealmLiveData<V> extends MutableLiveData<List<V>> {
+
+    private List<V> unfilteredData;
+    private CompositeDisposable compositeDisposable;
+
+    public FilterableRealmLiveData(List<V> unfilteredData) {
```
iamareebjamal: Hmm, this should just be like other RealmDataWrappers, it should also handle the realm stuff, like other wrappers. It is just an extension for those guys

This is not handling the registering and unregistering of data 

Please correct me if I'm wrong. Can I use this as a drop in replacement for non filtering wrappers?

android/app/src/main/java/org/fossasia/openevent/utils/FilterableRealmLiveData.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A wrapper around MutableLiveData
+ * Allows filtering of the LiveData components using ReactiveX predicate.
+ * Should only be used with LiveData wrapping a List.
+ * @param <V>: The data type of the List.
+ */
+public class FilterableRealmLiveData<V> extends MutableLiveData<List<V>> {
+
+    private List<V> unfilteredData;
```
iamareebjamal: Instead, store RealmResult

android/app/src/main/java/org/fossasia/openevent/viewmodels/LocationsFragmentViewModel.java
```diff
@@ -37,4 +39,9 @@ public void setSearchText(String searchText) {
         this.searchText = searchText;
     }
 
+    @Override
+    protected void onCleared() {
+        filteredLocations.disposeCompositeDisposable();
```
iamareebjamal: This should be handled by the LiveData itself. That's the point of it being LiveData, it is LifeCycle aware

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/utils/FilterableRealmLiveData.java
dr0pdb: >  Can I use this as a drop in replacement for non filtering wrappers?

I don't think so, I have only handled the filter functionality for the MutableLiveData, everything else is same as the MutableLiveData, hence it doesn't handle the realm stuff. It is an extension for MutableLiveData not a Realm wrapper.

iamareebjamal: It should be just like wrapper with this extra functionality

dr0pdb:
@iamareebjamal I am having one issue, I don't know how to filter a RealmResult and store it back as another RealmResult using RxJava.
Here is what i have done.

```Java

public class FilterableRealmLiveData<V extends RealmModel> extends LiveData<RealmResults<V>> {

    private RealmResults<V> unfilteredData;
    private RealmResults<V> filteredData;
    private final CompositeDisposable compositeDisposable;
    private final RealmChangeListener<RealmResults<V>> listener = results -> setValue(results);

    public FilterableRealmLiveData(RealmResults<V> unfilteredData) {
        this.unfilteredData = unfilteredData;
        this.filteredData = unfilteredData;
        compositeDisposable = new CompositeDisposable();
    }

    public void filter(Predicate<V> predicate) {
        compositeDisposable.add(Observable.fromIterable(unfilteredData).filter(predicate).toList()
                .subscribe( filteredList -> {
                    setValue(filteredList);
                    Timber.d("Filtering done total results %d", filteredList.size());
                    if (filteredList.isEmpty()) {
                        Timber.e("No results published. There is an error in query. Check " + getClass().getName() + " filter!");
                    }
                }));
    }

    @Override
    protected void onActive() {
        filteredData.addChangeListener(listener);
    }

    @Override
    protected void onInactive() {
        filteredData.removeChangeListener(listener);
        if (!compositeDisposable.isDisposed()) {
            compositeDisposable.dispose();
        }
    }
}
```

In the above code i have created a LiveData of the RealmResults exactly like the other wrappers. Now the only problem is in the filter method. The line 
> setValue(filteredList)

returns a List<V> since i am using ```toList()```, hence it throws an error, since we need RealmResults<V> not List<V>.
How to do this filtering ?

srv-twry:
However this code works, the only change is that instead of extending ```LiveData<RealmResults<V>>```
i am extending ```MutableLiveData<List<V>>```.

``` Java
public class FilterableRealmLiveData<V extends RealmModel> extends MutableLiveData<List<V>> {

    private RealmResults<V> unfilteredData;
    private final CompositeDisposable compositeDisposable;

    public FilterableRealmLiveData(RealmResults<V> unfilteredData) {
        this.unfilteredData = unfilteredData;
        compositeDisposable = new CompositeDisposable();
    }

    public void filter(Predicate<V> predicate) {
        compositeDisposable.add(Observable.fromIterable(unfilteredData).filter(predicate).toList()
                .subscribe( filteredList -> {
                    setValue(filteredList);
                    Timber.d("Filtering done total results %d", filteredList.size());
                    if (filteredList.isEmpty()) {
                        Timber.e("No results published. There is an error in query. Check " + getClass().getName() + " filter!");
                    }
                }));
    }

    @Override
    protected void onInactive() {
        if (!compositeDisposable.isDisposed()) {
            compositeDisposable.dispose();
        }
        super.onInactive();
    }
}
```

iamareebjamal:
Great job! 

1. Why do we need to store filteredResults?
2. The solution here would be to store the RealmResults, but make the type of LiveData to be a `List`. RealmResults implements `List` itself so there shouldn't be a problem

Second, easier solution would be to find if RealmResults itself has a way to filter itself from Realm docs and if there is, then using it instead of Observable method

dr0pdb:
> Why do we need to store filteredResults?

Yes you're right we don't need it. Infact it was redundant since i was only assigning it and attaching/detaching a listener to it without using it.

I also realised that we don't need the listener. The Realm's wrapper over LiveData has a RealmResult<V> which has a listener attached to it. Now if that RealmResult changes the listener calls ```setValue``` with the new RealmResult and the value in the LiveData gets updated.

What we're trying to do is store the RealmResult<V> which is unfiltered and the filtered one in the LiveData<List<V>>. Now if the unfiltered data get changed in the database, we don't have any way to update the filtered data since that would need the Predicate<V>. 
Hence this wrapper can't be made a drop-in replacement for realm wrapper.

iamareebjamal:
But let's say we store the predicate, and we listen to the changes in the RealmData inside the LiveData, and once there is a change, we filter the results using predicate, and set the value? Sounds good?

dr0pdb:
@iamareebjamal 

> Sounds good?

Yes, thank you.

Changes: Added the FilterableRealmLiveData wrapper.
I am taking the Predicate<V> as a setter instead of as a parameter in the filter method. I added a null check on the Predicate in the filter method, if it's not null then filter it using the Predicate otherwise use the unfiltered data to set the value. In this way ```Predicate``` is optional and this FilterableRealmLiveData can be used as a drop-in replacement for RealmLiveData.

Please review.

android/app/src/main/java/org/fossasia/openevent/viewmodels/LocationsFragmentViewModel.java
```diff
@@ -5,28 +5,38 @@
...
+        if (!this.searchText.equals(searchText) || filteredLocations == null) {
+            setSearchText(searchText);
+            final String query = searchText.toLowerCase(Locale.getDefault());
+            Predicate<Microlocation> predicate = location -> location.getName()
+                    .toLowerCase(Locale.getDefault())
+                    .contains(query);
+            FilterableRealmLiveData<Microlocation> filterableRealmLiveData = RealmDataRepository.asFilterableLiveData(realmResults);
+            filterableRealmLiveData.setPredicate(predicate);
+            filterableRealmLiveData.filter();
+            filteredLocations = Transformations.map(filterableRealmLiveData, input -> input);
```
iamareebjamal: This needs not be created again and again whenever there is a change in searchText, only the predicate will be changed and filter will be called, and that's why the only public method in the live data needs to be `filter(Predicate)` because these two methods will never be used without each other

iamareebjamal (changes requested):
Great Job! Nicely done

A little last thing remains though

dr0pdb:
Changes: Storing FilterableLiveRealmData as a global variable in ViewModel. Not storing RealmResults<MIcrolocation> anymore in the ViewModel. 
Calling Transformers.map only once.

Please review.

iamareebjamal (approved):

PR #2078, opened 2017-12-31
feat: Added ViewModel in SessionDetailActivity

### LINKED ISSUE #2052: Create ViewModel for the SessionDetailActivity
**Actual Behaviour**

There is no ViewModel for the SessionDetailActivity

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes

NA

**Other information**

Parent issue #2015

dr0pdb:
Fixes #2052 

Changes: Added ViewModel for the SessionDetailActivity


iamareebjamal:
@srv-twry Please update PR to use RealmDataWrapper

dr0pdb:
Sure

android/app/src/main/java/org/fossasia/openevent/viewmodels/SessionDetailActivityViewModel.java
```diff
@@ -0,0 +1,77 @@
...
+    public SessionDetailActivityViewModel() {
+        realmRepo = RealmDataRepository.getDefaultInstance();
+    }
+
+    public LiveData<Session> getSessionById(int id, String loadedFlag) {
+        if (sessionByIdLiveData == null) {
+            sessionByIdLiveData = new MutableLiveData<>();
+            sessionById = realmRepo.getSession(id);
+            sessionById.addChangeListener((RealmChangeListener<Session>) loadedSession -> {
+                if (!loadedSession.isValid())
```
dr0pdb, line 32: @iamareebjamal If i use a RealmWrapper i won't be able to fetch the Session that it wraps, inside the ViewModel and hence i would have to do these checks in the Activity. Is that fine ?  otherwise using MutableLiveData is better.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SessionDetailActivityViewModel.java
```diff
@@ -0,0 +1,77 @@
...
+                    return;
+
+                if (loadedFlag == null || loadedFlag.equals(BY_ID)) {
+                    SharedPreferencesUtil.putInt(ConstantStrings.SESSION_MAP_ID, id);
```
dr0pdb: Similarly here also

android/app/src/main/java/org/fossasia/openevent/viewmodels/SessionDetailActivityViewModel.java
iamareebjamal: MutableLiveData sounds good

dr0pdb:
Changes: Only Updated the branch since it was already implemented with MutableLiveData.

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -151,21 +148,23 @@ public void onCreate(final Bundle savedInstanceState) {
...
+
         fabSessionBookmark.setOnClickListener(view -> {
             if(session == null)
                 return;
 
             if(session.getIsBookmarked()) {
                 Timber.tag(TAG).d("Bookmark Removed");
 
-                realmRepo.setBookmark(session.getId(), false).subscribe();
+                sessionDetailActivityViewModel.setBookmark(session,false);
```
iamareebjamal: Add space after comma

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -151,21 +148,23 @@ public void onCreate(final Bundle savedInstanceState) {
...
                 fabSessionBookmark.setImageResource(R.drawable.ic_bookmark_border_white_24dp);
 
                 Snackbar.make(speakersRecyclerView, R.string.removed_bookmark, Snackbar.LENGTH_SHORT).show();
             } else {
                 Timber.tag(TAG).d("Bookmark Added");
 
-                realmRepo.setBookmark(session.getId(), true).subscribe();
+                sessionDetailActivityViewModel.setBookmark(session,true);
```
iamareebjamal: Same here

android/app/src/main/java/org/fossasia/openevent/viewmodels/SessionDetailActivityViewModel.java
iamareebjamal: Don't use any Android Specific class inside ViewModel. All Android specific work should be done in fragments/activities

iamareebjamal (changes requested):

dr0pdb:
Changes: Added Proper space and moved SharedPreference code to Activity
@iamareebjamal Please review.

iamareebjamal (approved):
@srv-twry Please create an issue for simplifying the logic in this component and removing the reliability on SharedPreference

SharedPreference may only be used as last resort may in onStop in fragment when it is being killed and loaded onStart but not for all immediate actions, only in memory variables should be needed there

dr0pdb:
@iamareebjamal Can you please provide some more details about what other simplifications need to be mentioned in the issue apart from the reliance on SharedPreferences ?

iamareebjamal:
The issue about loading by ID and name can be simplified. Create 2 issues then, we can tackle the simplification issue later on. Basically, if we can reduce the problem into a standard realm interaction and then use a map Transformation, then the issue will be solved. Right now, we are relying on an intermediate system of MutableLiveData, which is the only possibility in current scenario

PR #2081, opened 2018-01-01
feat: Create ViewModel for the TrackSessionsActivity

### LINKED ISSUE #2053: Create ViewModel for TrackSessionsActivity
**Actual Behaviour**

There is no ViewModel for the TrackSessionsActivity

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes

NA

**Other information**

Parent issue #2015

dr0pdb:
Fixes #2053  

Changes: Added ViewModel for the TrackSessionsActivity.
I have implemented the search query inside the ViewModel now, but since the RecyclerView adapter is shared with ```LocationActivity```, i can't delete the filter method inside the Adapter right now. It will be delete in the PR for ```LocationActivity```. 


android/app/src/main/java/org/fossasia/openevent/viewmodels/TrackSessionsActivityViewModel.java
```diff
@@ -0,0 +1,131 @@
...
+            sessions = new ArrayList<>();
+            filteredSessions = new ArrayList<>();
+            getDataFromRepo(searchText);
+            setSearchText(searchText);
+        } else if (!this.searchText.equals(searchText)) {
+            filterData(searchText);
+            setSearchText(searchText);
+        }
+        return trackSessionLiveData;
+    }
```
iamareebjamal: @srv-twry Please update the PR

iamareebjamal (changes requested):

dr0pdb:
@iamareebjamal There is one issue. The Activity needs mainly two objects , the Tracks object and a ```List<Session>```. The ```List<Session>``` should also implement filtering. The ```List<Session>``` is fetched from the tracks object using ```track.getSessions()```. So everytime the tracks object changes inside the database we need to update the ```List<Session>``` also.
Considering all these constraints, here is how i have implemented these.

``` Java
    public LiveData<Track> getTracks(int trackId) {
        if (trackLiveData == null || this.trackId != trackId) {
            this.trackId = trackId;
            track = realmRepo.getTrack(trackId);
            filteredSessionsLiveData = new FilterableRealmLiveData<>(track.getSessions().sort("startsAt"));
            track.addChangeListener((RealmObjectChangeListener<Track>) (track, objectChangeSet) -> {
                filterData(searchText);
            });
            trackLiveData = RealmDataRepository.asLiveDataForObject(track);
        }
        return trackLiveData;
    }

    private void filterData(String searchText) {
        setSearchText(searchText);
        final String query = searchText.toLowerCase(Locale.getDefault());

        Predicate<Session> predicate = session -> session.getTitle()
                .toLowerCase(Locale.getDefault())
                .contains(query);

        filteredSessionsLiveData.filter(predicate);
        calculateUpcomingOngoingPosition();
    }

    public LiveData<List<Session>> getSessions(String searchText) {
        filterData(searchText);
        return filteredSessionsLiveData;
    }
```
Now the problem is finding the upcoming and ongoing positions which needs the contents of the ```List<Session>```. There are two ways:
1. Either i move the logic for finding the upcoming and ongoing positions to the View i.e. Activity.
2. Use MutableLiveData and get it's value inside the ViewModel and then find it inside the ViewModel. In that case i won't be able to use FilterableRealmLiveData and hence i would have to implement filtering on the MutableLiveData or create a wrapper for it also.
What do you think is the proper way to handle this ?

iamareebjamal:
The first time `getTracks` is called, save the reference of LiveData in a separate variable, this is will give you the reference to track which has all the sessions. Now you can find out the upcoming session through this

android/app/src/main/java/org/fossasia/openevent/activities/TrackSessionsActivity.java
```diff
@@ -120,54 +118,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            sessionsRecyclerView.setSystemUiVisibility(View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR);
+        }
+    }
 
-            //coloring status bar icons for marshmallow+ devices
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && (sessionsRecyclerView != null) && (fontColor != Color.WHITE)) {
-                sessionsRecyclerView.setSystemUiVisibility(View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR);
+    private void loadTracks() {
+        trackSessionsActivityViewModel.getTrack(trackId).observe(this, track -> {
+            if (track != null) {
```
iamareebjamal: Don't send track from ViewModel if it was null

android/app/src/main/java/org/fossasia/openevent/viewmodels/TrackSessionsActivityViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+                sessions.addChangeListener(this::calculateUpcomingOngoingPosition);
+            });
+            trackLiveData = RealmDataRepository.asLiveDataForObject(track);
+        }
+        return trackLiveData;
+    }
+
+    private void filterData(String searchText) {
+        if (!this.searchText.equals(searchText) || firstTime) {
+            firstTime = false;
```
iamareebjamal: Why we need firstTime

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/TrackSessionsActivityViewModel.java
dr0pdb: Yes, we don't need it. I was getting a blank screen so i tried to avoid it. I found the actual bug now, it was actually in my implementation of the FilterableRealmLiveData.

dr0pdb:
Changes: Removed null check and firstTime boolean variable.
Fixed FilterableRealmLiveData to set it's value to the unfiltered data in the constructor.

@iamareebjamal Please review.

iamareebjamal (approved):

Issue #1841, opened 2017-07-22
Make the image zoom in on click in the live feed


rohanagarwal94:
**Actual Behaviour**

The user cannot zoom in the pictures in the livefeed.

**Expected Behaviour**

The image should have functionality to allow zoom on click.

**Would you like to work on the issue?**

Yes. @iamareebjamal @the-dagger views?

**Target Milestone to solve the issue**

[milestone 12](https://github.com/fossasia/open-event-android/milestone/35)

iamareebjamal:
It'd be great if this can be done for every imageview

ivary43:
@rohanagarwal94   Can i work on this ?

iamareebjamal:
Yes, please go ahead

dr0pdb:
@iamareebjamal Can i go ahead ?

iamareebjamal:
Yes

dr0pdb:
For some reason this issue wasn't automatically closed on merge. 
@iamareebjamal Please close this as it has been done. Thanks.

iamareebjamal:
Closed by #2084

PR #2084, opened 2018-01-02
fix: Made images zoomable

dr0pdb:
First step for #1841 

Changes: Made the images zoomable in the LiveFeed. 

Library used: [PhotoView](https://github.com/chrisbanes/PhotoView)

Screenshots for the change: 
videotogif_2018 01 02_09 55 43

@iamareebjamal @the-dagger Please review. If it is fine then i will implement it on every image view.


android/app/src/main/java/org/fossasia/openevent/adapters/FeedAdapter.java
```diff
@@ -70,17 +73,22 @@
...
+    public FeedAdapter(Context context, AdapterCallback adapterCallback, ZoomableImageModule zoomableImageModule, List<FeedItem> feedItems) {
         super(feedItems);
         this.feedItems = feedItems;
         this.context = context;
         try {
             this.mAdapterCallback = adapterCallback;
+            this.zoomableImageModule = zoomableImageModule;
         } catch (ClassCastException e) {
-            throw new ClassCastException("Activity must implement AdapterCallback.");
+            throw new ClassCastException("Activity must implement Callbacks.");
```
iamareebjamal: Not good, throwing runtime exceptions based on class casting exception is like using Java like JavaScript. Java is type safe, there should be no need to convert an object in raw type and then check if it is of the correct type. I see that you did not write this implementation, but don't add more to it. There shouldn't be a `ClassCastException` if we expect the right type in the arguments.

Please also open an issue to fix the existing code and removing the exception check

dr0pdb:
Changes: Removed the unnecessary check.
@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/adapters/FeedAdapter.java
```diff
@@ -70,10 +73,14 @@
...
             });
+
+            feedImageView.setOnClickListener(v -> {
+                zoomImage(Utils.parseImageUri(feedItems.get(getPosition()).getFullPicture()));
+            });
         }
     }
 
-    public FeedAdapter(Context context, AdapterCallback adapterCallback, List<FeedItem> feedItems) {
+    public FeedAdapter(Context context, AdapterCallback adapterCallback, ZoomableImageModule zoomableImageModule, List<FeedItem> feedItems) {
```
iamareebjamal: Instead of adding it to constructor, make it optional and ask it as a setter

iamareebjamal: `ZoomableImageModule` is ambiguous. Use name like `OnImageZoomListener` and the method being `onZoom(String)`

android/app/src/main/java/org/fossasia/openevent/adapters/FeedAdapter.java
```diff
@@ -70,10 +73,14 @@
                 else
                     Snackbar.make(v, context.getResources().getString(R.string.no_comments), Snackbar.LENGTH_SHORT).show();
             });
+
+            feedImageView.setOnClickListener(v -> {
+                zoomImage(Utils.parseImageUri(feedItems.get(getPosition()).getFullPicture()));
+            });
```
iamareebjamal: Once you make it optional, just add a null check here as well

iamareebjamal (changes requested):
And you have to remove reference to the ZoomListener is onStop of fragment or activity as well or it will create a memory leak

dr0pdb:
Changes: Renamed the Listener to ```OnImageZoomListener```, Made it optional by setting it by setter, Removed reference in the onStop().

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/adapters/FeedAdapter.java
```diff
@@ -150,4 +158,18 @@ public void onBindViewHolder(RecyclerViewHolder holder, int position) {
...
+    }
+
+    public void removeOnImageZoomListener() {
+        if (onImageZoomListener != null) {
+            onImageZoomListener = null;
+        }
+    }
+
+    private void zoomImage(String imageUri) {
+        onImageZoomListener.onZoom(imageUri);
```
iamareebjamal: Suppose that the ZoomListener was not null at the time of construction but it is removed later on. Now once the image is clicked, NPE would be thrown.

android/app/src/main/java/org/fossasia/openevent/adapters/FeedAdapter.java
```diff
@@ -150,4 +158,18 @@ public void onBindViewHolder(RecyclerViewHolder holder, int position) {
...
     }
+
+    public void setOnImageZoomListener(OnImageZoomListener onImageZoomListener) {
+        this.onImageZoomListener = onImageZoomListener;
+    }
+
+    public void removeOnImageZoomListener() {
+        if (onImageZoomListener != null) {
+            onImageZoomListener = null;
+        }
```
iamareebjamal: No need to add null check here

android/app/src/main/java/org/fossasia/openevent/activities/MainActivity.java
```diff
@@ -874,4 +877,14 @@ public void onMethodCallback(List<CommentItem> commentItems) {
         newFragment.setArguments(bundle);
         newFragment.show(fragmentManager, "Comments");
     }
+
+    @Override
+    public void onZoom(String imageUri) {
+        ZoomableImageDialogFragment zoomableImageDialogFragment = new ZoomableImageDialogFragment();
+        zoomableImageDialogFragment.setStyle(DialogFragment.STYLE_NO_TITLE,0);
+        Bundle bundle = new Bundle();
+        bundle.putString(ConstantStrings.IMAGE_ZOOM_KEY,imageUri);
```
iamareebjamal: Add a space after comma
 
In other places as well

iamareebjamal (changes requested):
Great job. Just a few changes required :smile:

dr0pdb:
Changes: Added null check in ```zoomImage```, removed null check in ```removeOnImageZoomListener``` and added commas.
@iamareebjamal Please review.

iamareebjamal (approved):

PR #2088, opened 2018-01-03
fix : made ViewModel for SearchActivity (fixes #2051)

### LINKED ISSUE #2051: Create ViewModel for SearchActivity
**Actual Behaviour**

There is no ViewModel for the SearchActivity

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes

NA

**Other information**

Parent issue #2015

dreadpool2:
Fixes #2051

Changes: ViewModel for SearchActivity has been created.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
@@ -1,6 +1,7 @@
 package org.fossasia.openevent.activities;
 
 import android.app.SearchManager;
+import android.arch.lifecycle.ViewModelProviders;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
@@ -17,20 +18,13 @@ import android.widget.TextView;
 import org.fossasia.openevent.OpenEventApp;
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.adapters.GlobalSearchAdapter;
-import org.fossasia.openevent.data.Microlocation;
-import org.fossasia.openevent.data.Session;
-import org.fossasia.openevent.data.Speaker;
-import org.fossasia.openevent.data.Track;
+import org.fossasia.openevent.viewmodels.SearchActivityViewModel;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Locale;
 
 import butterknife.BindView;
-import io.realm.Case;
 import io.realm.Realm;
-import io.realm.RealmResults;
-import timber.log.Timber;
 
 public class SearchActivity extends BaseActivity implements SearchView.OnQueryTextListener {
 
@@ -43,6 +37,7 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
     private String searchText;
 
     private final String SEARCH = "SAVE_KEY_ON_ROTATE";
+    private SearchActivityViewModel searchViewModel;
 
     @BindView(R.id.search_recyclerView)
     protected RecyclerView searchRecyclerView;
@@ -57,20 +52,14 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
 
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
+        setUpRecyclerView();
         handleVisibility();
-
-        globalSearchAdapter = new GlobalSearchAdapter(results, this);
-        searchRecyclerView.setAdapter(globalSearchAdapter);
-        searchRecyclerView.addItemDecoration(new DividerItemDecoration(this, DividerItemDecoration.VERTICAL));
-        searchRecyclerView.setLayoutManager(new LinearLayoutManager(this));
-
-        if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
-            searchText = savedInstanceState.getString(SEARCH);
-        }
 
         if (Intent.ACTION_SEARCH.equals(getIntent().getAction())) {
             handleIntent(getIntent());
         }
+        searchViewModel = ViewModelProviders.of(this).get(SearchActivityViewModel.class);
+        searchText = searchViewModel.getSearchText();
     }
 
     @Override
@@ -100,14 +89,15 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
 
     @Override
     public boolean onQueryTextChange(String query) {
-        if (query != null) {
-            searchQuery(query);
-            searchText = query;
-        } else {
-            results.clear();
-            handleVisibility();
-        }
+        searchQuery(query);
         return true;
+    }
+
+    private void setUpRecyclerView() {
+        globalSearchAdapter = new GlobalSearchAdapter(results, this);
+        searchRecyclerView.setAdapter(globalSearchAdapter);
+        searchRecyclerView.addItemDecoration(new DividerItemDecoration(this, DividerItemDecoration.VERTICAL));
+        searchRecyclerView.setLayoutManager(new LinearLayoutManager(this));
     }
 
     @Override
@@ -128,99 +118,21 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
     }
 
     private void searchQuery(String constraint) {
-        results.clear();
-
-        if (!TextUtils.isEmpty(constraint)) {
-            String query = constraint.toLowerCase(Locale.getDefault());
-            String wildcardQuery = String.format("*%s*", query);
-            addResultsFromTracks(wildcardQuery);
-            addResultFromSessions(wildcardQuery);
-            addResultsFromSpeakers(wildcardQuery);
-            addResultsFromLocations(wildcardQuery);
-        }
-    }
-
-    @Override
-    protected void onSaveInstanceState(Bundle outState) {
-        if (searchView != null) {
-            outState.putString(SEARCH, searchText);
-        }
-        super.onSaveInstanceState(outState);
+        searchViewModel.getSearchResults(constraint).observe(SearchActivity.this, searches -> {
+            results.clear();
+            results.add(searches);
+            globalSearchAdapter.setCopyOfSearches(searches);
+            globalSearchAdapter.notifyDataSetChanged();
+            handleVisibility();
+        });
     }
 
     @Override
     public void onDestroy() {
         super.onDestroy();
         OpenEventApp.getEventBus().unregister(this);
-
-        //Closing realm instance and detaching listeners to avoid memory leaks
-        realm.close();
-        realm.removeAllChangeListeners();
-
         //Set listener to null to avoid memory leaks
         if (searchView != null) searchView.setOnQueryTextListener(null);
-    }
-
-    public void addResultsFromTracks(String queryString) {
-        RealmResults<Track> filteredTracks = realm.where(Track.class)
-                .like("name", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
-
-        filteredTracks.addChangeListener(tracks -> {
-            if (tracks.size() > 0) {
-                results.add("Tracks");
-            }
-            results.addAll(tracks);
-            globalSearchAdapter.notifyDataSetChanged();
-            Timber.d("Filtering done total results %d", tracks.size());
-            handleVisibility();
-        });
-    }
-
-    public void addResultsFromSpeakers(String queryString) {
-        RealmResults<Speaker> filteredSpeakers = realm.where(Speaker.class)
-                .like("name", queryString, Case.INSENSITIVE).or()
-                .like("country", queryString, Case.INSENSITIVE).or()
-                .like("organisation", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
-
-        filteredSpeakers.addChangeListener(speakers -> {
-            if (speakers.size() > 0) {
-                results.add("Speakers");
-            }
-            results.addAll(speakers);
-            globalSearchAdapter.notifyDataSetChanged();
-            Timber.d("Filtering done total results %d", speakers.size());
-            handleVisibility();
-        });
-    }
-
-    public void addResultsFromLocations(String queryString) {
-        RealmResults<Microlocation> filteredMicrolocations = realm.where(Microlocation.class)
-                .like("name", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
-
-        filteredMicrolocations.addChangeListener(microlocations -> {
-            if (microlocations.size() > 0) {
-                results.add("Locations");
-            }
-            results.addAll(filteredMicrolocations);
-            globalSearchAdapter.notifyDataSetChanged();
-            Timber.d("Filtering done total results %d", microlocations.size());
-            handleVisibility();
-        });
-    }
-
-    public void addResultFromSessions(String queryString) {
-        RealmResults<Session> filteredSessions = realm.where(Session.class)
-                .like("title", queryString, Case.INSENSITIVE).findAllSortedAsync("title");
-
-        filteredSessions.addChangeListener(sessions -> {
-            if (sessions.size() > 0) {
-                results.add("Sessions");
-            }
-            results.addAll(filteredSessions);
-            globalSearchAdapter.notifyDataSetChanged();
-            Timber.d("Filtering done total results %d", sessions.size());
-            handleVisibility();
-        });
     }
 
     public void handleVisibility() {
@@ -232,5 +144,6 @@ public class SearchActivity extends BaseActivity implements SearchView.OnQueryTe
             searchRecyclerView.setVisibility(View.VISIBLE);
         }
     }
+
 }
 
--- a/android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
+++ b/android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
@@ -1,0 +1,112 @@
+package org.fossasia.openevent.viewmodels;
+
+import android.arch.lifecycle.ViewModel;
+import android.arch.lifecycle.LiveData;
+import android.arch.lifecycle.MutableLiveData;
+import android.text.TextUtils;
+
+import org.fossasia.openevent.data.Microlocation;
+import org.fossasia.openevent.data.Session;
+import org.fossasia.openevent.data.Speaker;
+import org.fossasia.openevent.data.Track;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+import io.realm.Case;
+import io.realm.Realm;
+import io.realm.RealmResults;
+import timber.log.Timber;
+
+public class SearchActivityViewModel extends ViewModel {
+
+    private String searchText = "";
+    private List<Object> objectList = new ArrayList<>();
+    private Realm realm;
+    private MutableLiveData<List<Object>> searchResults = new MutableLiveData<>();
+
+    public SearchActivityViewModel() {
+        realm = Realm.getDefaultInstance();
+    }
+
+    public LiveData<List<Object>> getSearchResults(String search) {
+        objectList.clear();
+        String wildcardQuery = String.format("*%s*", search.toLowerCase(Locale.getDefault()));
+        searchText = search;
+        if (!TextUtils.isEmpty(search)) {
+            addResultFromSessions(wildcardQuery);
+            addResultsFromLocations(wildcardQuery);
+            addResultsFromSpeakers(wildcardQuery);
+            addResultsFromTracks(wildcardQuery);
+        }
+
+        searchResults.setValue(objectList);
+        return searchResults;
+    }
+
+    public String getSearchText() {
+        return searchText;
+    }
+
+    public void addResultsFromTracks(String queryString) {
+        RealmResults<Track> filteredTracks = realm.where(Track.class)
+                .like("name", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
+
+        filteredTracks.addChangeListener(tracks -> {
+            if (tracks.size() > 0) {
+                objectList.add("Tracks");
+            }
+            objectList.addAll(tracks);
+            Timber.d("Filtering done total results %d", tracks.size());
+        });
+    }
+
+    public void addResultsFromSpeakers(String queryString) {
+        RealmResults<Speaker> filteredSpeakers = realm.where(Speaker.class)
+                .like("name", queryString, Case.INSENSITIVE).or()
+                .like("country", queryString, Case.INSENSITIVE).or()
+                .like("organisation", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
+
+        filteredSpeakers.addChangeListener(speakers -> {
+            if (speakers.size() > 0) {
+                objectList.add("Speakers");
+            }
+            objectList.addAll(speakers);
+            Timber.d("Filtering done total results %d", speakers.size());
+        });
+    }
+
+    public void addResultsFromLocations(String queryString) {
+        RealmResults<Microlocation> filteredMicrolocations = realm.where(Microlocation.class)
+                .like("name", queryString, Case.INSENSITIVE).findAllSortedAsync("name");
+
+        filteredMicrolocations.addChangeListener(microlocations -> {
+            if (microlocations.size() > 0) {
+                objectList.add("Locations");
+            }
+            objectList.addAll(microlocations);
+            Timber.d("Filtering done total results %d", microlocations.size());
+        });
+    }
+
+    public void addResultFromSessions(String queryString) {
+        RealmResults<Session> filteredSessions = realm.where(Session.class)
+                .like("title", queryString, Case.INSENSITIVE).findAllSortedAsync("title");
+
+        filteredSessions.addChangeListener(sessions -> {
+            if (sessions.size() > 0) {
+                objectList.add("Sessions");
+            }
+            objectList.addAll(sessions);
+            Timber.d("Filtering done total results %d", sessions.size());
+        });
+    }
+
+    @Override
+    public void onCleared(){
+        //Closing realm instance and detaching listeners to avoid memory leaks
+        realm.close();
+        realm.removeAllChangeListeners();
+    }
+}
--- a/android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
+++ b/android/app/src/main/java/org/fossasia/openevent/adapters/GlobalSearchAdapter.java
@@ -39,6 +39,9 @@ public class GlobalSearchAdapter extends BaseRVAdapter<Object, RecyclerView.View
     public GlobalSearchAdapter(List<Object> dataList, Context context) {
         super(dataList);
         this.context = context;
+    }
+
+    public void setCopyOfSearches(List<Object> dataList){
         this.filteredResultList = dataList;
     }
 


dreadpool2:
@srv-twry @iamareebjamal  please review !

dr0pdb:
@dreadpool2 Please resolve merge conflicts first.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -57,25 +62,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        handleVisibility();
-
-        globalSearchAdapter = new GlobalSearchAdapter(results, this);
-        searchRecyclerView.setAdapter(globalSearchAdapter);
-        searchRecyclerView.addItemDecoration(new DividerItemDecoration(this, DividerItemDecoration.VERTICAL));
-        searchRecyclerView.setLayoutManager(new LinearLayoutManager(this));
+        setUpRecyclerView();
 
         if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
             searchText = savedInstanceState.getString(SEARCH);
```
dr0pdb: Remove this now, you don't need to save the searchText in the ```saveInstanceState```. You can save it inside the ViewModel which will outlast the life of this Activity.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -57,25 +62,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
+             });
+        }
+        handleVisibility();
+
     }
 
     @Override
     protected int getLayoutResource() {
-        return R.layout.activity_search;
+       return R.layout.activity_search;
```
dr0pdb: Indentation is off, infact don't touch the part of the code which your issue isn't associated with.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,81 @@
...
+import io.realm.RealmResults;
+
+/**
+ * Created by sanyog on 1/1/18.
+ */
+
+public class SearchActivityViewModel extends ViewModel {
+
+
+    private MutableLiveData<List<Object>> searchResults;
```
dr0pdb: Don't add unnecessary lines.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,81 @@
...
+        return searchText;
+    }
+
+    public void setSearchText(String searchText) {
+        this.searchText = searchText;
+    }
+
+    @Override
+    protected void onCleared() {
+        super.onCleared();
```
dr0pdb: You should dispose all the changeListeners if you have any, otherwise this method is redundant since it is only calling it's super class.

dr0pdb (changes requested):
Apart from the inline comments that i have made, i would recommend you to revert all the unnecessary changes that you have made, for eg: i don't think you need to make changes in the layout for this issue.
Please focus on the issue only, if you want to make other changes , then please create an issue for it and send us a PR :smile:

dreadpool2:
@srv-twry Sir, Thanks for all the suggested changes  :smile:

dr0pdb:
@dreadpool2 Please don't call me sir :)

dreadpool2:
@srv-twry Is the PR alright for review or should I change anything else ?

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
```diff
@@ -357,6 +357,10 @@ public Session getSessionSync(int sessionId) {
         return realm.where(Session.class).equalTo("id", sessionId).findFirst();
     }
 
+    public RealmResults<Session> getSessions() {
```
dr0pdb: Can you explain why you needed to add this ? I mean how was the session fetched before implementing the ViewModel ?

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import io.realm.RealmResults;
+
+/**
+ * Created by sanyog on 1/1/18.
+ */
+
```
dr0pdb: Please delete this template

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,74 @@
...
+            int c = realmResults_speaker.size();
+            realmResults_microloc = realmRepo.getLocations();
+            objectList.add(realmResults_microloc);
+
+            int d = realmResults_track.size();
+            if(a+b+c+d == 0){
+                return null;
+            }
+            else {
```
dr0pdb: According to standard guidelines, ```else``` should start from the end of the ```if``` block. Don't add a new line.

android/app/src/main/java/org/fossasia/openevent/fragments/LocationsFragment.java
```diff
@@ -77,10 +77,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         //set up view model
         locationsFragmentViewModel = ViewModelProviders.of(this).get(LocationsFragmentViewModel.class);
         searchText = locationsFragmentViewModel.getSearchText();
-        locationsFragmentViewModel.getLocations().observe(LocationsFragment.this, microlocations ->  {
+        locationsFragmentViewModel.getLocations().observe(LocationsFragment.this, microlocations -> {
                 locations.clear();
```
dr0pdb: I don't understand why you had to edit ```LocationsFragment``` if you are creating ViewModel for ```SearchActivity```. Was changing this file absolutely necessary ?

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,74 @@
...
+            realmResults_speaker = realmRepo.getSpeakers();
+            objectList.add(realmResults_speaker);
+
+            int c = realmResults_speaker.size();
+            realmResults_microloc = realmRepo.getLocations();
+            objectList.add(realmResults_microloc);
+
+            int d = realmResults_track.size();
+            if(a+b+c+d == 0){
+                return null;
```
dr0pdb: You're returning ```null```, wouldn't that lead to null pointer exception ?

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
dreadpool2: Like actually I wanted to fetch all microlocations , speakers , tracks and sessions so I created a method for Sessions and Speakers differently as they were not present before (I sorted them according to their title and name respectively). Then I added all of them to an arraylist of object type.

android/app/src/main/java/org/fossasia/openevent/fragments/LocationsFragment.java
dreadpool2: This was a silly error 😅

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: Yeah Sure

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: Actually on the other side, the Search Activity side, I have implemented a conditional which would disregard the null statements.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
dr0pdb: Actually my question is if it wasn't present before , then how was it implemented before ? I mean it was working before, it's just a refactor from View to ViewModel. Ideally there shouldn't be any changes to the repository layer.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -57,20 +62,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        }
+        setUpRecyclerView();
 
         if (Intent.ACTION_SEARCH.equals(getIntent().getAction())) {
             handleIntent(getIntent());
         }
+
+        searchViewModel = ViewModelProviders.of(this).get(SearchActivityViewModel.class);
+        searchText = searchViewModel.getSearchText();
+        if(searchViewModel.getSearchResults()!=null) {
```
dr0pdb: It's not the right way to do it. Infact even after doing this null check, the ```searches``` list could be null. The LiveData is just a wrapper around it. I would suggest you remove this null check, and instead of returning ```null```, return an empty list from the ViewModel

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dr0pdb: I have added another comment. Please check it.

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
dreadpool2: Actually Realm.where(Speaker.class) is used to find the full list. There wasn't a method before like it did in the case of microlocations and tracks 😅

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: Oh, yeah that could be a problem. Sure I will change it 👍

android/app/src/main/java/org/fossasia/openevent/dbutils/RealmDataRepository.java
dr0pdb: Yes, then please don't change the implementation. Implement it the way it was done before. Incase you want to change the implementation, create an issue for it.

dr0pdb: Please find the methods like ```addResultsFromLocations``` to get the idea of how to fetch only ```Locations``` from the database. I think you should first go through and understand it.

dreadpool2:
@srv-twry After I went through the repository carefully I realized I had missed two methods corresponding to both speaker and session items :sweat_smile:

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -123,11 +138,20 @@ protected void onNewIntent(Intent intent) {
     }
 
     public void handleIntent(Intent intent) {
+
         final String query = intent.getStringExtra(SearchManager.QUERY);
         searchQuery(query);
     }
+    @Override
```
iamareebjamal: Add empty line above

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -227,6 +245,7 @@ public void handleVisibility() {
         if (results.size() == 0) {
             noResultsView.setVisibility(View.VISIBLE);
             searchRecyclerView.setVisibility(View.INVISIBLE);
+
```
iamareebjamal: Please don't add unnecessary empty lines

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,71 @@
...
+    private MutableLiveData<List<Object>> searchResults;
+    private RealmDataRepository realmRepo;
+    private String searchText = "";
+
+    public SearchActivityViewModel() {
+        realmRepo = RealmDataRepository.getDefaultInstance();
+    }
+
+    public LiveData<List<Object>> getSearchResults() {
+        RealmResults<Microlocation> realmResults_microloc;
```
iamareebjamal: Please use Java camel case notation to name variables

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,71 @@
...
+            realmResults_microloc = realmRepo.getLocations();
+            objectList.add(realmResults_microloc);
+
+            int d = realmResults_track.size();
+            if(a+b+c+d == 0){
+                return new MutableLiveData<>();
+            }else{
+                 searchResults.setValue(objectList);
+            }
+        }
```
iamareebjamal: The naming is very poor here. Please change it to meaningful names rather than using a b c d e

There's no clue what the variables are used for.

Also, please see that the if cases are correctly used. You are doing everything in null check branch, whereas if the searchResults is not null, none of this will be called and what you will set as a value of objectList will be empty. Null check branch should only be used to instantiate the searchResults object and nothing else

iamareebjamal: Also, in case of the sum being 0, you are sending a new MutableLiveData which is not needed. You should just set the empty value to the existing LiveData

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
iamareebjamal: Also please add spaces where necessary, like around the `else`

dreadpool2:
@iamareebjamal I have done the requested changes. There was no need of an else statement in the results summation part actually :sweat_smile:

dr0pdb (changes requested):
I suggest you please have a look at methods like ```addResultFromSessions``` etc. You need to move this logic to ```ViewModel```. Database interaction isn't supposed to be done inside the Activity.

dreadpool2:
@srv-twry Somehow I am not able to receive the results from the viewmodel. It is returning empty even after a search query being set. Am I missing something?

dr0pdb:
It's difficult to say what could be wrong. What i would do in this situation is to check how other ViewModel's have been implemented and how data is fetched from these ViewModel's by other Fragment/Activity.

dreadpool2:
@srv-twry Actually the thing here is that the data is not fully loaded to be filtered like in other activities or fragments like SpeakerListFragment or LocationFragment. That's why its a little bit different than others :sweat_smile:

dr0pdb:
@dreadpool2 In the ViewModel that you have implemented, i would suggest you try removing the non-empty check on line 47 i.e. ```!searchText.equals("")```. I don't understand the need of this check, so can you explain why it's needed ?

dreadpool2:
@srv-twry I actually did remove it after a brief check as it was indeed not necessary!

dreadpool2:
@srv-twry Actually I imported the addResultFromSessions ..etc methods from SearchActivity to the ViewModel, thus the check was redundant.

dr0pdb:
I guess probably @iamareebjamal could help you.

dreadpool2:
@iamareebjamal Even after connecting both the viewmodel and the SearchActivity successfully, an empty list is being returned. How do I overcome this issue? :sweat:

dreadpool2:
@iamareebjamal Is the PR alright?

iamareebjamal:
@dreadpool2 Sorry I did not get the time to inspect the code. Are you facing any difficulty in completing the goal of the issue or is it complete?

dreadpool2:
@iamareebjamal Sir actually I want to know whether the pushed code is correct as I have a doubt whether there should be any database interaction in SearchActivity.

dreadpool2:
@srv-twry Sorry to bother you again :sweat_smile: .I actually managed to separate the database interaction from the activity. But I am facing problem in the `OnQueryTextChange` method of SearchView. The frustrating thing which is happening is that when I start typing my query, like If I type "a",the results are updated below, but then if I go on typing "ab", the query results of "ab" are appended below the results of "a". Like I am not at all able to clear the previous results present in the RecyclerView. I have tried all the `clear` methods but still the problem occurs.

dreadpool2:
@srv-twry  I have solved it :sweat_smile:

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -100,16 +88,27 @@ public boolean onCreateOptionsMenu(Menu menu) {
 
     @Override
     public boolean onQueryTextChange(String query) {
-        if (query != null) {
-            searchQuery(query);
+        String query1 = query.toLowerCase(Locale.getDefault());
+        String wildcardQuery = String.format("*%s*", query1);
```
dr0pdb: Do all these things in the ViewModel only.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -100,16 +88,27 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+        searchViewModel.setSearchText(wildcardQuery);
+
+        if (!query.equals("")) {
```
dr0pdb: Can you explain the use of this non-empty check ?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +126,27 @@ public void handleIntent(Intent intent) {
         searchQuery(query);
     }
 
+    @Override
+    public void onSaveInstanceState(Bundle bundle) {
+        if (searchView != null && searchViewModel != null) {
```
dr0pdb: No need for a null check on searchView if you aren't using it anywhere.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +126,27 @@ public void handleIntent(Intent intent) {
...
+            searchViewModel.setSearchText(searchText);
+        }
+        super.onSaveInstanceState(bundle);
+    }
+
     private void searchQuery(String constraint) {
         results.clear();
```
dr0pdb: Add this clear method inside the getSearchResults call.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +126,27 @@ public void handleIntent(Intent intent) {
...
+        globalSearchAdapter.clear();
```
dr0pdb: Please explain the significance of this line

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public LiveData<List<Object>> getSearchResults() {
+        objectList.clear();
+        MutableLiveData<List<Object>> searchResults = new MutableLiveData<>();
+
+        addResultFromSessions(searchText);
+        addResultsFromLocations(searchText);
+        addResultsFromSpeakers(searchText);
+        addResultsFromTracks(searchText);
+
```
dr0pdb: Don't add unnecessary lines.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+    private List<Object> objectList = new ArrayList<>();
+    private Realm realm;
+
+    public SearchActivityViewModel() {
+        realm = Realm.getDefaultInstance();
+    }
+
+    public LiveData<List<Object>> getSearchResults() {
+        objectList.clear();
+        MutableLiveData<List<Object>> searchResults = new MutableLiveData<>();
```
dr0pdb: You're creating a LiveData everytime the method is called. Make it global and do it only once.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -100,16 +88,27 @@ public boolean onCreateOptionsMenu(Menu menu) {
 
     @Override
     public boolean onQueryTextChange(String query) {
-        if (query != null) {
-            searchQuery(query);
+        String query1 = query.toLowerCase(Locale.getDefault());
```
dr0pdb: No need for this temporary varialble, why not just add it inside the String.format call

dr0pdb (changes requested):
I couldn't find where you're fetching the searchQuery text from the ViewModel inside the Activity.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +126,27 @@ public void handleIntent(Intent intent) {
...
-    @Override
-    protected void onSaveInstanceState(Bundle outState) {
-        if (searchView != null) {
-            outState.putString(SEARCH, searchText);
+            searchViewModel.getSearchResults().observe(SearchActivity.this, searches -> {
+                results.add(searches);
+                globalSearchAdapter.setCopyOfSearches(searches);
+                globalSearchAdapter.notifyDataSetChanged();
+                handleVisibility();
+            });
```
dreadpool2: @srv-twry I am keeping all the results in the viewmodel. I am fetching those in the above `searchQuery` method

dr0pdb: I asked about the searchText variable that you've stored in the ViewModel, where are you using the ```getSearchText``` method of the ViewModel inside the Activity ?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @srv-twry Sure

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: @srv-twry I actually grouped all the four methods together :sweat_smile:

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @srv-twry Oh,I completely forgot to invoke the function in the `searchQuery` method.

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: @srv-twry Is it correct to do so ?

dr0pdb: No need for that, it's still clear. Remove the empty lines

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @srv-twry I did this as if the query is empty then the `handleVisibility` function will be invoked

dreadpool2: @srv-twry Do I proceed with this ?

dreadpool2:
@srv-twry @iamareebjamal Please review

dr0pdb:
Please resolve conflicts.

dreadpool2:
@srv-twry Resolved the conflicts 😅

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+        filteredSessions.addChangeListener(sessions -> {
+            if (sessions.size() > 0) {
+                objectList.add("Sessions");
+            }
+            objectList.addAll(sessions);
+            Timber.d("Filtering done total results %d", sessions.size());
+        });
+    }
+
+    public void onDestroy(){
```
iamareebjamal: This should be overriden method, `onCleared`

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: @iamareebjamal Done the change!

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +121,28 @@ public void handleIntent(Intent intent) {
...
 
-    @Override
-    protected void onSaveInstanceState(Bundle outState) {
-        if (searchView != null) {
-            outState.putString(SEARCH, searchText);
+    private void searchQuery(String constraint) {
+        searchViewModel.setSearchText(constraint);
+        searchText = searchViewModel.getSearchText();
+
+        if (!TextUtils.isEmpty(constraint)) {
```
iamareebjamal: The ifs and above two lines can be inside ViewModel, there is no need to just update the search text this way

iamareebjamal (commented):
One last bit of change

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
iamareebjamal: Also resolve conflicts

dreadpool2: @iamareebjamal I am setting the search text in the ViewModel as constraint. Why should it be inside the ViewModel ?

dreadpool2: @iamareebjamal And in the next line I am receiving the search text from the ViewModel and as it is a global variable we are using this search text in the `onSaveInstanceState` method, kind of storing it.

dreadpool2:
@srv-twry Then is there any application of the `getSearchText` method in the activity ?. Should I just remove it from both the ViewModel and SearchActivity ?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @iamareebjamal Like I wont receive any results while observing the model until and unless the 'constraint' query is stored in the ViewModel itself !.

iamareebjamal: You can pass the searchText in the method parameter and add the condition in ViewModel instead of activity. Setting the search text and then getting it is not appropriate. I never said about not saving it in the fragment.

About the search being done through the text, that's why I'm saying, when you pass it in the method then set it in the viewmodel itself -> Why is there a need of explicit setting of search text in view model, when it can be done from inside the viewmodel when it is received as an argument

dreadpool2: @iamareebjamal Is the `getSearchText` method redundant then?

dr0pdb: @dreadpool2 It's not redundant. You have to use it when the Activity is getting destroyed.

**Edit** - You can delete it since everytime the search text changes we are updating the string in the ViewModel also.

dreadpool2: @iamareebjamal I beg your pardon but I didn't understand the second part of your answer :sweat_smile: , like how the search text is obtained inside the viewmodel.

dreadpool2:
@srv-twry Why should I keep the search text when the activity itself gets destroyed?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dr0pdb: @dreadpool2 What he means to say is, everytime you want to filter, just pass in the string as an argument to the filter method and update the string inside the ViewModel in that method only. No need for the public method ```setSearchText()```

dr0pdb:
@dreadpool2 Yes, i have edited my above comments. You can delete the setSearchText method.

dr0pdb:
> Then is there any application of the getSearchText method in the activity ?. Should I just remove it from both the ViewModel and SearchActivity ?

Yes, it is used in case of screen rotation. If the screen is rotated the activity will be destroyed but the ViewModel will be alive hence when the activity is recreated, you can fetch the searchText that you had saved in the ViewModel.

dreadpool2:
@srv-twry I have pushed the code. I have done the requested changes.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -159,68 +148,7 @@ public void onDestroy() {
...
-        filteredSessions.addChangeListener(sessions -> {
-            if (sessions.size() > 0) {
-                results.add("Sessions");
-            }
-            results.addAll(filteredSessions);
-            globalSearchAdapter.notifyDataSetChanged();
-            Timber.d("Filtering done total results %d", sessions.size());
-            handleVisibility();
-        });
+        searchText = searchViewModel.getSearchText();
```
dr0pdb: Why are you getting the searchText in the ```onDestroy``` ?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -57,20 +52,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
-            searchText = savedInstanceState.getString(SEARCH);
-        }
+        setUpRecyclerView();
 
         if (Intent.ACTION_SEARCH.equals(getIntent().getAction())) {
             handleIntent(getIntent());
         }
+        searchViewModel = ViewModelProviders.of(this).get(SearchActivityViewModel.class);
     }
```
dr0pdb: Also fetch the searchText from the ViewModel after setting the ViewModel

dr0pdb (changes requested):

dr0pdb:
@dreadpool2 It seems like you're unable to understand the workflow. Here it is in an abstract way:

1. Activity created for the first time
2. ViewModel created with searchText = ""
3. Now you fetch the data. Now the user add a searchQuery.
4. You pass in the searchQuery to the filter method in the ViewModel and update the searchText variable.
5. You return the filtered results.
6. The user rotates the device, the activity is recreated. Now the user will expect you to start again with the filtered data not the original unfiltered data.
7. So you fetch the searchText variable from the ViewModel which would survive the configuration changes. (Replacing the functionality of the onSaveInstanceState method).
8. You fetch the data for the recreated Activity using that searchText.

If you have any problem understanding this, please ask.

dreadpool2:
@srv-twry I did the changes in a hurry. Please ignore my silly errors :sweat:

dreadpool2:
@srv-twry Thanks for such a long comment, it's crystal clear now :smile: . Though I believe I might have finally pushed the right code :sweat_smile:

dr0pdb:
@dreadpool2 Please resolve conflicts also.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +122,23 @@ public void handleIntent(Intent intent) {
...
-        if (!TextUtils.isEmpty(constraint)) {
-            String query = constraint.toLowerCase(Locale.getDefault());
-            String wildcardQuery = String.format("*%s*", query);
-            addResultsFromTracks(wildcardQuery);
-            addResultFromSessions(wildcardQuery);
-            addResultsFromSpeakers(wildcardQuery);
-            addResultsFromLocations(wildcardQuery);
+    @Override
+    public void onSaveInstanceState(Bundle bundle) {
+        if(searchViewModel!=null){
```
dr0pdb: Add space before and after ```!=```

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,25 +122,23 @@ public void handleIntent(Intent intent) {
...
+            searchView.setQuery(searchText,true);
```
dr0pdb: Add space after comma

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -100,7 +88,7 @@ public boolean onCreateOptionsMenu(Menu menu) {
 
     @Override
     public boolean onQueryTextChange(String query) {
-        if (query != null) {
+        if (!TextUtils.isEmpty(query)) {
```
dr0pdb: Can you please explain the use of this check? I mean we are already checking that in the line 37 of the ViewModel. Now if that is empty , then we will get an empty list from the ViewModel and we are already clearing the results List and handeling Visibility in ```searchQuery``` method in the Activity.
I know you haven't written this code but i feel this is redundant and should be removed to keep the code simpler.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
iamareebjamal: Yes, that was what I was saying, it should be inside ViewModel, and if it there, it shouldn't be here anyway

dreadpool2:
@iamareebjamal Did the changes

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -134,103 +123,37 @@ public void handleIntent(Intent intent) {
...
-            String query = constraint.toLowerCase(Locale.getDefault());
-            String wildcardQuery = String.format("*%s*", query);
-            addResultsFromTracks(wildcardQuery);
-            addResultFromSessions(wildcardQuery);
-            addResultsFromSpeakers(wildcardQuery);
-            addResultsFromLocations(wildcardQuery);
+    @Override
+    public void onSaveInstanceState(Bundle bundle) {
+        if(searchViewModel != null){
+            searchView.setQuery(searchText, true);
```
dr0pdb: Can you explain what you're trying to do here ?

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -134,103 +123,37 @@ public void handleIntent(Intent intent) {
...
-        super.onSaveInstanceState(outState);
+    private void searchQuery(String constraint) {
+        searchViewModel.getSearchResults(constraint).observe(SearchActivity.this, searches -> {
+            results.clear();
+            results.add(searches);
+            globalSearchAdapter.setCopyOfSearches(searches);
+            globalSearchAdapter.notifyDataSetChanged();
+            handleVisibility();
+        });
+        handleVisibility();
```
dr0pdb: No need to handle Visibility twice

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @srv-twry As I dont have a `setSearchText` method in viewmodel to save the text there after configuration change, I set the text in searchView to searchText in the `onSaveInstanceState` method

dr0pdb: Well , you could have created the method in case you needed them. Although here in my opinion you don't need to save it anywhere. Everytime you change the searchText in the search bar, the ViewModel is queried with that query and the searchText is saved in the ViewModel. So that means, the ViewModel always has the most updated searchText saved. No need to set it again.

iamareebjamal:
Please resolve conflicts

dreadpool2:
@iamareebjamal Resolved the conflicts 😅

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -127,102 +118,32 @@ public void handleIntent(Intent intent) {
...
-        if (!TextUtils.isEmpty(constraint)) {
-            String query = constraint.toLowerCase(Locale.getDefault());
-            String wildcardQuery = String.format("*%s*", query);
-            addResultsFromTracks(wildcardQuery);
-            addResultFromSessions(wildcardQuery);
-            addResultsFromSpeakers(wildcardQuery);
-            addResultsFromLocations(wildcardQuery);
-        }
+    @Override
+    public void onSaveInstanceState(Bundle bundle) {
```
dr0pdb: Please delete this method now. Since the only thing you're doing is calling super, there is no need for this method.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
dreadpool2: @srv-twry Okay

dreadpool2: @srv-twry But there can be future uses of this method right?

dr0pdb: This is an overridden method, which if not called explicitly for the SearchActivity will automatically be called for the super class. In case there are future requirements , one can easily create it again.

It is mainly used to save states, when orientation changes. Now that we have ViewModels which survive orientation changes, i don't think we will need it. In case we need it , we will create it.

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -107,16 +95,18 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
     public boolean onQueryTextChange(String query) {
-        if (!TextUtils.isEmpty(query)) {
-            searchQuery(query);
-            searchText = query;
-        } else {
-            results.clear();
-            handleVisibility();
-        }
+        searchQuery(query);
+        searchText = query;
```
iamareebjamal: There's no need for this line `searchText = query;`

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -134,103 +124,33 @@ public void handleIntent(Intent intent) {
...
     }
 
     @Override
     public void onDestroy() {
         super.onDestroy();
         OpenEventApp.getEventBus().unregister(this);
-
         //Closing realm instance and detaching listeners to avoid memory leaks
         realm.close();
         realm.removeAllChangeListeners();
```
iamareebjamal: Why are there realm instances here now? Every database interaction should be moved to ViewModel

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public SearchActivityViewModel() {
+        realm = Realm.getDefaultInstance();
+    }
+
+    public LiveData<List<Object>> getSearchResults(String search) {
+        objectList.clear();
+        String wildcardQuery = String.format("*%s*", search.toLowerCase(Locale.getDefault()));
+        searchText = search;
+        if (!TextUtils.isEmpty(wildcardQuery)) {
```
iamareebjamal: If you look closely, wildcardQuery can NEVER be empty. It will always at least be ``**'`. So you need this check before formatting, not after

iamareebjamal (changes requested):

dreadpool2:
@iamareebjamal Did the changes

android/app/src/main/java/org/fossasia/openevent/activities/SearchActivity.java
```diff
@@ -134,103 +121,30 @@ public void handleIntent(Intent intent) {
...
-        if (!TextUtils.isEmpty(constraint)) {
-            String query = constraint.toLowerCase(Locale.getDefault());
-            String wildcardQuery = String.format("*%s*", query);
-            addResultsFromTracks(wildcardQuery);
-            addResultFromSessions(wildcardQuery);
-            addResultsFromSpeakers(wildcardQuery);
-            addResultsFromLocations(wildcardQuery);
-        }
+    @Override
+    public void onSaveInstanceState(Bundle bundle) {
```
dr0pdb: Why is this method still here ?

dr0pdb (changes requested):
Also fix the build

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public SearchActivityViewModel() {
+        realm = Realm.getDefaultInstance();
+    }
+
+    public LiveData<List<Object>> getSearchResults(String search) {
+        objectList.clear();
+        String wildcardQuery = String.format("*%s*", search.toLowerCase(Locale.getDefault()));
+        searchText = search;
+        if (!TextUtils.isEmpty(search)) {
```
iamareebjamal, line 37: As I reviewed previously `!TextUtils.isEmpty(search)` will always be true

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/SearchActivityViewModel.java
dreadpool2: @iamareebjamal No! thats not the formatted string

dreadpool2: @iamareebjamal Its directly the `search` text. I have even checked it locally

iamareebjamal: Alright, my mistake

iamareebjamal (approved):

PR #2119, opened 2018-01-09
Fix #2112 variables are being assigned only once

### LINKED ISSUE #2112: Variables are being assigned twice in AboutFragmentViewModel
**Actual Behaviour**

Live Data variables eventLiveData and sessions are being assigned twice in AboutFragmentViewModel

**Expected Behaviour**

They should be assigned only once

Yes

nikit19:
Fixes #2112 

Changes:variables are being assigned only once

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
+++ b/android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
@@ -1,7 +1,6 @@
 package org.fossasia.openevent.viewmodels;
 
 import android.arch.lifecycle.LiveData;
-import android.arch.lifecycle.MutableLiveData;
 import android.arch.lifecycle.Transformations;
 import android.arch.lifecycle.ViewModel;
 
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
```diff
@@ -28,34 +28,28 @@
...
-        subscribeToSessions(dateListString);
-    }
-
-    private void subscribeToEvent() {
-        LiveRealmDataObject<Event> liveRealmDataObject = RealmDataRepository.asLiveDataForObject(realmRepo.getEvent());
-        eventLiveData = Transformations.map(liveRealmDataObject, input -> input);
+        if (sessions == null) {
+            LiveRealmData<Session> sessionLiveRealmData = RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions());
+            sessions = Transformations.map(sessionLiveRealmData, input -> getSessionsList(dateListString, input));
+        }
```
iamareebjamal: Similar should be done for dateList as well

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
nikit19: Please review now

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
```diff
@@ -28,34 +28,33 @@
...
-        LiveRealmData<Session> sessionLiveRealmData = RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions());
-        sessions = Transformations.map(sessionLiveRealmData, input -> getSessionsList(dates, input));
-    }
-
     private void subscribeToDateList() {
-        RealmResults<EventDates> eventDates = realmRepo.getEventDatesSync();
-        ArrayList<String> dateListString = new ArrayList<>();
-        for (EventDates eventDate : eventDates) {
-            dateListString.add(eventDate.getDate());
+        if (dateList == null || sessions == null) {
```
iamareebjamal: We should not need the subscribe to date list meth9d

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
nikit19: @iamareebjamal  subscribeToDateList method removed. Please review

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
```diff
@@ -28,42 +28,36 @@
...
+            LiveRealmDataObject<Event> liveRealmDataObject = RealmDataRepository.asLiveDataForObject(realmRepo.getEvent());
+            eventLiveData = Transformations.map(liveRealmDataObject, input -> input);
+        }
         return eventLiveData;
     }
 
     public LiveData<ArrayList<String>> getDateList() {
+        if (dateList == null) {
+            ArrayList<String> dateListString = new ArrayList<>();
+            if (dateList == null) {
```
iamareebjamal: Why is the same condition nested?

Also, if using LiveData, use async method to get things and if using synchronous method, then don't return LiveData

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
```diff
@@ -28,42 +28,36 @@
...
+                    dateListString.add(eventDate.getDate());
+                }
+                dateList.setValue(dateListString);
+            }
+        }
         return dateList;
     }
 
     public LiveData<List<Object>> getSessions() {
+        if (sessions == null) {
```
iamareebjamal: Change to getBookmarkedSessions

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/AboutFragmentViewModel.java
nikit19: I have little confusion about the second part. Should I move the LiveRealmData code inside getBookmarkedSessions() and getEvent() to AboutFragmentViewModel() ?

iamareebjamal: No, here you are using getEventDatesSync method which is a blocking call. So you don't need to return livedata. LiveData is for async calls and for data which changes

nikit19: I am having little trouble understanding how to achieve this. If I make the return type of getDateList method void, what changes should I make in the aboutFragment when this method is called ?

iamareebjamal: `getDateList` should directly get the list of dates, because the function called inside is synchronous

```Java
public List<> getDateList() {
    ...
    return dateList;
}
```

nikit19: I think I did it correctly this time. Please review.
Thank you

iamareebjamal (approved):

PR #2127, opened 2018-01-10
feat: Add ViewModel in ChangePasswordActivity

### LINKED ISSUE #2121: Create ViewModel for ChangePasswordActivity
**Actual Behaviour**

There is no ViewModel for ChangePasswordActivity

**Expected Behaviour**

There should be one.

Yes

dr0pdb:
Fixes #2121 
Changes: Added ViewModel in ChangePasswordActivity


android/app/src/main/java/org/fossasia/openevent/activities/auth/ChangePasswordActivity.java
```diff
@@ -94,60 +90,36 @@ private void checkCurrentPassword(String currentPassword) {
...
-                            Timber.d("Error changing password: " + throwable.getMessage());
-                        },
-                        () -> {
-                            showProgressBar(false);
-                            //Old password matched change
-                            changePassword();
-                        },
-                        disposable -> showProgressBar(true));
+        showProgressBar(true);
+        changePasswordActivityViewModel.checkCurrentPassword(currentPassword).observe(this, response -> {
```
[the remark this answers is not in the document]
srv-twry: @iamareebjamal Is it necessary to have a default label ? It's not required here.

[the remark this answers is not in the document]
iamareebjamal: Yup, it is necessary for a fallback, if it is not needed, add it and a comment saying `// Not implemented`

dr0pdb:
Changes: Added default label for both switch statements. 

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/activities/auth/ChangePasswordActivity.java
```diff
@@ -94,60 +90,40 @@ private void checkCurrentPassword(String currentPassword) {
...
+                default:
+                    //Not implemented
+            }
+            showProgressBar(false);
+        });
     }
 
     private void changePassword() {
-        if (newPassword == null)
+        if (newPassword == null){
```
iamareebjamal: Instead, check if `TextUtils.isEmpty`

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+        realmRepo = RealmDataRepository.getDefaultInstance();
+    }
+
+    public LiveData<User> getUser() {
+        if (userLiveData == null) {
+            userLiveData = new MutableLiveData<>();
+            user = realmRepo.getUser();
+            user.addChangeListener(realmModel -> {
+                userLiveData.setValue(user);
+            });
```
iamareebjamal: Use Realm Data Wrapper

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+            e.printStackTrace();
+        }
+        changePasswordDisposable = APIClient.getOpenEventAPI().updateUser(User.builder().id(id).password(newPassword).build(), id)
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(user -> {
+                            RealmDataRepository
+                                    .getDefaultInstance()
+                                    .saveUser(user)
+                                    .subscribe();
```
iamareebjamal: You can chain this call above using `mapCompletable` or equivalent above

iamareebjamal: Or better yet, leave the responsibility of updating user on the `updateUser` method, moving the call there

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    protected void onCleared() {
+        if (checkPasswordDisposable != null && !checkPasswordDisposable.isDisposed()) {
+            checkPasswordDisposable.dispose();
+        }
+        if (changePasswordDisposable != null && !changePasswordDisposable.isDisposed()){
+            changePasswordDisposable.dispose();
+        }
```
iamareebjamal: Add both in a composite disposable and dispose at end

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
dr0pdb: > Or better yet, leave the responsibility of updating user on the updateUser method, moving the call there

Sorry, i didn't understand what you mean by moving the call there. Can you explain a bit more.

dr0pdb:
Changes: Chained update user call using flatMapCompleteable, Using TextUtils.empty instead of null check. Using RealmWrapper and CompositeDisposable.

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
dr0pdb: I have chained the calls using flatMapCompletable.

android/app/src/main/java/org/fossasia/openevent/activities/auth/ChangePasswordActivity.java
```diff
@@ -77,6 +68,12 @@ protected void onCreate(Bundle savedInstanceState) {
         });
     }
 
+    private void getUser() {
+        changePasswordActivityViewModel.getUser().observe(this, user -> {
+            this.user = user;
+        });
+    }
```
iamareebjamal: We can have this function inline in the area called. It is not so big and only used once

android/app/src/main/java/org/fossasia/openevent/activities/auth/ChangePasswordActivity.java
```diff
@@ -77,6 +68,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
     private void saveChanges() {
         if (currentPasswordInput == null || newPasswordInput == null || confirmPasswordInput == null)
```
iamareebjamal: These can't be null I guess, please confirm by removing the checks. If the views are null. I think the app **should** crash for good and not silently keep working

android/app/src/main/java/org/fossasia/openevent/activities/auth/ChangePasswordActivity.java
```diff
@@ -94,60 +91,41 @@ private void checkCurrentPassword(String currentPassword) {
...
+                default:
+                    //Not implemented
+            }
+            showProgressBar(false);
+        });
     }
 
     private void changePassword() {
-        if (newPassword == null)
+        if (TextUtils.isEmpty(newPassword)){
```
iamareebjamal: Move this inside ViewModel

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public ChangePasswordActivityViewModel() {
+        realmRepo = RealmDataRepository.getDefaultInstance();
+        compositeDisposable = new CompositeDisposable();
+    }
+
+    public LiveData<User> getUser() {
+        if (userLiveData == null) {
+            user = realmRepo.getUser();
+            user.addChangeListener(realmModel -> {});
```
iamareebjamal: Why this additional change litener?

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public LiveData<Integer> changePassword(String newPassword) {
+        if (changePasswordResponse == null) {
+            changePasswordResponse = new MutableLiveData<>();
+        }
+        int id = 0;
+        try {
+            id = JWTUtils.getIdentity(AuthUtil.getAuthorization());
+        } catch (JSONException e) {
+            e.printStackTrace();
```
iamareebjamal: You should not handle error this way. Send the error immediately to View and log it using Timber. Because if there's an error here, id would be wrong already. No point in continuing, so move the entire code in try or return early

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
dr0pdb: I checked the method ```getUser```. The comment above it states to add a change listener.
``` Java
/**
     * Returns Future style User which is null
     * To get the contents of User, add an OnRealmChangeListener
     * which notifies about the object state asynchronously
     *
     * @return User Returns User Future
*/
```

But when i removed the changeListener, still i am able to get it to work. If user is null, then how does it still work ?

iamareebjamal: It isn't null. If it'd be null, you would have gotten NPE. It's an error in comment. I didn't check the implementation of the method, so if it is different than the rest, i.e., not asynchronous, maybe we should change the method.

Also, adding change listener is what is done automatically by LiveData wrapper

dr0pdb:
Changes: Moved the getUser functionality inline, Removed changeListener, removed null check.
@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/ChangePasswordActivityViewModel.java
dr0pdb: It is asynchronous.
 > adding change listener is what is done automatically by LiveData wrapper

Yeah, i missed this point.

iamareebjamal (approved):

PR #2135, opened 2018-01-11
memory leak fix inside track fragment

### LINKED ISSUE #2134: Memory leak inside track fragment
**Actual Behaviour**

Recycler View Instances didn't unregister adapterDataObserver

**Expected Behaviour**

Recycler View should unregister and adapter should be removed

yes

simarsingh24:
Fixes #2134


android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
```diff
@@ -103,12 +104,13 @@ private void setUpRecyclerView() {
 
         final StickyRecyclerHeadersDecoration headersDecoration = new StickyRecyclerHeadersDecoration(tracksListAdapter);
         tracksRecyclerView.addItemDecoration(headersDecoration);
-        tracksListAdapter.registerAdapterDataObserver(new RecyclerView.AdapterDataObserver() {
+        adapterDataObserver=new RecyclerView.AdapterDataObserver() {
```
dr0pdb: Please add space before and after ```=```

android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
```diff
@@ -131,6 +133,10 @@ protected int getLayoutResource() {
     public void onDestroyView() {
         super.onDestroyView();
         Utils.unregisterIfUrlValid(this);
+        if(tracksListAdapter!=null){
```
dr0pdb: Add proper spaces here also

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
```diff
@@ -131,6 +133,10 @@ protected int getLayoutResource() {
...
+            if(adapterDataObserver!=null)adapterDataObserver=null;
```
iamareebjamal: No need for a null check here

iamareebjamal (changes requested):

simarsingh24 (commented):
@iamareebjamal  if adapterDataObserver is null and is passed to unregisterAdapterDataObserver would that be erroneous? Although we havent set adapterDataObserver to null anywhere but adding an if block certainly assures we are not calling unregisterAdapterDataObserver when adapterDataObserver is null, your views?

iamareebjamal:
@simarsingh24 

```
if(adapterDataObserver!=null)adapterDataObserver=null;
```

Checking object is null just to set it null makes no sense

android/app/src/main/java/org/fossasia/openevent/fragments/TracksFragment.java
```diff
@@ -131,6 +133,9 @@ protected int getLayoutResource() {
     public void onDestroyView() {
         super.onDestroyView();
         Utils.unregisterIfUrlValid(this);
+        tracksListAdapter.unregisterAdapterDataObserver(adapterDataObserver);
+        adapterDataObserver = null;
+        tracksRecyclerView.setAdapter(null);
```
iamareebjamal: Last 2 lines are not needed

iamareebjamal (changes requested):

iamareebjamal (approved):

PR #2141, opened 2018-01-13
Memory bug fixes

### LINKED ISSUE #2137: Potential memory leak in many fragments
**Actual Behaviour**

Memory leaks in DayScheduleFragment, FeedFragment, LocationsFragment, ScheduleFragment, SpeakerListFragment, SponsorsFragment, TracksFragment.

Recycler View Instances didn't unregister adapterDataObserver

**Expected Behaviour**

Recycler View should unregister and adapter should be removed

yes

simarsingh24:
Fixes #2137


android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -116,7 +116,7 @@ public void handleVisibility() {
     @Override
     public void onDestroyView() {
         super.onDestroyView();
-
+        feedAdapter.setOnImageZoomListener(null);
```
dr0pdb: Actually feedAdapter has a method called ```removeOnImageZoomListener```. Also it is already called inside the ```onStop```, so you don't need this.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
simarsingh24: Oh yeah true! Thanks :)

android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -230,6 +230,8 @@ public void closeFilterBar() {
     public void onDestroyView() {
         super.onDestroyView();
         OpenEventApp.getEventBus().unregister(this);
+        viewPager.removeOnPageChangeListener(onPageChangeListener);
+        viewPager.setAdapter(null);
```
iamareebjamal: Setting adapter to null doesn't help

iamareebjamal (changes requested):
Nice work, just once change

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
```diff
@@ -116,7 +116,6 @@ public void handleVisibility() {
     @Override
     public void onDestroyView() {
         super.onDestroyView();
-
```
dr0pdb, line 120: @simarsingh24 It's advisable to not do changes that are not related to the issue. Even though this one is very small and hence isn't a problem but it can be a problem if done on a large scale. In case you want to make changes, opening an issue would be the recommended way. Just something to keep in mind. :smile:

dr0pdb (approved):

iamareebjamal (approved):

android/app/src/main/java/org/fossasia/openevent/fragments/FeedFragment.java
simarsingh24: Yeah totally! I will keep that in mind :)

PR #2159, opened 2018-01-18
WIP: Added a demo implementation for the refactor

### LINKED ISSUE #2155: Remove dependency on event bus pattern in MainActivity
**Actual Behaviour**

EventBus pattern in the app is abused leading to no testability and hence should be removed. The first step would be to remove it from MainActivity where it is majorly used

**Expected Behaviour**

- [ ] Move event handling to a separate component
- [ ] Wire that component with MainActivity through listener
- [ ] Remove event handling in the separate component

This issue may be broken up in several PRs and hence more issues once taken up (i.e. first PR sent), if need be

Parent Issue: #2152

dr0pdb:
Fixes #2155 

Changes: Added a demo implementation for the refactor.

I have opened this since i would like to know if i am thinking in the right direction before actually carrying out the changes. Once i understand the procedure, i will carry out the changes in this PR only and changes the title of the PR also.

@iamareebjamal Is it in a nutshell the way you want me to remove event handeling in MainActivity ? 
Please ignore other things like bad and confusing naming conventions or commented lines etc. I'll fix them.


iamareebjamal (commented):
Yes, that is the way, but the actual event handling stuff is needed to be done in another component. 

Also, what you have tried to do here is move a URL check in another component, that is not required, that can be done in the activity itself. We don't need callbacks just for knowing if the URL is correct or not. We need to move the majority if code which is not related to the activity itself, like synchronizing and listening to the downloads to a new component.

dr0pdb:
> what you have tried to do here is move a URL check in another component, that is not required

Yes, i only selected it for demo purposes because it was very simple. 

> but the actual JSON handling stuff is needed to be done in another component.

Yes, understood.

dr0pdb:
Closing this as code maintainence is not a priority right now. 
Will re-open it when needed if someone else hasn't already worked on it.

PR #2165, opened 2018-01-18
fix : unnecessary elements in shared text have disappeared

### LINKED ISSUE #1971: Unnecessary elements showing up in shared text
**Actual Behaviour**

Unnecessary elements are showing up while sharing the session details.

**Expected Behaviour**

Relevant Details must show up while sharing details.

**Steps to reproduce it**

We parse the JSON elements correctly and align the details correctly.

**Screenshots of the issue**
screenshot_20171101-022526
screenshot_20171101-022536

Yes, I will work on this issue

dreadpool2:
Fixes #1971 

Changes: The Description part has been changed to a shorter abstract version of it and the Speakers names have been filtered out.

Screenshots for the change: 
screenshot_20180119-041703

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
@@ -358,16 +358,26 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                boolean firstSpeaker = true;
+                for (Speaker speaker : session.getSpeakers()) {
+                    if (!firstSpeaker) {
+                        speakersNameBuilder.append(", ");
+                    }
+                    speakersNameBuilder.append(speaker.getName());
+                    firstSpeaker = false;
+                }
+
                 String startTime = DateConverter.formatDateWithDefault(DateConverter.FORMAT_DATE_COMPLETE, session.getStartsAt());
                 String endTime = DateConverter.formatDateWithDefault(DateConverter.FORMAT_DATE_COMPLETE, session.getEndsAt());
-                String shareText = String.format("Session Track: %s \n" +
-                                "Title: %s \n" +
-                                "Start Time: %s \n" +
-                                "End Time: %s\n" +
-                                "Speakers: %s\n" +
-                                "Location: %s",
-                        trackName, title, startTime, endTime, StringUtils.join(speakers, ", "), location) +
-                        "\nDescription: " + Views.fromHtml(session.getLongAbstract());
+                String shareText = String.format(getResources().getString(R.string.session_track_details) + " %s \n" +
+                                getResources().getString(R.string.title_details) + " %s \n" +
+                                getResources().getString(R.string.start_time_details) + " %s \n" +
+                                getResources().getString(R.string.end_time_details) + " %s \n" +
+                                getResources().getString(R.string.speakers_details) + " %s \n" +
+                                getResources().getString(R.string.location_details) + " %s ",
+                        trackName, title, startTime, endTime, speakersNameBuilder.toString(), location) +
+                        "\n"+getResources().getString(R.string.description_details) + " " + Views.fromHtml(session.getShortAbstract());
 
                 Intent sendIntent = new Intent();
                 sendIntent.setAction(Intent.ACTION_SEND);
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -301,4 +301,14 @@
     <string name="speakers_bio">About me</string>
     <string name="social_media">Social media</string>
     <string name="sessions_title_card">Sessions</string>
+
+    <!--Session-detail-strings-->
+    <string name="session_track_details">Session Track :</string>
+    <string name="title_details">Title :</string>
+    <string name="start_time_details">Start Time :</string>
+    <string name="end_time_details">End Time :</string>
+    <string name="location_details">Location :</string>
+    <string name="speakers_details">Speaker(s) :</string>
+    <string name="description_details">Description :</string>
+
 </resources>


android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,20 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                String speakers = "";
+                for(int i = 0; i < session.getSpeakers().size(); i++){
+                    speakers +=  ((i == 0)?(""):(", ")) + session.getSpeakers().get(i).getName();
```
dr0pdb: We should use ```StringBuilder``` here for concatenation since it is the recommended way in case of a loop. 

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
iamareebjamal: Yes true

dreadpool2:
@iamareebjamal DId the changes. Please review

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,20 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakers  = new StringBuilder("");
+                for(int i = 0; i < session.getSpeakers().size(); i++){
+                    speakers.append(((i == 0)?(""):(", ")) + session.getSpeakers().get(i).getName());
+                }
```
dr0pdb: Where are you converting it back to the String ? After the for loop , convert it to a String using ```toString()```.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
iamareebjamal: No need for `new StringBuilder("")`, `new StringBuilder()` will do

iamareebjamal: Rename `speakers` to `speakersNameBuilder`

iamareebjamal: Use for each loop instead

```
for (Speaker speaker: speakers) {
    // append
}

// remove the last character if builder is not empty
```

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                for(Speaker speaker : session.getSpeakers()){
+                    speakersNameBuilder.append(speaker.getName()+", ");
+                }
+                speakersNameBuilder.delete(speakers.length()-3,speakers.length()-1);
```
dr0pdb: What is the use of this line ? 
The build fails because it can't find the ```speakers``` variable. Is it even working locally ?
Please always run the application after you make changes before pushing.

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
                 String endTime = DateConverter.formatDateWithDefault(DateConverter.FORMAT_DATE_COMPLETE, session.getEndsAt());
-                String shareText = String.format("Session Track: %s \n" +
-                                "Title: %s \n" +
-                                "Start Time: %s \n" +
-                                "End Time: %s\n" +
-                                "Speakers: %s\n" +
-                                "Location: %s",
-                        trackName, title, startTime, endTime, StringUtils.join(speakers, ", "), location) +
-                        "\nDescription: " + Views.fromHtml(session.getLongAbstract());
+                String shareText = String.format(getResources().getString(R.string.session_track_details) + "%s \n" +
```
iamareebjamal: Add a space before %s like all other places

iamareebjamal (changes requested):

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
iamareebjamal: Also add appropriate spacing after for, +, - and comma

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                for(Speaker speaker : session.getSpeakers()){
+                    speakersNameBuilder.append(speaker.getName() + ", ");
+                }
+                speakersNameBuilder.delete(speakersNameBuilder.length() - 2, speakersNameBuilder.length() - 1);
```
dr0pdb: I believe you're deleting the comma and the space after the last name of the speaker.
Please add a comment above this line ```deleting comma and space after last name``` or you can use a clearer way by using a boolean flag.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
dreadpool2: @srv-twry I added the comment 😅

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,23 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                for(Speaker speaker : session.getSpeakers()){
+                    speakersNameBuilder.append(speaker.getName() + ", ");
+                }
+                //deleting comma and space after last name
+                speakersNameBuilder.delete(speakersNameBuilder.length() - 2, speakersNameBuilder.length() - 1);
```
dr0pdb: Even though i asked you to add that comment, now i feel it would be much better to use a boolean flag to add a comma and space i.e. don't add a comma and space when it's not required rather that adding it first and then deleting it.
Right now, the problem is, let's say the session.getSpeakers() returns an empty list. Then that for loop won't run at all but still you will try to delete something that isn't there and that would lead to error.

Can you use a boolean flag and only add comma and space when required ?

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
dreadpool2: @srv-twry The list will always have a speaker right!

dr0pdb: I hope it has, but always good to be safe.

dreadpool2: @srv-twry Well I could put a null check ?

dreadpool2: @srv-twry Or I could implement your suggested method.

dr0pdb: The problem won't be resolved with a null check. Here is the problem:
1. You created a speakerNameBuilder object with empty string.
2. Now the session.getSpeakers() returns an empty list.
3. You try to delete the last two characters of the speakerNameBuilder, but wait !, it's empty. So you will get the error. It's like you're trying to delete something which isn't there.

Got it ?

The speakerNameBuilder won't be null but can be empty.

dreadpool2: @srv-twry Yeah I did the changes 😅

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,26 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                for(Speaker speaker : session.getSpeakers()){
+                    if(speakersNameBuilder.toString().equals("")){
```
dr0pdb: This(toString) is O(n) for each operation. So the whole ```for``` loop will be O(n^2) which is bad. Please use a boolean flag.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
dreadpool2: @srv-twry Using a boolean flag might involve another integer variable and `.size()` method too. Is it still advisable?

dr0pdb: It seems you haven't understood what i am suggesting. So here is the full implementation.

``` Java
boolean first = true;
for(Speaker speaker : session.getSpeakers()){
    if(!first) {
         speakersNameBuilder.append(", ");
    }
    speakersNameBuilder.append(speaker.getName());
    first = false;
}
```

Ignore indentation/formatting errors here.

dreadpool2: @srv-twry Wow! That was neat! 😄

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,26 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
 
             case R.id.action_share:
+                StringBuilder speakersNameBuilder  = new StringBuilder();
+                boolean firstSpeaker = true;
+                for(Speaker speaker : session.getSpeakers()){
```
iamareebjamal: Add a space after for and closing brace

android/app/src/main/java/org/fossasia/openevent/activities/SessionDetailActivity.java
```diff
@@ -358,16 +358,26 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                    if(!firstSpeaker) {
```
iamareebjamal: Add space after if

iamareebjamal (changes requested):

dreadpool2:
@iamareebjamal did the changes

iamareebjamal (approved):

PR #2181, opened 2018-01-20
feat: Add ViewModel in SignUpActivity

### LINKED ISSUE #2123: Create ViewModel for SignUpActivity
**Actual Behaviour**

There is no ViewModel in SignUpActivity

**Expected Behaviour**

There should be one.

Yes!

### LINKED ISSUE #2124: Create ViewModel for LoginActivity
**Actual Behaviour**

There is no ViewModel in LoginActivity

**Expected Behaviour**

There should be one.

Yes!

dr0pdb:
Fixes #2123 

Changes: Added ViewModel class for SignUpActivity

@iamareebjamal Please review.


iamareebjamal (changes requested):
Move the actual API call in ViewModel

dr0pdb:
Changes: Moved the actual API call to the ViewModel
Since the ```loginUser``` method inside AuthUtils is also used by the LoginActivity, hence i couldn't edit it. So i created another method ```loginUserWithoutProgressBar```. It will be renamed to ```loginUser``` and the actual ```loginUser``` method will be deleted.
Since the ```signUpUser``` method was only used by the SignUpActivity hence i modified it.

GIF:
videotogif_2018 01 22_15 09 42


@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/activities/auth/SignUpActivity.java
```diff
@@ -61,6 +64,8 @@ protected void onCreate(Bundle savedInstanceState) {
         createPasswordInput = (TextInputEditText) createPasswordWrapper.getEditText();
         confirmPasswordInput = (TextInputEditText) confirmPasswordWrapper.getEditText();
 
+        signUpActivityViewModel = ViewModelProviders.of(this).get(SignUpActivityViewModel.class);
+
```
simarsingh24, line 72: Please remove this space

dr0pdb: @simarsingh24 The aim is to make things clearer to the reader. If you observe the whole method closely(the parts that are left untouched in this PR), you will realise that spaces have been provided between unrelated things. For eg:

```Java

setSupportActionBar(toolbar);
getSupportActionBar().setDisplayHomeAsUpEnabled(true);

switchToLogin.setOnClickListener(v -> startActivity(new Intent(SignUpActivity.this, LoginActivity.class)));
```

So to remain consistent with what has been done, i think it's better to add a space as that line is unrelated to anything above/below it.

simarsingh24:
Why don't you use a boolean for progress bar inside` loginUser` method of AuthUtils and if it is being used by LoginActivity why not considering making some changes there also otherwise it would cause duplication of code!? @srv-twry what do you think?

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/SignUpActivityViewModel.java
```diff
@@ -0,0 +1,60 @@
+package org.fossasia.openevent.viewmodels.auth;
+
+import android.arch.lifecycle.LiveData;
+import android.arch.lifecycle.MutableLiveData;
+import android.arch.lifecycle.ViewModel;
+
```
simarsingh24: remove this space please

dr0pdb: This space is automatically provided by Android studio. Android studio automatically groups the similar imports as one and adds spaces between unrelated imports .It's suitable not to change it.

android/app/src/main/java/org/fossasia/openevent/activities/auth/SignUpActivity.java
simarsingh24: if that's the idea then its fine :)

dr0pdb:
@simarsingh24 

> Why don't you use a boolean for progress bar insideloginUser method of AuthUtils 

Pardon, didn't get you. Can you please elaborate.

> if it is being used by LoginActivity why not considering making some changes there also otherwise it would cause duplication of code!?

Because that is not the aim of this PR. Editing LoginActivity interaction should be done in a separate PR. Infact since we will adding a ViewModel for LoginActivity, it's better to refactor it then.

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +36,45 @@
...
                             showMessage(R.string.error_in_signing_up);
                             Timber.d(throwable.toString());
+                            signUpResponse.setValue(SignUpActivityViewModel.INVALID);
                         },
                         () -> {
-                            showProgressBar(progressBar, false);
                             showMessage(R.string.signed_up_successfully);
                             Timber.d("Signed up successfully");
+                            signUpResponse.setValue(SignUpActivityViewModel.VALID);
+                        });
```
iamareebjamal: Since this observable emits a single, completed handler is same as onNext handler, move this up to `user -> {}`

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +36,45 @@
...
+    }
+
+    public static void loginUserWithoutProgressBar(String email, String password, MutableLiveData<Integer> loginResponse) {
+        APIClient.getOpenEventAPI().login(new Login(email, password))
```
iamareebjamal: Add to composite disposable

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +36,45 @@
...
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(
+                        response -> {
+                            //Save token & email in shared preferences
+                            SharedPreferencesUtil.putString(ConstantStrings.TOKEN, response.getAccessToken());
+                            SharedPreferencesUtil.putString(ConstantStrings.USER_EMAIL, email);
```
iamareebjamal: Don't use Android-specific classes in ViewModels

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +36,45 @@
 
     private static Authenticator authenticator;
 
-    public static void signUpUser(Context context, String email, String password, ProgressBar progressBar) {
+    public static void signUpUser(String email, String password, MutableLiveData<Integer> signUpResponse) {
```
iamareebjamal: The idea is to deconstruct AuthUtil class, move this whole method inside the signup viewmodel. I mean there is no point of hoisting a single use method in a different class

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/SignUpActivityViewModel.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        AuthUtil.signUpUser(email, password, signUpResponse);
+        return signUpResponse;
+    }
+
+    public LiveData<Integer> loginUserAfterSignUp(String email, String password) {
+        if (loginResponse == null) {
+            loginResponse = new MutableLiveData<>();
+        }
+        AuthUtil.loginUserWithoutProgressBar(email, password, loginResponse);
```
iamareebjamal: If this is not used anywhere else. Move this in here please

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
dr0pdb: It is using the Application context which i think shouldn't be a problem. Any other reason why we should avoid it in ViewModels ?
I'll move it to the view though.

android/app/src/main/java/org/fossasia/openevent/viewmodels/auth/SignUpActivityViewModel.java
dr0pdb: It will be used in the future by LoginActivity. 
I had to refactor both the signUp and loginUser methods. signUp user method was only used by SignUpActivity so i did it. The loginUser method was also used by the LoginActivity, hence i couldn't afford editing that. Hence i created a temporary method loginUserWithoutProgressBar to serve the purpose.
The idea is to remove the loginUser method and rename this loginUserWithoutProgressBar method to loginUser and make the LoginActivity use this method using it's ViewModel once it is created.

dr0pdb:
Changes: Moved signUpUser code to ViewModel, Using composite disposable, Using Android-specific classes inside the View, Moved onComplete handler code to onNext.

@iamareebjamal Please review.

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +37,22 @@
...
-                        },
-                        () -> {
-                            showProgressBar(progressBar, false);
-                            showMessage(R.string.signed_up_successfully);
-                            Timber.d("Signed up successfully");
-                        },
-                        disposable -> showProgressBar(progressBar, true));
+                            loginResponse.setValue(new Pair<>(SignUpActivityViewModel.INVALID, ""));
+                        }));
+        Timber.d(email);
```
iamareebjamal: Please remove

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +37,22 @@
...
-        APIClient.getOpenEventAPI().signUp(User.builder().email(email).password(password).build())
+    public static void loginUserWithoutProgressBar(String email, String password,
+                                                   MutableLiveData<Pair<Integer,String>> loginResponse,
+                                                   CompositeDisposable compositeDisposable) {
+        compositeDisposable.add(APIClient.getOpenEventAPI().login(new Login(email, password))
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe(
-                        user -> loginUser(context, email, password, progressBar),
+                        response -> {
```
iamareebjamal: Move to previous line

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
```diff
@@ -34,23 +37,22 @@
 
     private static Authenticator authenticator;
 
-    public static void signUpUser(Context context, String email, String password, ProgressBar progressBar) {
-        APIClient.getOpenEventAPI().signUp(User.builder().email(email).password(password).build())
+    public static void loginUserWithoutProgressBar(String email, String password,
+                                                   MutableLiveData<Pair<Integer,String>> loginResponse,
+                                                   CompositeDisposable compositeDisposable) {
+        compositeDisposable.add(APIClient.getOpenEventAPI().login(new Login(email, password))
```
iamareebjamal: Please move this entire code to ViewModel itself. Is this code used anywhere else as well?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
dr0pdb: No, but it will be used by the LoginActivity when it's ViewModel is implemented.

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
dr0pdb: Sorry, didn't get you. Please explain a bit more.

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
iamareebjamal: For now, move it into ViewModel because it is simply taking the environment of ViewModel and passing into the arguments. We'll think of something better when there is time for sharing code

android/app/src/main/java/org/fossasia/openevent/utils/AuthUtil.java
iamareebjamal: response -> can be moved to previous line, i.e., subscribe(response ->

dr0pdb:
Changes: Moved loginUser code to ViewModel and minor formatting changes. Removed logging of email
@iamareebjamal Please review.

iamareebjamal (approved):

PR #2194, opened 2018-01-22
Download Latest Data bug fix

### LINKED ISSUE #2175: Nothing happens on selecting DownloadLatestData in menu
**Actual Behaviour**

Selecting DownloadLatestData menu item does nothing once the data has been downloaded.

**Expected Behaviour**

It should show the data fetch dialog.

Yes

@iamareebjamal am not sure if it has been made that way !?

simarsingh24:
Fixes #2175


dr0pdb (approved):

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
```diff
@@ -254,6 +256,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 displaySpeakersCallInformation();
                 break;
             case R.id.action_download_latest_data:
+                SharedPreferencesUtil.putBoolean(ConstantStrings.IS_DOWNLOAD_DONE, false);
```
iamareebjamal: This is not the correct way. Download may be done. Saving the wrong information is ... wrong. Simply show the dialog even if the data is downloaded, i.e., don't check if it is downloaded. 

Second this you would have to do is that currently, if user denies to download latest data, it is loaded from JSON. But if the data is already downloaded, it should not be loaded from JSON, so when handling negative response, add the check if the data is downloaded. If it is not, then load from JSON, otherwise do nothing

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/fragments/AboutFragment.java
simarsingh24: @iamareebjamal Yeah true! Should I create a separate method that will take into account everything you mentioned then?

iamareebjamal: Yes

simarsingh24:
@iamareebjamal please review

iamareebjamal:
There is too much repetition of code. Try to reduce it

simarsingh24:
@iamareebjamal yeah I also felt that, a better approach would be passing a boolean `needChecking` to the original `downloadData()` which would decide wether to check from  SharedPreference or not ! what do you say?

iamareebjamal:
Simply make a method which unconditionally shows a dialog, and where we don't need to show it, add the check before calling

simarsingh24:
Changes: now `downloadData` shows dialog irrespective of the `IS_DOWNLOAD_DONE` status, which is taken care by calling function now.

iamareebjamal (approved):

mayank8318 (approved):

PR #2211, opened 2018-01-25
fix: Using AuthUtils in SignUpActivity and removed usage of Pair in ViewModels

### LINKED ISSUE #2200: Refactor the SignUpActivityViewModel and remove all usage of Pair from ViewModels
**Actual Behaviour**

Currently the Pair class is being used inside the SignUpActivityViewModel. Also it has code used by both itself and LoginActivity.

**Expected Behaviour**

Pair shouldn't be used as it is android specific. Also remove it's usage in other ViewModels.
Delete common methods and use the common code present in AuthUtils.

Yes

Can only be done after #2197 is merged.

dr0pdb:
Fixes #2200 

Changes: Using AuthUtils in SignUpActivity. Removed the usage of Pair class in ViewModels.

@iamareebjamal Please review.


android/app/src/main/java/org/fossasia/openevent/fragments/ScheduleFragment.java
```diff
@@ -100,7 +100,7 @@ private void setupViewPager(final ViewPager viewPager) {
 
         scheduleFragmentViewModel.getEventDateString().observe(this, datePair -> {
```
iamareebjamal: Rename datePair

iamareebjamal (changes requested):

dr0pdb:
Renamed datePair to dateStrings.
@iamareebjamal Please review.

iamareebjamal (approved):

PR #2212, opened 2018-01-25
IS_DOWNLOAD_DONE moved inside method block

### LINKED ISSUE #2195: Moving IS_DOWNLOAD_DONE logic inside methods in MainActivity
**Expected Behaviour**

`IS_DOWNLOAD_DONE` should be set true or false inside the methods `downloadFromAssest()` and `downloadFromNetwork()` depending on whether these events were successful or not.

Yes

This way we don't have to put some value in it every time we call this function like we are doing here inside mainActivity
 `if (Utils.isBaseUrlEmpty()) {
            if (!SharedPreferencesUtil.getBoolean(ConstantStrings.IS_DOWNLOAD_DONE, false)) {
                downloadFromAssets();
                SharedPreferencesUtil.putBoolean(ConstantStrings.IS_DOWNLOAD_DONE, true);
            }`

@iamareebjamal your views?

simarsingh24:
Fixes #2195 
Changes: moved IS_DOWNLOAD_DONE inside the method logic

@iamareebjamal  didn't add anything for `downloadFromNetwork` because the statement was already present in `syncComplete`
`SharedPreferencesUtil.putBoolean(ConstantStrings.IS_DOWNLOAD_DONE, true);`
please review if this is the right way to do this?


mayank8318 (commented):
LGTM. Just squash the commits :)

PR #2217, opened 2018-01-25
fix: correct text size of the toolbar text is displayed in landscape

### LINKED ISSUE #2036: Text appears smaller in toolbar on rotation
**Actual Behaviour**

Currently if we keep our phone in landscape mode and open any activity,the text size of the toolbar is decreased (on rotating the screen normally ie after opening the activity, the text size remains the same)

**Expected Behaviour**

The text size should be same in all cases.

**Steps to reproduce it**

Keep your phone in landscape mode.
Open any activity.

**Screenshots of the issue**

screenshot_20171222-125933

ezgif com-video-to-gif 27

Yes

nikit19:
Fixes #2036 

Changes: correct text size of the toolbar text is displayed in landscape mode in all activities

Screenshots for the change: 
screenshot_20180125-235822
screenshot_20180125-235838

### DIFF
--- a/android/app/src/main/res/layout-sw600dp/activity_main.xml
+++ b/android/app/src/main/res/layout-sw600dp/activity_main.xml
@@ -37,6 +37,7 @@
                     style="@style/ToolbarTheme"
                     android:layout_width="match_parent"
                     android:layout_height="?attr/actionBarSize"
+                    app:titleTextAppearance="@style/ToolbarTitle"
                     app:popupTheme="@style/PopupOverlay" />
 
                 <ProgressBar
--- a/android/app/src/main/res/layout/activity_locations.xml
+++ b/android/app/src/main/res/layout/activity_locations.xml
@@ -21,6 +21,7 @@
             style="@style/ToolbarTheme"
             android:layout_width="match_parent"
             android:layout_height="?attr/actionBarSize"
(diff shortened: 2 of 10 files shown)


iamareebjamal:
This would mean adding this style to add new components. Isn't there a better way to handle this? Why is this happening in the first place

nikit19:
@iamareebjamal I guess this is the default behaviour in AppCompat,it has different text sizes for landscape and portrait. 
https://stackoverflow.com/questions/28042331/android-toolbar-small-title-text-in-landscape-mode

mayank8318:
Yes this seems to be the only way of overriding it.

iamareebjamal:
If this is a standard by AppCompat, let it remain to be so as I don't think it is a bug in the implementation. The clients of the generator would want the app to be as standard as possible

nikit19:
Currently the textsize is same in portrait and landscape mode but if you keep your phone in landscape mode and open a new activity,the textSize becomes small for both portrait and landscape mode until that activity is destroyed.
Getting a new textSize for the toolbar seems to be a bug.

I guess the standard was to always have a smaller textSize in landscape mode but if I am right that doesn't happen now as textSize is automatically kept same for both landscape and portrait mode

iamareebjamal (approved):

PR #2218, opened 2018-01-26
Fix #2202 - Shift bindView of SessionListAdapter to its view holder.

### LINKED ISSUE #2202: Shift bindView of SessionListAdapter to its view holder
**Actual Behaviour**

Currently bindView is written in the adapter itself.

**Expected Behaviour**

Shift bindView to its view holder class as is done in other adapters.

Yes

mayank8318:
Fixes #2202 
Changes: Move code of bindView from adapter to view holder.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionViewHolder.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionViewHolder.java
@@ -1,66 +1,276 @@
 package org.fossasia.openevent.core.track.session;
 
+import android.content.Context;
+import android.graphics.Color;
+import android.graphics.PorterDuff;
+import android.support.v4.content.ContextCompat;
 import android.support.v7.widget.CardView;
 import android.support.v7.widget.RecyclerView;
+import android.text.TextUtils;
 import android.view.View;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import com.amulyakhare.textdrawable.TextDrawable;
+
 import org.fossasia.openevent.R;
+import org.fossasia.openevent.common.date.DateConverter;
+import org.fossasia.openevent.common.date.DateService;
+import org.fossasia.openevent.common.notification.NotificationUtil;
+import org.fossasia.openevent.common.ui.Views;
+import org.fossasia.openevent.common.ui.WidgetUpdater;
+import org.fossasia.openevent.common.utils.Utils;
+import org.fossasia.openevent.core.bookmark.BookmarkStatus;
+import org.fossasia.openevent.core.bookmark.OnBookmarkSelectedListener;
+import org.fossasia.openevent.data.Session;
+import org.fossasia.openevent.data.Speaker;
+import org.fossasia.openevent.data.Track;
+import org.fossasia.openevent.data.repository.RealmDataRepository;
+import org.threeten.bp.ZonedDateTime;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
+import io.reactivex.Observable;
+import timber.log.Timber;
+
+import static org.fossasia.openevent.core.bookmark.BookmarkStatus.Status.CODE_ERROR;
+import static org.fossasia.openevent.core.bookmark.BookmarkStatus.Status.CODE_UNDO_ADDED;
+
 
 public class SessionViewHolder extends RecyclerView.ViewHolder {
 
     @BindView(R.id.session_title)
-    public TextView sessionTitle;
+    TextView sessionTitle;
 
     @BindView(R.id.session_subtitle)
-    public TextView sessionSubtitle;
+    TextView sessionSubtitle;
 
     @BindView(R.id.trackImageDrawable)
-    public ImageView trackImageIcon;
+    ImageView trackImageIcon;
 
     @BindView(R.id.session_track)
-    public TextView sessionTrack;
+    TextView sessionTrack;
 
     @BindView(R.id.session_date)
-    public TextView sessionDate;
+    TextView sessionDate;
 
     @BindView(R.id.session_speaker)
-    public TextView sessionSpeaker;
+    TextView sessionSpeaker;
 
     @BindView(R.id.icon_speaker)
-    public ImageView speakerIcon;
+    ImageView speakerIcon;
 
     @BindView(R.id.icon_location)
-    public ImageView locationIcon;
+    ImageView locationIcon;
 
     @BindView(R.id.session_time)
-    public TextView sessionTime;
+    TextView sessionTime;
 
     @BindView(R.id.session_location)
-    public TextView sessionLocation;
+    TextView sessionLocation;
 
     @BindView(R.id.session_bookmark_status)
-    public ImageView sessionBookmarkIcon;
+    ImageView sessionBookmarkIcon;
 
     @BindView(R.id.session_details)
-    public LinearLayout sessionDetailsHolder;
+    LinearLayout sessionDetailsHolder;
 
     @BindView(R.id.session_card)
-    public CardView sessionCard;
+    CardView sessionCard;
 
     @BindView(R.id.titleLinearLayout)
-    public LinearLayout sessionHeader;
+    LinearLayout sessionHeader;
 
     @BindView(R.id.session_status)
-    public TextView sessionStatus;
+    TextView sessionStatus;
 
-    public SessionViewHolder(View itemView) {
+    private Session session;
+    private Context context;
+    private OnBookmarkSelectedListener onBookmarkSelectedListener;
+    private SessionsListAdapter.OnItemClickListener onItemClickListener;
+
+    private static final int locationWiseSessionList = 1;
+    private static final int trackWiseSessionList = 4;
+    private static final int speakerWiseSessionList = 2;
+
+    // TODO : Move all unrelated logic out of view holder
+    SessionViewHolder(View itemView, Context context, OnBookmarkSelectedListener onBookmarkSelectedListener, SessionsListAdapter.OnItemClickListener onItemClickListener) {
         super(itemView);
         ButterKnife.bind(this, itemView);
+        this.context = context;
+        this.onBookmarkSelectedListener = onBookmarkSelectedListener;
+        this.onItemClickListener = onItemClickListener;
+    }
+
+    public void bindSession(int type, int colorInTracks, RealmDataRepository realmRepo) {
+
+        String sessionTitleString = Utils.checkStringEmpty(session.getTitle());
+        String sessionSubTitle = Utils.checkStringEmpty(session.getSubtitle());
+        int color = ContextCompat.getColor(context, R.color.color_primary);
+
+        sessionTitle.setText(sessionTitleString);
+
+        if (Utils.isEmpty(sessionSubTitle)) {
+            sessionSubtitle.setVisibility(View.GONE);
+        } else {
+            sessionSubtitle.setVisibility(View.VISIBLE);
+            sessionSubtitle.setText(sessionSubTitle);
+        }
+
+        sessionStatus.setVisibility(View.GONE);
+
+        setSessionStatus();
+
+        Track track = session.getTrack();
+
+        if (!RealmDataRepository.isNull(track)) {
+            int storedColor = Color.parseColor(track.getColor());
+
+            if (type != trackWiseSessionList) {
+                color = storedColor;
+            } else {
+                color = colorInTracks;
+            }
+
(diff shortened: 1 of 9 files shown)


android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
```diff
@@ -1,66 +1,287 @@
...
+    private static final int speakerWiseSessionList = 2;
+
+    public SessionViewHolder(View itemView, Context context, OnBookmarkSelectedListener onBookmarkSelectedListener) {
         super(itemView);
         ButterKnife.bind(this, itemView);
+        this.context = context;
+        this.onBookmarkSelectedListener = onBookmarkSelectedListener;
+    }
+
+    public void bindSession(int type, int color, RealmDataRepository realmRepo) {
```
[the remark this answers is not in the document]
srv-twry: Let's break this down into smaller pieces and call them from the method. It's way too long.

android/app/src/main/java/org/fossasia/openevent/adapters/SessionsListAdapter.java
```diff
@@ -100,7 +75,7 @@ public void setColor(int color) {
     public SessionViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
         LayoutInflater layoutInflater = LayoutInflater.from(parent.getContext());
         View view = layoutInflater.inflate(R.layout.tracksactvity_item, parent, false);
-        return new SessionViewHolder(view);
+        return new SessionViewHolder(view, context, onBookmarkSelectedListener);
```
iamareebjamal: view itself is a context

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
```diff
@@ -1,66 +1,307 @@
...
+
+            TextDrawable drawable = OpenEventApp.getTextDrawableBuilder().round()
+                    .build(String.valueOf(track.getName().charAt(0)), storedColor);
+
+            trackImageIcon.setImageDrawable(drawable);
+            trackImageIcon.setBackgroundColor(Color.TRANSPARENT);
+            sessionTrack.setText(track.getName());
+
+
+            itemView.setOnClickListener(v -> {
```
iamareebjamal: Best would be to transfer control back to fragment if an item is clicked

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
```diff
@@ -1,66 +1,307 @@
...
+            sessionBookmarkIcon.setImageResource(R.drawable.ic_bookmark_white_24dp);
+        } else {
+            sessionBookmarkIcon.setImageResource(R.drawable.ic_bookmark_border_white_24dp);
+        }
+        if (!Utils.isEmpty(color))
+            sessionBookmarkIcon.setColorFilter(Color.parseColor(color), PorterDuff.Mode.SRC_ATOP);
+    }
+
+    public void setSession(Session session) {
+        this.session = session;
```
iamareebjamal: All the above functions should be in ViewModel

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
mayank8318: setSession()? But this session is used in this ViewHolder only. I will have to pass the view model in the recylerview adapter then right?

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
mayank8318: Ok I will pass the session object through the interface method or should I pass the session properties that I need?

android/app/src/main/java/org/fossasia/openevent/adapters/viewholders/SessionViewHolder.java
iamareebjamal: I meant all above logic

mayank8318: Ok so that involves just the observable part right, cause everything else is just setting the UI or am I missing something here?

iamareebjamal: Yes the logic except UI should be moved

mayank8318: @iamareebjamal I made an interface in `SessionListAdapter` which returns this :
    public interface LinkViewModel {
        Observable<String> getSpeakerList();
    }

Since the adapter can't have a view model, should I put the code in all the view models in which the adapter is used or make a common helper method?

iamareebjamal: Why should this interface be needed? I think we are handling click in the adapter right, just create a listener in adapter which tells fragment which item was clicked and this information will be passed to viewmodel and it will handle the click

mayank8318: Currently we are handling clicks in the view holder itself, ie, we are setting the `onClickListener` while setting up the UI and directly handling clicks from there like launching new activity etc. Clicks I will handle in the fragment only like we normally do.

I made that interface to handle this :
image

I am stuck with this method.

mayank8318:
@iamareebjamal This part of the code is repeated in the view models. Where should I put this code?
image

iamareebjamal:
Views

mayank8318:

iamareebjamal:
@mayank8318 Please resolve conflicts

mayank8318:
@iamareebjamal Resolved!

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionsListAdapter.java
```diff
@@ -1,61 +1,43 @@
...
 import io.reactivex.disposables.CompositeDisposable;
-import timber.log.Timber;
-
-import static org.fossasia.openevent.core.bookmark.BookmarkStatus.Status.CODE_ERROR;
-import static org.fossasia.openevent.core.bookmark.BookmarkStatus.Status.CODE_UNDO_ADDED;
 
+/**
+ * User: MananWason
+ * Date: 26-06-2015
+ */
```
iamareebjamal: Why add author comments?

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionsListAdapter.java
```diff
@@ -1,61 +1,43 @@
...
+    private HandleItemClicks handleItemClicks;
 
     private RealmDataRepository realmRepo = RealmDataRepository.getDefaultInstance();
     private final CompositeDisposable compositeDisposable = new CompositeDisposable();
 
     private int color;
 
+    public interface HandleItemClicks {
+        void itemOnClick(int adapterPosition, int layoutPosition);
+    }
```
iamareebjamal: Actually, this is not helpful. Interface names should be adjectives, adverbs or nouns describing what implementer will be doing. 

Like BlahBlahBlahListener.

It can simply be `OnItemClickListener`

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionViewHolder.java
```diff
@@ -1,66 +1,282 @@
...
+
+        // Set color generated by palette on views
+        sessionHeader.setBackgroundColor(color);
+        if (track != null && track.isValid()) {
+            sessionTitle.setTextColor(Color.parseColor(track.getFontColor()));
+            setBookmarkIcon(sessionBookmarkIcon, session.getIsBookmarked(), track.getFontColor());
+        }
+    }
+
+    private void setObservable() {
```
iamareebjamal: The name is generic and unuseful

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionViewHolder.java
```diff
@@ -1,66 +1,282 @@
...
+                        String name = Utils.checkStringEmpty(speaker.getName());
+                        speakerName.add(name);
+                    }
+
+                    if (speakers.isEmpty()) {
+                        sessionSpeaker.setVisibility(View.GONE);
+                        speakerIcon.setVisibility(View.GONE);
+                    }
+
+                    return TextUtils.join(", ", speakerName);
```
iamareebjamal: Actually you are not using Observable here at all. All things are happening in serial. It's the same as writing it all imperatively with a for loop for Speakers, If you do want to use functional approach you may and it is better, but write it  completely

```Java
Observable.just(session.getSpeakers())
                .map(speakers ->
```

is just like 

```Java
List<Speaker> speakers = session.getSpeakers();
List<Speaker> yetAnotherSpeakers = speakers;
```

Write it like this:

```Java
Observable.fromIterable(session.getSpeakers())
    .map(speaker.getName())
    .map(Utils::checkStringEmpty) // Also, what is checkStringEmpty rename it to mean what it does
    .toList()
    .map(names -> TextUtils.join(", ", names))
    .subscribe(sessionSpeaker::setText);
```

And you are checking `if (speakers.isEmpty())` after creating n observable and iterating through speakers. Doesn't make sense. Add the check before and only do stuff if it is not empty

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionViewHolder.java
```diff
@@ -1,66 +1,282 @@
...
+                                onBookmarkSelectedListener.showSnackbar(new BookmarkStatus(-1, -1, CODE_ERROR));
+                        });
+
+                realmRepo.setBookmark(sessionId, true).subscribe();
+                setBookmarkIcon(
+                        sessionBookmarkIcon, true, track.getFontColor());
+            }
+            WidgetUpdater.updateWidget(context);
+        });
+    }
```
iamareebjamal, line 261: Add a todo to move all this unrelated to ViewHolder logic to somewhere else. A ViewHolder should only handle setting up UI

android/app/src/main/java/org/fossasia/openevent/core/track/session/TrackSessionsActivityViewModel.java
```diff
@@ -120,4 +123,16 @@ protected void onCleared() {
...
+        return listPosition;
+    }
+
+    public void setListPosition(int listPosition) {
+        this.listPosition = listPosition;
+    }
+
+    public Intent openSessionDetail(Session session, Intent intent) {
+        return Views.openSessionDetails(session, intent);
+    }
```
iamareebjamal: I don't think there is any need of storing and retrieving just listPosition here. And listPosition is not stable. It can change if items are sorted or filtered. Also, there is no need to put openSessionDetail in ViewModel, it can be in the View itself

android/app/src/main/java/org/fossasia/openevent/core/track/session/TrackSessionsActivity.java
```diff
@@ -358,4 +361,11 @@ public void showSnackbar(BookmarkStatus bookmarkStatus) {
         SnackbarUtil.setSnackbarAction(this, snackbar, bookmarkStatus)
                 .show();
     }
+
+    @Override
+    public void itemOnClick(int adapterPosition, int layoutPosition) {
+        trackSessionsActivityViewModel.setListPosition(layoutPosition);
+        Intent intent = new Intent(this, SessionDetailActivity.class);
+        startActivity(trackSessionsActivityViewModel.openSessionDetail(sessions.get(adapterPosition), intent));
+    }
```
iamareebjamal: Unfortunately, even after moving the click handling in the view, we still don't have any less complex ViewHolder. Still, it is progress enough. But don't pass adapterPosition and layoutPosition here. Pass the session which was clicked

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/track/session/TrackSessionsActivity.java
mayank8318: Most of the part is just setting views depending on the value, so we will have to make some helper class to handle the if-else part then

mayank8318:

iamareebjamal (approved):

PR #2238, opened 2018-01-26
fix: Replaced string concatenation with placeholders

### LINKED ISSUE #2204: Replace String concatenation with placeholders in the project
**Actual Behaviour**

There are a lot of instances of string concatenation inside the setText argument.

**Expected Behaviour**

These can and should be replaced by Placeholders in String resources.

Yes

dr0pdb:
Fixes #2204 

Changes: Replaced hardcoded string concatenation with placeholders in string resources.

@iamareebjamal Please review.


android/app/src/main/res/values/strings.xml
```diff
@@ -318,4 +318,11 @@
     <string name="status_upcoming">UPCOMING</string>
     <string name="status_ongoing">ONGOING</string>
 
+    <!--Strings with placeholders -->
+    <string name="filters_text">Filters(%d): %s</string>
+    <string name="about_fragment_license_text">%s %s</string>
```
iamareebjamal: Just space is not a great template as it can be reused almost anywhere else. Either remove completely or rename tot general use

iamareebjamal (changes requested):

dr0pdb:
Changes: Renamed to ```space_separated_strings```.

@iamareebjamal Please review.

iamareebjamal (approved):

Issue #2098, opened 2018-01-06
Make images zoomable throughout the application


dr0pdb:
**Actual Behaviour**

Currently only the images inside the FeedFragment are zoomable.

**Expected Behaviour**

All the images should be zoomable.

**Would you like to work on the issue?**

Yes

iamareebjamal:
@srv-twry Should this be closed?

dr0pdb:
I'll try and search for images that aren't zoomable yet. If not found, then i will close it.

simarsingh24:
@srv-twry the profile photo inside "My Profile" should also be made zoomable! what do you think?

dr0pdb:
Yes, should i work on it or do you want to work on it ?

simarsingh24:
@srv-twry yeah, I would like to work on it :)

PR #2268, opened 2018-01-28
made profile photo zoomable

simarsingh24:
Parent Issue #2098 
changes: made profile photo zoomable
Screenshots after the change : 
image


simarsingh24:
@srv-twry I did it the way you implemented it inside speakersDetailsActivity, but it doesn't look that nice here also it is taking too long to load, what should we do?

dr0pdb:
Yes, it does look bad. I am not sure why this is the case. Let's wait for @iamareebjamal to suggest something.
Also i think it's not that big a use case to make the profile image zoomable. I know the two of us discussed to make it zoomable, but it's certainly not standard or important either, specially in this app. Realised it late.

PR #2283, opened 2018-01-30
fix: download data from JSON when using the app for the first time and back button is pressed         

### LINKED ISSUE #2281: Download dialog shouldn't be cancelable when using the app for the first time
**Actual Behaviour**

Download dialog is cancelable even when using the app for the first time.
If you cancel the download dialog when using the app for the first(by pressing the back button)
You are left with an empty app which crashes at various places.

**Expected Behaviour**

Download dialog shouldn't be cancelable when using the app for the first time so that the user will either download the latest data or the one from the JSON  

**Steps to reproduce it**
Clear data. Open the app. Press the back button. Explore the empty app

**Screenshots of the issue**

screenshot_20180131-011450

Yes

nikit19:
Fixes #2281 

Changes: download dialog is not cancelable when using the app for the first time

Screenshots for the change: 
ezgif com-video-to-gif 35

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/main/MainActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/main/MainActivity.java
@@ -113,7 +113,7 @@ import io.realm.RealmList;
 import io.realm.RealmResults;
 import timber.log.Timber;
 
-public class MainActivity extends BaseActivity implements FeedAdapter.OpenCommentsDialogListener, OnImageZoomListener, AboutFragment.OnMapSelectedListener {
+public class MainActivity extends BaseActivity implements FeedAdapter.OpenCommentsDialogListener, OnImageZoomListener, AboutFragment.OnMapSelectedListener, DialogFactory.DialogListener {
 
     private static final String STATE_FRAGMENT = "stateFragment";
     private static final String NAV_ITEM = "navItem";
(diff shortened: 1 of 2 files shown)


iamareebjamal:
The default behaviour should be to parse JSON if cancelled

nikit19:
When the user uses the app for the first time and presses the back button ,data is downloaded from the JSON
ezgif com-video-to-gif 36

@iamareebjamal Please review

nikit19:
conflicts resolved

android/app/src/main/java/org/fossasia/openevent/core/main/MainActivity.java
```diff
@@ -895,4 +897,10 @@ public void onMapSelected(boolean value) {
     public void onZoom(String imageUri) {
         ZoomableImageUtil.showZoomableImageDialogFragment(fragmentManager, imageUri);
     }
+
+    @Override
+    public void onDialogBackPress() {
+        //it used to call the method downloadFromAssets() if the user is using the app for
+        //the first time and presses the back button in the download dialog
+    }
```
iamareebjamal, line 905: Why is it empty though?

nikit19: This was overriden when onDialogClickListener was initialized using lamdas

Issue #2280, opened 2018-01-30
Cluttering of 'No Results Found' text and Map fragment together.


dreadpool2:
**Actual Behaviour**

In LocationActivity, when there are no results found for a query and then the map icon is clicked, the map fragment formed clutters with the 'No Results Found' text. Even the results clutter with the map being in behind.

**Expected Behaviour**

The cluttering should not occur.

**Screenshots of the issue**

![screenshot_20180131-010129](https://user-images.githubusercontent.com/28607714/35587025-7a1dbfd2-0622-11e8-9d14-8728a3b09425.png)

![screenshot_20180131-010156](https://user-images.githubusercontent.com/28607714/35587009-6f51ba40-0622-11e8-90bc-66d8fd51237c.png)

**Would you like to work on the issue?**

Yes

PR #2285, opened 2018-01-30
fix : clicking on map aborted while searching in LocationsActivity

dreadpool2:
First steps for issues #2280 

Changes: The clicking on map icon has been simply aborted while searching in `LocationActivity` as it is quite irrelevant. Thus the cluttering wont even occur.


dreadpool2:
@iamareebjamal @srv-twry Is it ok to follow this procedure?

Issue #2249, opened 2018-01-27
Adding "Ticket System" to the app


simarsingh24:
**Current Behaviour**
The API is returning ticket related information with the event which is currently unutilised by the App

**Expected Behaviour**

The App should exploit ticket related info.

**Would you like to work on the issue?**

Yes

@iamareebjamal  what are your views?

iamareebjamal:
What information would you want to show?

simarsingh24:
This is a parent issue various thing that can be added like "Tickets available under an event", "Tickets purchased by user", "attendees" etc.
I wanted to start with "Tickets purchased by the user" but I don't know if this is there in the API (getting ticket purchased using User ID) or we may start by "Attendees" of the selected event

iamareebjamal:
Why should we show attendees of an event to the user? Also, the user does not even have the permission to view this information

simarsingh24:
the count of attendees, just to give him an idea how many people are interested in that particular event

iamareebjamal:
You can see if the API allows that

simarsingh24:
@iamareebjamal I was thinking of adding tickets under an event , but the problem is that the API is returning empty response.
I tried changing the events but i am getting empty response, what should we do?
<img width="680" alt="screen shot 2018-02-13 at 11 55 05 pm" src="https://user-images.githubusercontent.com/16715499/36166735-77ebb69a-1119-11e8-934d-70cd96f02cfd.png">

iamareebjamal:
I think complete ticketing is a complex feature and not ready to be implemented in the app. It requires handling money as well if we want to be complete, so let's handle this later

PR #2291, opened 2018-01-31
adding ticket related info to event

simarsingh24:
Mention #2249


simarsingh24:
@srv-twry  @mayank8318 I am getting an empty response don't know why? can you please go through the changes and let me know what wrong am I doing, and is this the right way to do it? 
PS : please ignore formatting errors I will remove those later.

android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.java
```diff
@@ -495,6 +496,27 @@ public Completable saveSponsors(final List<Sponsor> sponsors) {
         return realm.where(Sponsor.class).findAllSortedAsync("level", Sort.DESCENDING, "name", Sort.ASCENDING);
     }
 
+    // Tickets Section
+
+    private void saveTicketsInRealm(List<Ticket> tickets) {
+        Realm realm = Realm.getDefaultInstance();
```
dr0pdb, line 502: How about logging and checking if the tickets array is empty or not.

android/app/src/main/java/org/fossasia/openevent/common/api/processor/TicketResponseProcessor.java
```diff
@@ -0,0 +1,27 @@
...
+import org.fossasia.openevent.data.Ticket;
+import org.fossasia.openevent.data.repository.RealmDataRepository;
+
+import java.util.List;
+
+public class TicketResponseProcessor extends ResponseProcessor<List<Ticket>> {
+
+    @Override
+    protected void onSuccess(List<Ticket> ticket) {
+        complete(RealmDataRepository.getDefaultInstance()
```
dr0pdb, line 14: Log it and check if the ticket list is actually complete or empty.

dr0pdb (commented):
It's difficult to say from here, but how about you log and check where the tickets list is empty. 
If it is empty right from start, then the API call is returning empty, then i would recommend you try and make an api call using some other tool and see what response you get.

simarsingh24:
@srv-twry true! I was actually using postman software for viewing the received response for this get request, I think the API is returning an empty response. I will check once again and discuss with you!

PR #2302, opened 2018-02-01
feat: Add Featured speakers in AboutFragment

### LINKED ISSUE #2301: Add featured speakers in the home screen.
**Expected Behaviour**

The front end shows featured speakers separately, hence it should be shown in the app also.

Yes.

**Parent issue**

#1708

dr0pdb:
Fixes #2301 

Changes: Added Featured speakers in AboutFragment.

Screenshots for the change: 

screenshot_2018-02-01-22-00-29-82
screenshot_2018-02-01-22-00-33-95

@iamareebjamal Please review.


android/app/src/main/res/layout/fragment_about.xml
```diff
@@ -19,7 +12,8 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:background="@color/white"
-            android:orientation="vertical">
+            android:orientation="vertical"
+            android:descendantFocusability="blocksDescendants">
```
dr0pdb, line 16: In case you're wondering about the use of this line.
If i don't add this line, then the home screen starts at the middle of the speakers recyclerview instead of being at the top of the layout.
[Here](https://stackoverflow.com/a/40474022/6748052) is the stackoverflow answer which suggests adding this.

dr0pdb (commented):
As it turns out, the last merged PR about fixing the social links position at the bottom was redundant and it had to be removed here.

iamareebjamal:
@srv-twry Don't show staggered layout on home screen, show speakers like they are shown in session detail activity

dr0pdb:
@iamareebjamal Actually in the parent issue, you have asked to use the existing view holder of Speakers Activity.
Anyways, i would only show the image, name and designation as shown in the SessionDetailsActivity.

iamareebjamal:
Yes, the existing view holder can be used from SessionDetailsActivity

dr0pdb:
screenshot_2018-02-02-00-43-02-16

screenshot_2018-02-02-00-45-27-85



**NOTE**: There is already an issue created and an active PR to remove the problem with the speaker's designation. Hence i haven't changed anything in this PR.

iamareebjamal (approved):

PR #2314, opened 2018-02-04
Fix #2284 - Add FAQs section to the app.

### LINKED ISSUE #2284: Add FAQs section
**Actual Behaviour**

There is no FAQ section in the app.

**Expected Behaviour**

There should be one.

Yes

@iamareebjamal Should we add this?

mayank8318:
Fixes #2284 

Changes:
Step 1) Create the data model.
@iamareebjamal Please review if I am on the right track here. Also, I checked the sample data for the event here - https://github.com/fossasia/open-event/tree/master/sample but I didn't find any json for the FAQs. How should I add it to the `assets` for offline use? Should I create my own sample FAQs?

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/common/api/APIClient.java
+++ b/android/app/src/main/java/org/fossasia/openevent/common/api/APIClient.java
@@ -8,6 +8,7 @@ import com.github.jasminb.jsonapi.retrofit.JSONAPIConverterFactory;
 import org.fossasia.openevent.BuildConfig;
 import org.fossasia.openevent.OpenEventApp;
 import org.fossasia.openevent.data.Event;
+import org.fossasia.openevent.data.FAQ;
 import org.fossasia.openevent.data.Microlocation;
 import org.fossasia.openevent.data.Session;
 import org.fossasia.openevent.data.Speaker;
@@ -87,7 +88,7 @@ public final class APIClient {
                     .build();
 
             ObjectMapper objectMapper = getObjectMapper();
-            Class[] classes = {Event.class, Track.class, Speaker.class, Sponsor.class, Session.class, Microlocation.class, User.class};
+            Class[] classes = {Event.class, Track.class, Speaker.class, Sponsor.class, Session.class, Microlocation.class, User.class, FAQ.class};
 
             openEventAPI = new Retrofit.Builder()
                     .client(okHttpClient)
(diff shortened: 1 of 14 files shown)


iamareebjamal:
No, if not present, don't show any FAQ section

mayank8318:
@iamareebjamal But it is documented in the API docs?

iamareebjamal:
But if only JSON is available, we can't show FAQ

mayank8318:
Okay

mayank8318:
Changes: Complete the FAQ functionality.

@iamareebjamal This is the response from Postman:

{
    "meta": {
        "count": 0
    },
    "links": {
        "self": "/v1/events/173/faqs"
    },
    "data": [],
    "jsonapi": {
        "version": "1.0"
    }
}

And this is the app log :
open1

android/app/src/main/java/org/fossasia/openevent/common/events/FAQDownloadEvent.java
```diff
@@ -0,0 +1,13 @@
+package org.fossasia.openevent.common.events;
+
+/**
+ * Created by mayank on 4/2/18.
+ */
+
+public class FAQDownloadEvent extends DownloadEvent {
```
iamareebjamal: Please remove author comment

iamareebjamal (changes requested):
Similarly from other areas

android/app/src/main/java/org/fossasia/openevent/common/events/FAQDownloadEvent.java
mayank8318: Okay

mayank8318:
@iamareebjamal I made a new event id **678** and added a faq using the api. I checked with postman and I am getting the results on querying. Also, I checked it by changing the api link in the config file and I am getting correct results.

mayank8318:
Screenshots for the added FAQs :
image

iamareebjamal:
Improve UI a little bit. Use cards, don't use colors too much. And make it so the cards are collapsed by default and expand when clicked. Please add a screenshot of what happens where there is no FAQ

mayank8318:
Changes :
Made the cards collapsible.
27420260_1967770233251781_8946862106762805248_n

- FAQs need the user to be logged in. So I am opening login screen if the user is not logged in.
-  If the user presses back, appropriate message is shown
image

Empty FAQ screen :
image

iamareebjamal:
Good job. Just change the radius of cards to be 2dp instead

mayank8318:
@iamareebjamal Ok will do in an hour :)

mayank8318:
@iamareebjamal made the changes.

iamareebjamal (approved):

PR #2328, opened 2018-02-10
feat : Viewmodel for DayScheduleFragment

### LINKED ISSUE #2031: Create ViewModel for the DayScheduleFragment
**Actual Behaviour**

There is no ViewModel for the DayScheduleFragment

**Expected Behaviour**

There should be one.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes!

NA

**Other information**

Parent issue #2015

leeyh20:
Fixes #2031

Changes: Adds ViewModel for DayScheduleFragment

Screenshots for the change: None

@srv-twry Do tell me if I missed out anything to shift to the ViewModel! Additionally, what is an effective way to test if it is working well? Thank you!

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
@@ -1,5 +1,6 @@
 package org.fossasia.openevent.core.schedule;
 
+import android.arch.lifecycle.ViewModelProviders;
 import android.content.Context;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
@@ -29,27 +30,21 @@ import org.fossasia.openevent.common.network.NetworkUtils;
 import org.fossasia.openevent.common.ui.Views;
 import org.fossasia.openevent.common.ui.base.BaseFragment;
 import org.fossasia.openevent.common.ui.recyclerview.stickyheadersrecyclerview.StickyRecyclerHeadersDecoration;
-import org.fossasia.openevent.common.utils.SortOrder;
 import org.fossasia.openevent.common.utils.Utils;
 import org.fossasia.openevent.config.StrategyRegistry;
 import org.fossasia.openevent.core.bookmark.OnBookmarkSelectedListener;
 import org.fossasia.openevent.data.Session;
-import org.fossasia.openevent.data.repository.RealmDataRepository;
-
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 import butterknife.BindView;
-import io.realm.RealmResults;
 import timber.log.Timber;
 
 public class DayScheduleFragment extends BaseFragment implements SearchView.OnQueryTextListener {
 
-    final private String SEARCH = "searchText";
-
     private Context context;
+    private String date;
     private String searchText = "";
     private SearchView searchView;
 
@@ -58,15 +53,12 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
     @BindView(R.id.txt_no_schedule) TextView noSchedule;
     @BindView(R.id.txt_no_result_schedule) protected TextView noResultsSchedule;
 
-    private List<Session> sessions = new ArrayList<>();
     private List<Session> filteredSessions = new ArrayList<>();
     private DayScheduleAdapter dayScheduleAdapter;
-    private OnBookmarkSelectedListener onBookmarkSelectedListener;
 
-    private String date;
-    private RealmDataRepository realmRepo = RealmDataRepository.getDefaultInstance();
-    private RealmResults<Session> realmResults;
+    private OnBookmarkSelectedListener onBookmarkSelectedListener;
     private RecyclerView.AdapterDataObserver adapterDataObserver;
+    private DayScheduleFragmentViewModel dayScheduleFragmentViewModel;
 
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
@@ -85,32 +77,22 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
         Utils.registerIfUrlValid(swipeRefreshLayout, this, this::refresh);
         setUpRecyclerView();
 
-        if (savedInstanceState != null && savedInstanceState.getString(SEARCH) != null) {
-            searchText = savedInstanceState.getString(SEARCH);
-        }
+        // Set up view model
+        dayScheduleFragmentViewModel = ViewModelProviders.of(this).get(DayScheduleFragmentViewModel.class);
+        searchText = dayScheduleFragmentViewModel.getSearchText();
 
-        realmResults = realmRepo.getSessionsByDate(date, SortOrder.sortTypeSchedule());
-        realmResults.addChangeListener((sortedSessions, orderedCollectionChangeSet) -> {
-            sessions.clear();
-            sessions.addAll(sortedSessions);
+        loadSessions();
+        handleVisibility();
+        return view;
+    }
 
+    private void loadSessions() {
+        dayScheduleFragmentViewModel.getSessionsByDate(date, searchText).observe(DayScheduleFragment.this, sessions ->  {
             filteredSessions.clear();
-            filteredSessions.addAll(sortedSessions);
-
-            dayScheduleAdapter.setCopy(sortedSessions);
+            filteredSessions.addAll(sessions);
             dayScheduleAdapter.notifyDataSetChanged();
-            if (!Utils.isEmpty(searchText))
-                dayScheduleAdapter.filter(searchText);
-
-            if(SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING)
-                Collections.reverse(filteredSessions);
-
             handleVisibility();
         });
-
-        handleVisibility();
-
-        return view;
     }
 
     private void setUpRecyclerView() {
@@ -121,7 +103,6 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
         dayRecyclerView.setLayoutManager(new LinearLayoutManager(context, LinearLayoutManager.VERTICAL, false));
         dayRecyclerView.addItemDecoration(new DividerItemDecoration(context, DividerItemDecoration.VERTICAL));
         dayRecyclerView.setAdapter(dayScheduleAdapter);
-        dayScheduleAdapter.setEventDate(date);
 
         final StickyRecyclerHeadersDecoration headersDecoration = new StickyRecyclerHeadersDecoration(dayScheduleAdapter);
         dayRecyclerView.addItemDecoration(headersDecoration);
@@ -134,38 +115,25 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
         dayScheduleAdapter.registerAdapterDataObserver(adapterDataObserver);
     }
 
-    public void filter(List<String> selectedTracks) {
-        if(dayScheduleAdapter == null)
-            return;
-
-        realmRepo.getSessionsByDate(date, SortOrder.sortTypeSchedule())
-                .addChangeListener((sortedSessions, orderedCollectionChangeSet) -> {
-                    sessions.clear();
-                    sessions.addAll(sortedSessions);
-
-                    filteredSessions.clear();
-                    if (selectedTracks.isEmpty()) {
-                        filteredSessions.addAll(sessions);
-                    } else {
-                        for(int i = 0; i < sessions.size() ; i++) {
-                            String trackName = sessions.get(i).getTrack().getName();
-                            if (selectedTracks.contains(trackName)) {
-                                filteredSessions.add(sessions.get(i));
-                            }
-                        }
-
-                        if(searchText!=null)
-                            dayScheduleAdapter.filter(searchText);
-                    }
-
-                    if(SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING)
-                        Collections.reverse(filteredSessions);
-
-                    dayScheduleAdapter.notifyDataSetChanged();
-
-                    handleVisibility();
-                });
+    public void filterByTracks(List<String> selectedTracks) {
+        dayScheduleFragmentViewModel.getSortedSessionsByTracks(searchText, selectedTracks).observe(DayScheduleFragment.this, trackSessions -> {
+            filteredSessions.clear();
+            filteredSessions.addAll(trackSessions);
+            dayScheduleAdapter.notifyDataSetChanged();
+            handleVisibility();
+        });
     }
+
+    public void filterByQuery(String query) {
+        searchText = query;
+        dayScheduleFragmentViewModel.getSessionsBySearchText(searchText).observe(DayScheduleFragment.this, matchingSessions -> {
+            filteredSessions.clear();
+            filteredSessions.addAll(matchingSessions);
+            dayScheduleAdapter.notifyDataSetChanged();
+            handleVisibility();
+        });
+    }
+
 
     private void handleVisibility() {
         if (dayRecyclerView != null && noSchedule != null) {
@@ -176,7 +144,6 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
             }
         }
     }
-
 
     @Override
     protected int getLayoutResource() {
@@ -190,7 +157,6 @@ public class DayScheduleFragment extends BaseFragment implements SearchView.OnQu
         dayScheduleAdapter.unregisterAdapterDataObserver(adapterDataObserver);
 
         // Remove listeners to fix memory leak
-        realmResults.removeAllChangeListeners();
         if(swipeRefreshLayout != null) swipeRefreshLayout.setOnRefreshListener(null);
         if(searchView != null) searchView.setOnQueryTextListener(null);
     }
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -85,32 +81,29 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
 
-            dayScheduleAdapter.setCopy(sortedSessions);
+            dayScheduleAdapter.setCopy(sessions);
             dayScheduleAdapter.notifyDataSetChanged();
-            if (!Utils.isEmpty(searchText))
-                dayScheduleAdapter.filter(searchText);
 
-            if(SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING)
+            // Should this be done in the view model instead?
+            if(SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING) {
```
dr0pdb: Yes, Please do it in the ViewModel and return data accordingly.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -135,36 +128,20 @@ public void onChanged() {
     }
 
     public void filter(List<String> selectedTracks) {
-        if(dayScheduleAdapter == null)
+        if (dayScheduleAdapter == null)
```
dr0pdb: Please remove this check. The adapter can't be null IMO.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -135,36 +128,20 @@ public void onChanged() {
...
-                    dayScheduleAdapter.notifyDataSetChanged();
-
-                    handleVisibility();
-                });
+        dayScheduleFragmentViewModel.getSessionsByTracks(date, selectedTracks).observe(DayScheduleFragment.this, trackSessions -> {
+            filteredSessions.clear();
+            filteredSessions.addAll(trackSessions);
+            if (searchText != null)
+                dayScheduleAdapter.filter(searchText);
+            if (SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING)
```
dr0pdb: Return reversed data from the ViewModel itself.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -190,7 +166,7 @@ public void onDestroyView() {
         dayScheduleAdapter.unregisterAdapterDataObserver(adapterDataObserver);
 
         // Remove listeners to fix memory leak
-        realmResults.removeAllChangeListeners();
+        //realmResults.removeAllChangeListeners();
```
leeyh20: Does this need to be removed?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -135,36 +128,20 @@ public void onChanged() {
...
-                        Collections.reverse(filteredSessions);
-
-                    dayScheduleAdapter.notifyDataSetChanged();
-
-                    handleVisibility();
-                });
+        dayScheduleFragmentViewModel.getSessionsByTracks(date, selectedTracks).observe(DayScheduleFragment.this, trackSessions -> {
+            filteredSessions.clear();
+            filteredSessions.addAll(trackSessions);
+            if (searchText != null)
```
dr0pdb: No need for this check.

dr0pdb: No need for this check

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
dr0pdb: Why not delete this ?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -135,36 +128,20 @@ public void onChanged() {
...
+                dayScheduleAdapter.filter(searchText);
```
leeyh20: Should this filtering be placed in viewmodel or fragment?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -214,8 +190,8 @@ public void onSessionsDownloadDone(SessionDownloadEvent event) {
 
     @Override
     public void onSaveInstanceState(@NonNull Bundle bundle) {
-        if (isAdded() && searchView != null) {
-            bundle.putString(SEARCH, searchText);
+        if (isAdded() && searchView != null && dayScheduleFragmentViewModel != null) {
```
dr0pdb: the ViewModel will never be null IMO.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
dr0pdb: Yes, please do it ViewModel. The aim is to make the view as dumb as possible and hence all the logic to ViewModel if possible.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -85,32 +81,29 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+            filteredSessions.addAll(sessions);
 
-            dayScheduleAdapter.setCopy(sortedSessions);
+            dayScheduleAdapter.setCopy(sessions);
             dayScheduleAdapter.notifyDataSetChanged();
-            if (!Utils.isEmpty(searchText))
-                dayScheduleAdapter.filter(searchText);
 
-            if(SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING)
+            // Should this be done in the view model instead?
```
leeyh20: Should this be done in viewmodel?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
leeyh20: Ok thanks for the help :)!

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,88 @@
...
+            mutableFilterSessions.filter(predicate);
+        }
+    }
+
+    private void sortSessions() {
+        List<Session> sessionList = mutableFilterSessions.getValue();
+
+        // Sort sessions according to global sortTypeSchedule
+        if (SortOrder.sortOrderSchedule() == SortOrder.SORT_ORDER_DESCENDING) {
+            sessionList = ((RealmResults<Session>) sessionList).sort(SortOrder.sortTypeSchedule(), Sort.DESCENDING);
```
leeyh20: @srv-twry The only issue now is that the app crashes when filtering with tracks. The error is at line 76, java.util.ArrayList sessionList cannot be cast to io.realm.RealmResults. Is there a way of sorting by SortOrder.sortTypeSchedule() for the List<Session>? Thanks!

mayank8318: @leeyh20 Yes you cannot cast ArrayList to RealmResults. This works fine until `filterWithTracks` is executed. In this method, this line -> `mutableFilterSessions.filter(predicate);` changes the type of value stored in mutableFilterSessions from **RealmResults** to **ArrayList** and hence when you execute `mutableFilterSessions.getValue()` after this, you get a return type of ArrayList.
Check the super class method which you are using here.

leeyh20: @mayank8318 Thanks! I’ll try to solve this.

leeyh20:
@srv-twry @mayank8318 It works now!

mayank8318 (approved):

leeyh20:
@iamareebjamal Do you have any advice for this PR? Thank you!

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -215,7 +177,7 @@ public void onSessionsDownloadDone(SessionDownloadEvent event) {
     @Override
     public void onSaveInstanceState(@NonNull Bundle bundle) {
         if (isAdded() && searchView != null) {
-            bundle.putString(SEARCH, searchText);
+            dayScheduleFragmentViewModel.setSearchText(searchText);
```
dr0pdb: I think it's redundant. Everytime the searchText is changed, we already save it inside the ViewModel, so it always has the updated searchText value. 
You can go ahead and delete this whole redundant method.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.reactivex.functions.Predicate;
+import io.realm.RealmResults;
+import io.realm.Sort;
+
+public class DayScheduleFragmentViewModel extends ViewModel {
+
+    private RealmDataRepository realmRepo;
+    private FilterableRealmLiveData<Session> filterableSessions;
+
```
dr0pdb: Please remove this line also. I understand you tried to show a change in context but i think it's unnecessary since the variables are only three.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -85,32 +81,24 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+    }
 
+    private void loadSessionsByDate() {
+        dayScheduleFragmentViewModel.getSessionsByDate(date, searchText).observe(DayScheduleFragment.this, sessions ->  {
             filteredSessions.clear();
-            filteredSessions.addAll(sortedSessions);
+            filteredSessions.addAll(sessions);
 
-            dayScheduleAdapter.setCopy(sortedSessions);
+            dayScheduleAdapter.setCopy(sessions);
```
dr0pdb: Can you please check for the usage of this copy inside the adapter ? I think it was earlier used for filtering and may be won't be needed anymore after we have moved it to the VM. Please check it once.

dr0pdb (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
leeyh20: I have now changed the code to not use dayScheduleAdapter filtering anymore and removed the setCopy in dayScheduleAdapter.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
leeyh20: Yup it does seem to be redundant. I have noticed that LocationsFragment has a very similar method also called onSaveInstanceState though. Is it also redundant there since LocationsFragment also has a ViewModel now?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -134,39 +115,26 @@ public void onChanged() {
...
+            handleVisibility();
+        });
+    }
+
+    public void filterByQuery(String query) {
+        searchText = query;
+        dayScheduleFragmentViewModel.getSessionsBySearchText(searchText).observe(DayScheduleFragment.this, matchingSessions -> {
+            filteredSessions.clear();
+            filteredSessions.addAll(matchingSessions);
+            dayScheduleAdapter.notifyDataSetChanged();
```
leeyh20, line 132: @srv-twry I tried to use dayScheduleAdapter.animateTo(matchingSessions) for this but it lagged when I started typing in the search query. I'm not sure why the animateTo did not work well for this.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
```diff
@@ -238,10 +196,8 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
     @Override
     public boolean onQueryTextChange(String query) {
         searchText = query;
-
-        dayScheduleAdapter.filter(searchText);
+        filterByQuery(searchText);
```
leeyh20, line 199: @srv-twry Instead of dayScheduleAdapter.filter function, I created another function filterByQuery to use the ViewModel instead, will it be ok?

leeyh20: Now there is filterByQuery and filterByTracks.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragment.java
leeyh20: Its okay now, I now use animateTo in onQueryTextChange instead.

leeyh20:
Should I squash my commits?

dr0pdb:
@leeyh20 Yes please.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+    public DayScheduleFragmentViewModel() {
+        realmRepo = RealmDataRepository.getDefaultInstance();
+    }
+
+    public String getSearchText() { return searchText; }
+
+    public LiveData<List<Session>> getSessionsByDate(String date, String searchText) {
+        // Get entirely new data from realm
+        sessions = loadSessions(date);
+        // Pass an empty list for selectedTracks
```
dr0pdb: You can actually remove these comments. It's pretty clear what we're doing here. Also the method is extremely short, hence it's very easy to understand what we're doing anyways.

dr0pdb (changes requested):
LGTM :+1: , just remove those comments.

leeyh20:
@srv-twry Sorry, I squashed the commits wrongly and caused the PR to close unexpectedly. However, I have now restored the PR with the previous changes!

dr0pdb:
@leeyh20 LGTM, Please attach a GIF so that we can be sure about this.

leeyh20:
@srv-twry  Filtering by tracks and then searching and then cancelling the filter:
tracks

Sorting and then clicking "Ascending":
sorting

dr0pdb (approved):
@iamareebjamal Please go through it to make sure i haven't missed anything. LGTM.

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragmentViewModel.java
```diff
@@ -0,0 +1,78 @@
...
+        this.tracks = selectedTracks;
+        final String query = this.searchText.toLowerCase(Locale.getDefault());
+
+        Predicate<Session> predicate = session -> {
+                boolean hasQuery = session.getTitle()
+                        .toLowerCase(Locale.getDefault())
+                        .contains(query);
+
+                boolean isInTracks = true;
+                if (!this.tracks.isEmpty()) {
```
iamareebjamal, line 67: This check can be removed. Should `isInTracks` be true if tracks are empty?

android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.java
```diff
@@ -384,8 +384,8 @@ public Session getSession(String title) {
         return realm.where(Session.class).equalTo("microlocation.name", location).findAllSortedAsync(Session.START_TIME);
     }
 
-    public RealmResults<Session> getSessionsByDate(String date, String sortCriteria) {
-        return realm.where(Session.class).equalTo("startDate", date).findAllSortedAsync(sortCriteria);
```
iamareebjamal, line 388: Is there a reason to change this asynchronous sorting?

android/app/src/main/java/org/fossasia/openevent/core/schedule/DayScheduleFragmentViewModel.java
leeyh20: @iamareebjamal I think that `isInTracks` need to be true if tracks are empty because if there are no tracks selected in the tracks filter, then we want to show all sessions. If not, no sessions will be shown when tracks filter is not used. 

Or should I refactor this such that we don't check` isInTracks `at all if `this.tracks` is empty? If `this.tracks` is empty, we only want to filter by the searchText.

android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.java
leeyh20: @iamareebjamal The only reason is because this function is only used by DayScheduleFragment and that sorting is already done separately in another function in DayScheduleFragment. Sorting is repeated everytime `getSortedSessionsByTracks` is called because it is needed for the sorting that we specify using the UI (Ascending, descending, by tracks, by start time etc)  Thus when we retrieve the data from Realm, the sessions do not need to be sorted. 

We can also leave this to return sorted sessions though, if it is potentially used by other fragments/activities. I don't think it'll make a difference for DayScheduleFragment other than efficiency.

iamareebjamal (approved):
Good job. Thorough and good follow-ups with efficient communication

leeyh20:
Thank you, I will try to contribute more!

PR #2329, opened 2018-02-10
Fixes #2318 Events bookmarked from Featured Speakers are now visible in AboutFragment.

### LINKED ISSUE #2318: Events bookmarked from Featured Speakers are not visible in about fragment.
**Actual Behaviour**

The events bookmarked from the section of Featured Speakers are not visible in about fragment.

**Expected Behaviour**

The events should be visible.

**Steps to reproduce it**

See the below Gif to reproduce the issue.

**Gif of the issue**

open

Yes

raj10071997:
Fixes #2318

Changes: Before when the onResume() method was called in AboutFragment.java, it doesn't update the bookMarksListAdapter with the new data. Now call to the function getBookmarkedSessions() first checks the new data and then update the list for the bookMarksListAdapter.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragmentViewModel.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragmentViewModel.java
@@ -34,6 +34,7 @@ public class AboutFragmentViewModel extends ViewModel {
     private LiveData<Event> eventLiveData;
     private LiveData<List<Speaker>> featuredSpeakers;
     private MutableLiveData<Bitmap> eventLogo;
+    private Long numberOfBookmarkedSessions;
 
     public AboutFragmentViewModel() {
         realmRepo = RealmDataRepository.getDefaultInstance();
@@ -58,10 +59,12 @@ public class AboutFragmentViewModel extends ViewModel {
     }
 
     public LiveData<List<Object>> getBookmarkedSessions() {
-        if (sessions == null) {
+        if (sessions == null || numberOfBookmarkedSessions != realmRepo.getBookMarkedSessionsLength()) {
             LiveRealmData<Session> sessionLiveRealmData = RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions());
             sessions = Transformations.map(sessionLiveRealmData, this::getSessionsList);
+            numberOfBookmarkedSessions = realmRepo.getBookMarkedSessionsLength();
         }
+
         return sessions;
     }
 
--- a/android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.java
+++ b/android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.java
@@ -406,6 +406,10 @@ public class RealmDataRepository {
         return realm.where(Session.class).equalTo("isBookmarked", true).findAllAsync();
     }
 
+    public long getBookMarkedSessionsLength() {
+        return realm.where(Session.class).equalTo("isBookmarked", true).count();
+    }
+
     public List<Session> getBookMarkedSessionsSync() {
         Realm realm = Realm.getDefaultInstance();
         RealmResults<Session> sessions = realm.where(Session.class).equalTo("isBookmarked", true).findAll();


iamareebjamal:
IMO, this is a workaround and not a solution

raj10071997:
@iamareebjamal  Actually I thought of this solution because when we come back from SpeakerDetailsActivity.java, the onResume() method of AboutFragment.java is called which in turn calls the getBookmarkedSessions(). Now this method returns the previous list of sessions and not the updated one, so there must be a way to update it and that's why I thought of this solution.

raj10071997:
@iamareebjamal  Can you please tell in which cases the solution is failing so that I can make the changes?

shivaraju07:
I tried this method. It works fine for me.

iamareebjamal:
The thing is that we are using Realm, it should automatically update no matter if bound in onResume or onStart, the problem is in the reactivity and not the lifecycle. If we say that the bug is fixed by allowing this workaround, then it may pass several subtle bugs in the future

raj10071997:
@iamareebjamal Ok. Then I'll try to think of something else.

iamareebjamal:
@raj10071997 Status?

raj10071997:
@iamareebjamal  sorry for the late reply. The auto update of the objects in realm only happens when that object is a model object. I referred to this link - https://medium.com/@ffvanderlaan/realm-auto-updated-objects-what-you-need-to-know-b2d769d12d76. And LiveData only notifies active observers about updates. Also  "session" is an object of LiveData which will only notify the observer only when it gets updated. So when onResume method is called, it will notify only when the session object is updated. And since it session is not a Model object of realm it will not be auto updated and will return the old list.

android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragmentViewModel.java
```diff
@@ -58,10 +59,12 @@ public AboutFragmentViewModel() {
     }
 
     public LiveData<List<Object>> getBookmarkedSessions() {
-        if (sessions == null) {
+        if (sessions == null || numberOfBookmarkedSessions != realmRepo.getBookMarkedSessionsLength()) {
             LiveRealmData<Session> sessionLiveRealmData = RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions());
             sessions = Transformations.map(sessionLiveRealmData, this::getSessionsList);
```
iamareebjamal, line 64: `realmRepo.getBookMarkedSessions()` these should be reactive and live and hence should be updated automatically

raj10071997: But without going inside the if condition how the sessions object will be updated because we are returning sessions object. I added this condition "numberOfBookmarkedSessions != realmRepo.getBookMarkedSessionsLength()" so that the body of the if condition will be executed only when there is change in the realm.

raj10071997: Actually I tried and changed the body of the if condition to this,

``` 
 if(sessions == null) {
      sessions = Transformations.map(RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions()),this::getSessionsList);
   }  
```

I changed this because realmRepo.getBookMarkedSessions() should be executed if the object is auto updated but still after changing the code to this sessions object returns the old list.

iamareebjamal: I would recommend that you actually try to get bookmarked sessions from RealmRepository and add a listener on them and check if you are getting change event or not. This will reveal the problem

raj10071997: Ok.

raj10071997: Yes on attaching a listener to the object returned by realmRepo.getBookMarkedSessions(), I am getting the change event, but the sessions object is not getting auto updated. Also I think that sessions object is an object of LiveData so why it should be auto updated. I made the following changes in the constructor of AboutFragmentViewModel.java

```
private boolean check = false;

 public AboutFragmentViewModel() {
        realmRepo = RealmDataRepository.getDefaultInstance();
        s = realmRepo.getBookMarkedSessions();
        sessionLiveRealmData = RealmDataRepository.asLiveData(s);
        s.addChangeListener(sessions -> {
            sessionLiveRealmData = RealmDataRepository.asLiveData(sessions);
            check = true;
        });
    }
```
And changed the getBookMarkedSessions() method to this-
```
 public LiveData<List<Object>> getBookmarkedSessions() {
        if (sessions == null || check) {
            sessions = Transformations.map(sessionLiveRealmData, this::getSessionsList);
            check = false;
        }
        return sessions;
    }
```

But still the list is not getting updated. Can you please tell me where am I doing wrong?

iamareebjamal: > Also I think that sessions object is an object of LiveData so why it should be auto updated.

This is exactly why it should get updated. LiveData is meant to be reactive and change aware. So if it is not being updated, it is a bug in the implementation of conversion into LiveData

As to why your code is not working, you create an instance of `sessionLiveRealmData`, let's say it is instance `ada2313v`, so the fragment is listening to `ada2313v` and at each change you just reassign the reference to `sessionLiveRealmData` to a new instance, let's say `e4ew335r`. The fragment is still listening to `ada2313v`, it doesn't care about your local change. If things were simple (reactive) this way, we wouldn't even need LiveData, Observables and everything. Want to notify an object has changed, just reassign the variable to a new thing and it will be changed everywhere!

This confusion is the reason why we should always use final (non-reassignable) properties wherever possible so that changes can be propagated by mutating the LiveData and not by reassigning it (which does nothing but make the app slower by Object Instantiation)

raj10071997: ``` So if it is not being updated, it is a bug in the implementation of conversion into LiveData ``` 
Yes, it is the bug in this method Transformations.map(sessionLiveRealmData, this::getSessionsList); So should I make changes in this method ?

iamareebjamal: Yup

raj10071997: @iamareebjamal I got one solution and it is working. I changed the parent class of LiveRealData.java from LiveData<RealmResults<T>> to MutableLiveData<RealmResults<T>> 
and changed the constructor to this

```
public AboutFragmentViewModel() {
        realmRepo = RealmDataRepository.getDefaultInstance();
        s = realmRepo.getBookMarkedSessions();
        sessionLiveRealmData = RealmDataRepository.asLiveData(s);
        s.addChangeListener(sessions -> {
            sessionLiveRealmData.setValue(sessions);
        });
    }
```
The above solution is working. You were right, the ```Transformations.map(sessionLiveRealmData, this::getSessionsList) ``` method was listening to old reference of ```sessionLiveRealmData``` so it will not listen to new data and so I changed the  ```sessionLiveRealmData``` object parent class and changed it's value inside the listener. And it is working fine. Can you tell if anything wrong in this method ?

iamareebjamal: Well obviously this will work because now we are manually triggering the change through a change listener. The idea of RealmLiveData is to automatically do it removing the need of manual changes. This is not the solution unfortunately. I am currently busy or would have seen what's the issue. @srv-twry Can you please help?

raj10071997: @iamareebjamal  I thought of this solution as there is no way to update the value of LiveData object and the Transformations.map() method only executes if there is a change in the LIveData object passed in it as a parameter and since it was listening to old LiveData object(i.e. sessionLiveRealmData ) there is no other way to update it and the documentation says that MutableLiveData has the setValue method to change the stored value. That's why I thought of changing the parent class as the only difference between LiveData and MutableLiveData is that MutableLiveData has two extra methods to change the stored value. And that's why I thought of this solution. But if you say so I'll try to think of something else.

dr0pdb: @iamareebjamal Yes, looking into it.

dr0pdb: @raj10071997 After going through the issue, i don't think we need to think so much about it. Everything was correctly implemented except in the AboutFragment's ```loadData``` method.

Can you please replace 
``` Java
aboutFragmentViewModel.getBookmarkedSessions().observe(this, sessionsList -> {
            sessions.clear();
            sessions.addAll(sessionsList);
            bookMarksListAdapter = new GlobalSearchAdapter(sessions, getContext());
            bookMarksListAdapter.setOnBookmarkSelectedListener(this);
            bookmarksRecyclerView.setAdapter(bookMarksListAdapter);
            handleVisibility();
        });
```

with this 

``` Java
aboutFragmentViewModel.getBookmarkedSessions().observe(this, sessionsList -> {
            sessions.clear();
            sessions.addAll(sessionsList);
            bookMarksListAdapter.notifyDataSetChanged();
            handleVisibility();
        });
```

dr0pdb: videotogif_2018 03 07_01 11 59

raj10071997: @srv-twry I tried your changes locally. I only made the changes you specified and tested them locally. In my phone the bookmarked sessions appear only when they are equal to 3. And if I try to mark more than that, then only 4 bookmarked sessions are shown and then when the sessions become equal to 6, all the bookmarked sessions are shown. The gif shows only the issue about the 3 sessions.

videotogif_2018 03 07_13 22 49

raj10071997: @srv-twry The changes you mentioned are the only changes that needed to be done?

dr0pdb: Yes, they are the only changes that i made. 
The error you're getting is weird. Can you completely uninstall the application and reinstall it after making the above changes ?

raj10071997: I am still getting the same results. I also noticed one thing that when you install the app so you are basically logged out (now don't login) and then if you try to bookmark the events, the events don't get displayed at all in AboutFragment.java, doesn't matter how many events you bookmark. Are you getting the same results too when you install the app and you are logged out? I am using Android 7.0 Nougat (API 24).

dr0pdb: The above GIF that i uploaded has me logged out only.

raj10071997: Ok. I'll try to see why am I getting these errors. As I cleared the cache data and uninstalled the application and then installed it with your suggested changes. I have my development branch updated with the remote branch. And still I am getting these errors. I'll see why am I getting these errors and will notify you soon.

PR #2333, opened 2018-02-13
Fix #2326 - Move Twitter and Facebook feed under one fragment.

### LINKED ISSUE #2326: Move both Twitter and Facebook feed under one fragment
**Actual Behaviour**

Currently, there are two separate items in the drawer for twitter and facebook.

**Expected Behaviour**

I propose to move both the fragments under a common "Live Feed" fragment using a `View Pager`.

Yes

@iamareebjamal @srv-twry Please see if this is a good change and is it required atm?

mayank8318:
Fixes #2326 
Changes: Made a new fragment with a view pager and tab layout which shows the two feed fragments.
Also refactored feedFragment to FacebookFeedFragment.

Screenshots of the change: 
image

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/feed/FeedFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/feed/FeedFragment.java
@@ -1,0 +1,64 @@
+package org.fossasia.openevent.core.feed;
+
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.design.widget.TabLayout;
+import android.support.v4.view.ViewPager;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import org.fossasia.openevent.R;
+import org.fossasia.openevent.common.ui.base.BaseFragment;
+import org.fossasia.openevent.common.ui.image.OnImageZoomListener;
+import org.fossasia.openevent.core.feed.facebook.FacebookFeedFragment;
+import org.fossasia.openevent.core.feed.twitter.TwitterFeedFragment;
+
+import butterknife.BindView;
+
+public class FeedFragment extends BaseFragment {
+
+    @BindView(R.id.viewpager)
+    protected ViewPager viewPager;
+    @BindView(R.id.tabLayout)
+    protected TabLayout feedTabLayout;
+
+    private OpenCommentsDialogListener openCommentsDialogListener;
+    private OnImageZoomListener onImageZoomListener;
+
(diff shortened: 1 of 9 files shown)


android/app/src/main/java/org/fossasia/openevent/core/feed/FeedFragment.java
```diff
@@ -0,0 +1,64 @@
...
+        feedTabLayout.setupWithViewPager(viewPager);
+
+        return view;
+    }
+
+    private void setUpViewPager(ViewPager viewPager) {
+        FeedViewPagerAdapter feedViewPagerAdapter = new FeedViewPagerAdapter(getChildFragmentManager());
+        feedViewPagerAdapter.addFragment(FacebookFeedFragment.getInstance(openCommentsDialogListener, onImageZoomListener), "FACEBOOK");
+        feedViewPagerAdapter.addFragment(TwitterFeedFragment.getInstance(onImageZoomListener), "TWITTER");
+        feedViewPagerAdapter.notifyDataSetChanged();
```
iamareebjamal: Use strings.xml and don't capitalize by default, it is handled automatically by Andriod Tab Layout

iamareebjamal (changes requested):

mayank8318:

android/app/src/main/java/org/fossasia/openevent/core/feed/FeedFragment.java
```diff
@@ -0,0 +1,64 @@
...
+import butterknife.BindView;
+
+public class FeedFragment extends BaseFragment {
+
+    @BindView(R.id.viewpager)
+    protected ViewPager viewPager;
+    @BindView(R.id.tabLayout)
+    protected TabLayout feedTabLayout;
+
+    private FacebookFeedAdapter.OpenCommentsDialogListener openCommentsDialogListener;
```
iamareebjamal: If it is general, why is it inside FacebookFeedAdapter

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/feed/FeedFragment.java
mayank8318: @iamareebjamal 
Presently only Facebook comments are shown. Twitter feed doesn't support comments as of now.
Should I move it to a general location and make an issue for adding comments to twitter feed?

iamareebjamal: If it is used in FeedFragment, it should be in general package, let's not show twitter comments but just move it to general package

mayank8318:
Changes: Move openCommentsDialogListener to an independent interface.

iamareebjamal (approved):
Good job

Issue #2259, opened 2018-01-27
Make loading bar Invisible in login activity and show text


jindalAnuj:
**Actual Behaviour**

loading bar keeps on rotating by showing details invalid email 

**Expected Behaviour**

loading bar should be invisible and a toast is expected
**Steps to reproduce it**

loading bar should be replaced or made invisible

**Screenshots of the issue**

![ezgif com-video-to-gif 2](https://user-images.githubusercontent.com/31108317/35472439-2aef5e28-0395-11e8-8b53-aa24729efc17.gif)

**Would you like to work on the issue?**
yes

ahsan44411:
@jindalAnuj are you still working on this issue ?

PR #2341, opened 2018-02-16
Progress bar bug fixes 

### LINKED ISSUE #2258: Loading Bar in Main Activity
**Actual Behaviour**

There is no loading bar in main activity if anybody use app with slow internet connection then user may end up uninstalling the app.This all happens when app is opened first time

**Expected Behaviour**

There should be a loading bar in the main ativity

**Steps to reproduce it**

open App first time after installing

**Screenshots of the issue**

ezgif com-video-to-gif 1

yes

simarsingh24:
Fixes #2258 #2259 
since there was just minor 1 line addition in #2259 I clubbed the PR's


android/app/src/main/java/org/fossasia/openevent/core/main/MainActivity.java
```diff
@@ -849,7 +847,7 @@ public void downloadFromAssets() {
     }
 
     public void readJsonAsset(final String name) {
-        CommonTaskLoop.getInstance().post(new Runnable() {
+        CommonTaskLoop.getInstance().delayPost(new Runnable() {
```
simarsingh24, line 850: @iamareebjamal  this works, not sure if its a good way (adding delay), is there any better possible solution for this?

iamareebjamal: Use RxJava instead

iamareebjamal: This is not the correct way. Check out why the dialog box is not shown in the first place

PR #2344, opened 2018-02-16
feat: Add Session type in SessionDetailActivity (#2342)

### LINKED ISSUE #2342: Add session Type in session details activity
**Actual Behaviour**

Sesssion details activity doesn't mention anything about the type of the session.

**Expected Behaviour**

The type of the session should be displayed

**Steps to reproduce it**

Select any session from the sessions of any speaker

Yes

AkshayCHD:
Fixes #2342 

Changes: The type of session is now visible in SessionDetailActivity
screenshot_1518905232

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
@@ -52,6 +52,7 @@ import org.fossasia.openevent.core.bookmark.BookmarkStatus;
 import org.fossasia.openevent.core.bookmark.OnBookmarkSelectedListener;
 import org.fossasia.openevent.data.Microlocation;
 import org.fossasia.openevent.data.Session;
+import org.fossasia.openevent.data.SessionType;
 import org.fossasia.openevent.data.Speaker;
 import org.fossasia.openevent.data.Track;
 
@@ -112,6 +113,12 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
     protected ImageView youtubeThumbnail;
     @BindView(R.id.watch)
     protected ImageButton playButton;
+    @BindView(R.id.sessionType)
+    protected TextView sessionType;
+    @BindView(R.id.type)
+    protected TextView typeLabel;
+    @BindView(R.id.typeLayout)
+    protected LinearLayout typelayout;
 
     private String trackName, title, location;
     private int trackColor, darkColor, fontColor;
@@ -139,6 +146,7 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
 
         title = getIntent().getStringExtra(ConstantStrings.SESSION);
         trackName = getIntent().getStringExtra(ConstantStrings.TRACK);
+
         if (TextUtils.isEmpty(trackName))
             hasTrack = false;
         id = getIntent().getIntExtra(ConstantStrings.ID, 0);
@@ -205,6 +213,13 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
         } else {
             location = getString(R.string.location_not_decided);
             text_room1.setText(location);
+        }
+
+        SessionType type = session.getSessionType();
+
+        if (type != null) {
+            sessionType.setText(type.getName());
+            typelayout.setVisibility(View.VISIBLE);
         }
 
         text_title.setText(title);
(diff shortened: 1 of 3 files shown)


AkshayCHD:
@iamareebjamal please review

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -112,6 +113,13 @@
     protected ImageView youtubeThumbnail;
     @BindView(R.id.watch)
     protected ImageButton playButton;
+    @BindView(R.id.sessionType)
+    protected TextView session_type;
+    @BindView(R.id.dot)
+    protected TextView dot;
+    @BindView(R.id.type)
+    protected TextView type_textview;
+
```
iamareebjamal: That's not how you should do it. Define only one textview. If you need to hide it, wrap the layout in LinearLayout

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
AkshayCHD: @iamareebjamal but if i do that the grid view will distort, the gridview has three columns in every row,
If i create two dummy textviews after the LinearLayout then the positioning of ':' will not be proper.
Either i should create a separate issue moving from grid layout to linear layout, as in the current code already there is a problem that when the track name is empty only ':' would be displayed.

iamareebjamal: You can use framelayout as a wrapper. Check if that is possible

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -112,6 +113,10 @@
     protected ImageView youtubeThumbnail;
     @BindView(R.id.watch)
     protected ImageButton playButton;
+    @BindView(R.id.sessionType)
+    protected TextView session_type;
```
iamareebjamal: Java follows camel case style

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -207,6 +213,15 @@ private void updateSession() {
             text_room1.setText(location);
         }
 
+        SessionType type = session.getSessionType();
+
+        if(type != null) {
```
iamareebjamal: Syntax issue

android/app/src/main/res/layout/content_session_detail.xml
```diff
@@ -67,6 +67,42 @@
...
+                    <LinearLayout
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:layout_columnSpan="3"
+                        android:id="@+id/typeLayout"
+                        android:layout_rowSpan="1">
+
+                        <TextView
+                            android:id="@+id/type"
+                            android:layout_width="62sp"
```
iamareebjamal: Never hardcode width, and sp is used for text size

iamareebjamal (changes requested):

AkshayCHD:
@iamareebjamal please review

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -207,6 +215,15 @@ private void updateSession() {
             text_room1.setText(location);
         }
 
+        SessionType type = session.getSessionType();
+
+        if (type != null) {
+            location = type.getName();
+            sessionType.setText(type.getName());
```
iamareebjamal: Show the type layout if it is present, hide by default

iamareebjamal (changes requested):

AkshayCHD:
@iamareebjamal please review, there has been no change in the UI so the screenshots have not been updated

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -207,6 +215,14 @@ private void updateSession() {
             text_room1.setText(location);
         }
 
+        SessionType type = session.getSessionType();
+
+        if (type != null) {
+            location = type.getName();
```
iamareebjamal: Why is location set to type name?

AkshayCHD:
@iamareebjamal removed that, sorry

iamareebjamal (approved):

PR #2346, opened 2018-02-16
fix #2343: Fix crash, show invalid track message instead

### LINKED ISSUE #2343: App crashes while bookmarking session in SessionDetailActivity
**Actual Behaviour**

App crashes while bookmarking session in SessionDetailActivity

**Expected Behaviour**

App shouldn't crash

**Steps to reproduce it**

Go to 'Tracks' in nav drawer
Select 'Open Tech'
Click on the first item i.e. 'Keynote: Pro...'
Bookmark it

**LogCat for the issue**

`FATAL EXCEPTION: main
                                                                        Process: org.fossasia.openevent, PID: 12533
                                                                        java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.fossasia.openevent.data.Track.getColor()' on a null object reference
                                                                            at org.fossasia.openevent.core.track.session.SessionDetailActivity.lambda$null$0$SessionDetailActivity(SessionDetailActivity.java:171)`

Yes

pulakdp:
Fixes #2343

Changes: Added null check for track. If null, showed "Invalid Track" in Snackbar


android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -154,6 +154,11 @@ public void onCreate(final Bundle savedInstanceState) {
             if(session == null)
                 return;
 
+            if (session.getTrack() == null) {
+                Snackbar.make(speakersRecyclerView, R.string.invalid_track, Snackbar.LENGTH_SHORT).show();
+                return;
+            }
```
iamareebjamal: Use `RealmRepository.isTrackNull` instead

iamareebjamal (changes requested):

pulakdp:
@iamareebjamal Can you check the latest commit now? Let me know if any further changes are needed. I'm new to using github in collaboration so any help would be appreciated!

iamareebjamal (approved):

PR #2351, opened 2018-02-21
Fix: Speaker's color changes when searching

### LINKED ISSUE #2348: Speakers' color changes in SpeakerListFragment when search is being done
**Actual Behaviour**
Speakers' color changes in SpeakerListFragment when search is being done

**Expected Behaviour**
Speakers' color should not change in SpeakerListFragment when search is being done

**Screenshots of the issue**
bugatspeakers

See the speaker on the top right. His panel changes from red to blue.
It sometimes changes to yelllow too.

Yes, though I'm not sure where the colors of the speakers are set.
Has this issue been reported before?

leeyh20:
Fixes #2348 

Changes: Images (bitmaps) without a "vibrant" image profile will return null, and originally the background color of the speaker was never set. 

Now, it tries to get a "vibrant" image profile. If it fails, it will try to get a "dark muted" image profile. I chose "dark muted" because the text of the speaker info is white and requires a darker color. 

If there is still no "dark muted" profile, the bg color will be set to a default color, which is a shade of dark red.

Screenshots for the change: 
image
Originally the first two speakers did not have a "vibrant" image profile thus their colors kept changing.

When searching:
image


leeyh20:
@iamareebjamal Do help me to review, thank you!

mayank8318 (approved):

simarsingh24 (approved):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerViewHolder.java
```diff
@@ -95,8 +95,21 @@ public void bindSpeaker(Speaker speaker) {
...
+                    // bitmap has vibrant swatch profile
+                    speakerTextualInfo.setBackgroundColor(swatch.getRgb());
+                } else {
+                    Palette.Swatch swatch2 = palette.getDarkMutedSwatch();
+                    if (swatch2 != null) {
+                        // bitmap has dark muted swatch profile
+                        speakerTextualInfo.setBackgroundColor(swatch2.getRgb());
+                    } else {
+                        // Default color is red if swatches cannot be found
+                        int defaultColor = 0xFFb20000;
```
iamareebjamal: Don't use hardcoded color, load from resources

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerViewHolder.java
```diff
@@ -105,9 +106,7 @@ public void bindSpeaker(Speaker speaker) {
                         // bitmap has dark muted swatch profile
                         speakerTextualInfo.setBackgroundColor(swatch2.getRgb());
                     } else {
-                        // Default color is red if swatches cannot be found
-                        int defaultColor = 0xFFb20000;
-                        speakerTextualInfo.setBackgroundColor(defaultColor);
+                        speakerTextualInfo.setBackgroundColor(ContextCompat.getColor(context, R.color.speakers_viewholder_background));
```
iamareebjamal: No, please use one of the theme colors. It is customisable through the app generator

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerViewHolder.java
leeyh20: Oh, thanks for the clarification!

leeyh20:
@iamareebjamal Is it okay to use color primary dark?

iamareebjamal (approved):

PR #2353, opened 2018-02-25
Fix #2350 - Move implementation of commentsdialogfragment and onzooml…

### LINKED ISSUE #2350: Move implementation of commentsdialogfragment to respective fragment
**Actual Behaviour**

CommentsDialogFragment is implemented in the MainActivity/

**Expected Behaviour**

It should be implemented in the activity which uses it.

Yes

mayank8318:
…istener to respective fragments.

Fixes #2350 

Changes: Move the implementation of commentsdialogfragment and onzoomlistener to respective fragments.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/feed/facebook/FacebookFeedFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/feed/facebook/FacebookFeedFragment.java
@@ -17,9 +17,11 @@ import org.fossasia.openevent.common.ConstantStrings;
 import org.fossasia.openevent.common.network.NetworkUtils;
 import org.fossasia.openevent.common.ui.Views;
 import org.fossasia.openevent.common.ui.image.OnImageZoomListener;
+import org.fossasia.openevent.common.ui.image.ZoomableImageUtil;
 import org.fossasia.openevent.common.utils.SharedPreferencesUtil;
 import org.fossasia.openevent.core.feed.BaseFeedFragment;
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/org/fossasia/openevent/core/feed/facebook/FacebookFeedFragment.java
```diff
@@ -63,8 +61,8 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         facebookFeedFragmentViewModel = ViewModelProviders.of(this).get(FacebookFeedFragmentViewModel.class);
         RecyclerView.LayoutManager mLayoutManager = new LinearLayoutManager(getContext());
         feedRecyclerView.setLayoutManager(mLayoutManager);
-        facebookFeedAdapter = new FacebookFeedAdapter(getContext(), openCommentsDialogListener, feedItems);
-        facebookFeedAdapter.setOnImageZoomListener(onImageZoomListener);
+        facebookFeedAdapter = new FacebookFeedAdapter(getContext(), this, feedItems);
+        facebookFeedAdapter.setOnImageZoomListener(this);
```
iamareebjamal: Similar to zoom listener, add the comment dialog listener through a setter, and remove it on onDestroy

iamareebjamal (changes requested):

mayank8318:

iamareebjamal (changes requested):
Please remove both listeners onDestroy, and also, there is no need to check for adapter nullability as it can never be null

iamareebjamal (approved):

PR #2365, opened 2018-03-10
fix #2363: Fix undo button not updating bookmark icon

### LINKED ISSUE #2363: Undo button in SnackBar doesn't update bookmark icon in SpeakerDetailsActivity
**Actual Behaviour**

Clicking undo button in the SnackBar generated after adding a session to bookmarks doesn't update the bookmark icon on that session

**Expected Behaviour**

Clicking the undo button should also change the bookmark icon from bookmarked to not bookmarked

**Steps to reproduce it**

1. Click on any speaker from the Featured Speakers list or go to any speaker's detail activity from the list of speakers
2. Add any session to bookmarks and then click undo in the SnackBar shown.

**Screenshots of the issue**

Yes

pulakdp:
Fixes #2363 

Changes: Added method onBookmarksChanged() that gets called when a bookmark is added or removed. Refactored code that loads speaker details to avoid code repetition. Created method loadData() for that purpose.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
@@ -34,6 +34,7 @@ import com.squareup.picasso.Target;
 
 import org.fossasia.openevent.R;
 import org.fossasia.openevent.common.api.Urls;
+import org.fossasia.openevent.common.events.BookmarkChangedEvent;
 import org.fossasia.openevent.common.events.ConnectionCheckEvent;
 import org.fossasia.openevent.common.ui.SnackbarUtil;
 import org.fossasia.openevent.common.ui.Views;
@@ -54,6 +55,7 @@ import java.util.List;
 
 import butterknife.BindView;
 import butterknife.OnClick;
+import timber.log.Timber;
 
 public class SpeakerDetailsActivity extends BaseActivity implements AppBarLayout.OnOffsetChangedListener, OnBookmarkSelectedListener, SessionsListAdapter.OnItemClickListener {
 
@@ -286,6 +288,21 @@ public class SpeakerDetailsActivity extends BaseActivity implements AppBarLayout
         }
 
     }
+
+    private void loadData() {
+        speakerDetailsViewModel.getSpeaker(speakerName).observe(this, speakerData -> {
+            selectedSpeaker = speakerData;
+            loadSpeakerDetails();
+        });
+    }
+
+    @Subscribe
+    public void onBookmarksChanged(BookmarkChangedEvent bookmarkChangedEvent) {
+        Timber.d("Bookmarks Changed");
+        loadData();
+    }
+
+
     @Override
     protected void onResume() {
         super.onResume();
@@ -298,10 +315,7 @@ public class SpeakerDetailsActivity extends BaseActivity implements AppBarLayout
 
         gridLayoutManager.setSpanCount(spanCount);
 
-        speakerDetailsViewModel.getSpeaker(speakerName).observe(this, speakerData -> {
-            selectedSpeaker = speakerData;
-            loadSpeakerDetails();
-        });
+        loadData();
     }
 
     @Override
--- a/android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
@@ -296,6 +296,7 @@ public class SpeakerDetailsActivity extends BaseActivity implements AppBarLayout
         });
     }
 
+    //TODO: Workaround. Needs fix during implementation of RealmLiveData
     @Subscribe
     public void onBookmarksChanged(BookmarkChangedEvent bookmarkChangedEvent) {
         Timber.d("Bookmarks Changed");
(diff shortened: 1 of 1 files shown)


dr0pdb:
@pulakdp Please atttach a GIF showing that it indeed solves the issue.

pulakdp:
Sure @srv-twry. Here is the gif after the fix

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
```diff
@@ -286,6 +288,21 @@ private void hideEmptyURLButtons() {
...
+        speakerDetailsViewModel.getSpeaker(speakerName).observe(this, speakerData -> {
+            selectedSpeaker = speakerData;
+            loadSpeakerDetails();
+        });
+    }
+
+    @Subscribe
+    public void onBookmarksChanged(BookmarkChangedEvent bookmarkChangedEvent) {
+        Timber.d("Bookmarks Changed");
+        loadData();
```
iamareebjamal: I think this and many other solutions mask a greater problem with the implementation. We can accept these changes. But write a TODO, that this is a workaround and fix needs to be done in the implementation of RealmLiveData

Also, we are moving away from event bus architecture, so we don't want to add new code using the pattern

iamareebjamal (changes requested):

pulakdp:
@iamareebjamal Added the TODO, have a look

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
```diff
@@ -286,6 +288,22 @@ private void hideEmptyURLButtons() {
...
+            loadSpeakerDetails();
+        });
+    }
+
+    //TODO: Workaround. Needs fix during implementation of RealmLiveData
+    @Subscribe
+    public void onBookmarksChanged(BookmarkChangedEvent bookmarkChangedEvent) {
+        Timber.d("Bookmarks Changed");
+        loadData();
+    }
```
iamareebjamal, line 304: This needs to be added in onResume

And the solution should not contain event handling

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
pulakdp: > This needs to be added in onResume

You mean the TODO?

> And the solution should not contain event handling

How do you suggest I do this @iamareebjamal ? I'm relatively new to the project and I came to the solution after looking at how things were done in TrackSessionsActivity.java. setBookmark() in RealmDataRepository class posts the event so my first thought was to add this method onBookmarksChanged(). If there are other ways to do this, suggest me and I'll try that. Thanks!

iamareebjamal: > You mean the TODO?
Yes

And about the event based method, yes this is how we are doing things currently in the project but we are refactoring code to avoid this and hence we don't want to add new code using this bad practice. The LiveData itself should handle the reactivity and there should be no need of the change event

pulakdp: Okay, thanks for the clarification @iamareebjamal! For now, just moved the TODO.

pulakdp:
Moved the TODO

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
```diff
@@ -286,6 +288,22 @@ private void hideEmptyURLButtons() {
...
+
+    @Subscribe
+    public void onBookmarksChanged(BookmarkChangedEvent bookmarkChangedEvent) {
+        Timber.d("Bookmarks Changed");
+        loadData();
+    }
+
+
+    //TODO: Workaround. Needs fix during implementation of RealmLiveData
```
iamareebjamal: Should be at the method call of loadData

pulakdp:
Apologies! Moved TODO inside onResume where loadData is called @iamareebjamal

pulakdp:
@iamareebjamal Kindly review and let me know if any further changes are needed

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakerDetailsActivity.java
iamareebjamal: The event bus pattern is still used

pulakdp: > We can accept these changes. But write a TODO, that this is a workaround and fix needs to be done in the implementation of RealmLiveData

Isn't this why we added a TODO @iamareebjamal? I understand this isn't what you want in future but your earlier instruction was to include a TODO so I did likewise. If a solution with RealmLiveData is expected at the moment then I'll have to learn about LiveData more before I come up with the solution.

iamareebjamal:
@pulakdp The solution still consists of eventbus pattern

pulakdp:
> We can accept these changes. But write a TODO, that this is a workaround and fix needs to be done in the implementation of RealmLiveData

Isn't this why we added a TODO @iamareebjamal? I understand this isn't what you want in future but your earlier instruction was to include a TODO so I did likewise. If a solution with RealmLiveData is expected at the moment then I'll have to learn about LiveData more before I come up with the solution.

iamareebjamal:
The TODO is for the hook in onResume, event bus is deprecated, but I guess we can go ahead with it. Add another above the subscribe event

`TODO: High Priority: Remove event bus pattern`

pulakdp:
@iamareebjamal done.

iamareebjamal (approved):

PR #2381, opened 2018-03-18
fix: Fix implementation of LiveData in SpeakerDetailActivity.

### LINKED ISSUE #2380: Speaker LiveData does not change on modify sessions in SpeakerDetailActivity.
**Actual Behaviour**

In SpeakerDetailActivity modifying session does not automatically change Speaker LiveData. 

**Expected Behaviour**

Implementation of Speaker LiveData should be change so that modifying session should modify Speaker LiveData.

yes

ParthivMangukiya:
Fixes #2380

Changes: Added getSpeakerAsRealmResults method which return Speaker as RealmResult Instead of Realm Objects.


ParthivMangukiya:
@iamareebjamal Please Review.
I found that changeListener of RealmObject does not call if any object of spawnList is modify.
so I've added getSpeaker method in realmRepository which return speaker as RealmResult than took the first element on Transformation.map method.

getSpeaker() method which returns Speaker is no longer used should i remove it?

iamareebjamal:
Please rename the method to `getSpeakersForName`

ParthivMangukiya:
@iamareebjamal Should I change method name of getSpeakerAsRealmResults or getSpeaker of SpeakerDetailViewModel or both? 

Should I remove old getSpeaker() method of RealmDataRepository which returns Speaker not RealmResults<Speaker>?

iamareebjamal:
Of `getSpeakerAsRealmResults` 

And no, don't delete that method for now

ParthivMangukiya:
@iamareebjamal done, Please Review.

iamareebjamal (approved):

iamareebjamal:
Please ensure that #2363 is not occurring because the code removed was fixing that

ParthivMangukiya:
@iamareebjamal yes i am aware of that issue. I'd tested it. here is the screenshot for the same.

PR #2383, opened 2018-03-18
fix: Fix unexpected bookmark behaviour in AboutFragment

### LINKED ISSUE #2382: Unexpected behaviour in AboutFragment when bookmark is removed from SessionDetailActivity
**Actual Behaviour**

Session in AboutFragment is sometimes removed and sometimes not removed when bookmark is removed from SessionDetailActivity.

**Expected Behaviour**

Session in AboutFragment should always removed when bookmark is removed from SessionDetailActivity.

**Screenshots of the issue**

yes

ParthivMangukiya:
Fixes #2382

Changes: Changed implementation of getBookmarkedSession() so that it can handle case for session which was not before bookmarked and when bookmark is removed it only return bookmarkedSessions By rerun filter.

Screenshots for the change:


ParthivMangukiya:
@iamareebjamal Please Review.

GlobalSearchAdapter(which is bookMarksListAdapter) is created new each time when any change in LiveData. Is it ok to add reset method in GlobalSearchAdapter in this PR so that it does not need to create each time.

android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragmentViewModel.java
```diff
@@ -58,14 +60,15 @@ public AboutFragmentViewModel() {
...
-            LiveRealmData<Session> sessionLiveRealmData = RealmDataRepository.asLiveData(realmRepo.getBookMarkedSessions());
-            sessions = Transformations.map(sessionLiveRealmData, this::getSessionsList);
-        }
-        return sessions;
+        if (filterableRealmLiveData == null)
+            filterableRealmLiveData = RealmDataRepository.asFilterableLiveData(realmRepo.getSessions());
+        Predicate<Session> predicate = Session::getIsBookmarked;
+        filterableRealmLiveData.filter(predicate);
+
+        return Transformations.map(filterableRealmLiveData, this::getSessionsList);
```
iamareebjamal: Pulling in all sessions and then filtering is very inefficuent

ParthivMangukiya: @iamareebjamal this is inefficient but I think this should be expected behaviour. 
If some session is bookmarked (which was not before) than changelistener of realm object will not be triggered because it was not in source set.

When we will change to room than I think
it will be automatically taken care by room.

I've changed to this because this is ideal behaviour(changelistener call each time) and it will be easier to change to room. we will not need to change any other logic other than getBookmarkedSessions.

ParthivMangukiya:
@iamareebjamal Please Review

android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragmentViewModel.java
ParthivMangukiya: @iamareebjamal Sorry, my mistake. I've changed it to getBookmarkedSession and it worked because of change listener also notify about data changes in all sessions not only on received sessions.

iamareebjamal (approved):

PR #2394, opened 2018-03-24
added scroll bar in speakerslistfragment

### LINKED ISSUE #2207: Addition of a scroll back fab in speakers list fragment
**Actual Behaviour**

As is evident, the list of speakers is quite long and there is no way to go immediately very bottom or to the top(if at bottom).

**Expected Behaviour**

So there must be a **scroll to top** fab button as the user scrolls down. Other alternative is providing a list of alphabets on the right end, where tapping on a alphabet scrolls down to that part of the list.   

Yes!!

I would fix the issue in maximum two days time.

sriramr98:
Fixes #2207
Changes: Added a scroll bar in SpeakersListFragment which acts as a fast scroller. Scrolling using the scroll bar shows the starting alphabet of the name of the speaker in a bubble.

Screenshot :

gif


sriramr98:
@iamareebjamal please review this PR. I would recommend we reduce the width of each item. Because we cant bring the scroller any closer to the right end and it collides with the recycler view item. So I think we should either reduce the width a little bit or convert the staggered grid into a linear layout recycler view

iamareebjamal:
Simple, rather than using margin in recyclerview, use padding

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
```diff
@@ -137,4 +139,12 @@ else if (!TextUtils.isEmpty(speakerData) && (sortType == 1 || sortType == 2))
             holder.header.setText(String.valueOf("#"));
     }
 
+    @Override
+    public String getSectionTitle(int position) {
+        //this String will be shown in a bubble for specified position
+        Speaker speaker = getItem(position);
+        String name = speaker.getName();
+        return name.substring(0, 1);
+    }
```
iamareebjamal: This is only valid for sorted by name, will break for other sort criterias

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
sriramr98: I don't think so because the method executes for each item. So for each item, the first alphabet is taken and displayed.

sriramr98: I tested it with other sort criteria. It works as expected.

iamareebjamal: Not saying that. If the speakers are sorted by country, country's first character should be displayed in scroller and not name's

sriramr98: oh okay. Ill make the changes

sriramr98: @iamareebjamal For some speakers, there is no country or organization. What do you want me to do in that case?

iamareebjamal: `-`

sriramr98: done

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
```diff
@@ -143,8 +146,20 @@ else if (!TextUtils.isEmpty(speakerData) && (sortType == 1 || sortType == 2))
...
+        } else if (sortType == 1) {
+            title = speaker.getOrganisation();
+        } else {
+            title = speaker.getCountry();
+        }
+
+        if (title.length() == 0 || title.isEmpty()){
+            return "-";
+        }
+        return String.valueOf(title.charAt(0));
```
iamareebjamal: What's 0 and 1?

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
sriramr98: are you asking about the sortOrder?
Those are the sort orders used in the bindView method. 
I was looking for a constant but there was none.

iamareebjamal: There is, and if it isn't please create these

sriramr98: @iamareebjamal done

iamareebjamal:
Please add a gif of the functionality

sriramr98:
@iamareebjamal  added gif

iamareebjamal:
Scroller should align to right

sriramr98:
The scroller is at right.

iamareebjamal:
Not all the way

sriramr98:
Any more right and the user won't be able to drag it properly. And that's the default margin from the library. I haven't added any margin to the scroller as you can see.

iamareebjamal:
Alright

iamareebjamal (approved):

android/app/build.gradle
```diff
@@ -155,4 +155,7 @@ dependencies {
 
     // Zoomable image view
     implementation 'com.github.chrisbanes:PhotoView:2.1.3'
+
+    // fast scroll rv
+    compile 'com.futuremind.recyclerfastscroll:fastscroll:0.2.5'
```
iamareebjamal: Should be implementation

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
```diff
@@ -119,22 +125,42 @@ public HeaderViewHolder onCreateHeaderViewHolder(ViewGroup parent) {
...
+        Speaker speaker = getItem(position);
+        sortType = SharedPreferencesUtil.getInt(ConstantStrings.PREF_SORT_SPEAKER, SORTED_BY_NAME);
+        String title;
+        if (sortType == SORTED_BY_NAME) {
+            title = speaker.getName();
+        } else if (sortType == SORTED_BY_ORGANIZATION) {
+            title = speaker.getOrganisation();
+        } else {
+            title = speaker.getCountry();
+        }
```
iamareebjamal: Use switch case

android/app/src/main/java/org/fossasia/openevent/core/speaker/SpeakersListAdapter.java
```diff
@@ -119,22 +125,42 @@ public HeaderViewHolder onCreateHeaderViewHolder(ViewGroup parent) {
...
+
+        if (title.length() == 0 || title.isEmpty()){
```
iamareebjamal: TextUtils.isEmpty

iamareebjamal (changes requested):

iamareebjamal (approved):

sriramr98:
@iamareebjamal  This PR can be merged.

PR #2396, opened 2018-03-25
removed language setting from settings

### LINKED ISSUE #2384: Unexpected behavior : app crashes when change the language
**Actual Behaviour**

App crashes when change the language to anything rather than English , when closing the app and reopen it works but it crashes again when opening settings activity.

**Expected Behaviour**

App should work without crashing after changing language .

**Screenshots of the issue**

open_event_android

yes.

sriramr98:
Fixes #2384

Changes: Removed language change setting from the Settings Activity.

Screenshots of the change: 

Before language changed :

After language changed:


iamareebjamal:
Is the app crashing when the language is changed?

sriramr98:
No, it isn't crashing. I have attached a screenshot as proof. Some strings actually do change language while most don't.

iamareebjamal (approved):

Issue #2357, opened 2018-03-07
Adding Kotlin Support


sriramr98:
**Steps to reproduce it**
Kotlin is an awesome language that introduces verbosity and conciseness for the project. Adding Kotlin support to the project will improve the quality of the code with significantly reduced codebase. I would like to start by adding kotlin in incremental builds - one fragment at a time.

**Would you like to work on the issue?**

Yes. I would like to work on this issue

**Target Milestone to solve the issue**
I would like to take this up for GSOC 2018 as a milestone.

the-dagger:
@sriramr98 I think you can go ahead with converting one Pojo to Kotlin and we can review that and merge.
@iamareebjamal Does that sound good, correct me if this is not in our current implementation list?

sriramr98:
I think we should start with converting utility classes instead of a Pojo. There are some small issues with data classes and Realm that needs to be fixed before converting a Pojo. So I think we should start with some Utility Classes.

iamareebjamal:
@the-dagger Sounds good, And I agree, utility classes should be converted first as the data layer s already being reworked in this library https://github.com/iamareebjamal/openevent-java

sriramr98:
So I will start converting two-two utility classes at a time

mayank8318:
@sriramr98 Would you like to split up?

sriramr98:
@mayank8318 I'm sorry man, I'm trying to include this as a small part of my GSoC proposal. If I don't get selected for it, then we can split up the work, collab and get the job done.

iamareebjamal:
The PR does not solve this issue as this issue will be closed when entire project is completely converted to kotlin

sriramr98:
@iamareebjamal  Can we continue with this issue? Should we start with some util classes or start refactoring activities?

iamareebjamal:
Util classes first

sriramr98:
Alright. Ill push two util classes a day and we'll see how it goes.

sriramr98:
@iamareebjamal the utility classes have been converted. What next?

iamareebjamal:
Convert non android specific classes

PR #2399, opened 2018-03-27
Converted three util classes to Kotlin

sriramr98:
First steps for issues #2357 

Changes: Converted three Util classes to kotlin


android/app/src/main/java/org/fossasia/openevent/core/auth/AuthUtil.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    const val VALID = 0
+    const val EMPTY = 1
+    const val INVALID = 2
+
+    private var authenticator: Authenticator? = null
+
+    @JvmStatic
+    val authorization: String
+        get() = formatToken(token!!)
```
iamareebjamal: SYIF(TM)

android/app/src/main/java/org/fossasia/openevent/core/auth/AuthUtil.kt
```diff
@@ -0,0 +1,132 @@
...
+        context.startActivity(intent)
+        ActivityCompat.finishAffinity(context as Activity)
+    }
+
+    private fun showProgressBar(progressBar: ProgressBar, show: Boolean) {
+        if (show) {
+            progressBar.visibility = View.VISIBLE
+        } else {
+            progressBar.visibility = View.GONE
+        }
```
iamareebjamal: Use ternary expression inside

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/auth/AuthUtil.kt
```diff
@@ -0,0 +1,128 @@
...
+    const val INVALID = 2
+
+    private var authenticator: Authenticator? = null
+
+    @JvmStatic
+    val authorization: String
+        get() = formatToken(token)
+
+    private val token: String?
+        get() = SharedPreferencesUtil.getString(ConstantStrings.TOKEN, null)
```
iamareebjamal: If a token is nullable, authorization may as well

sriramr98: So you want the token not null ?

iamareebjamal: No, authorization should be nullable as well

iamareebjamal (approved):

PR #2405, opened 2018-03-31
Converted classes to Kotlin

sriramr98:
First steps for issues #2357  

Changes: Converted classes to Kotlin.


android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
```diff
@@ -85,15 +85,15 @@ object Utils {
             if (isBaseUrlEmpty) {
                 swipeRefreshLayout.isEnabled = false
             } else {
-                StrategyRegistry.instance.eventBusStrategy!!.eventBus.register(`object`)
+                StrategyRegistry.instance.eventBusStrategy?.getEventBus()?.register(`object`)
```
iamareebjamal: Why not ?.eventBus

sriramr98:
@iamareebjamal is JCenter down again?

android/app/src/main/java/org/fossasia/openevent/data/repository/RealmDataRepository.kt
```diff
@@ -0,0 +1,673 @@
...
+            val realm1 = Realm.getDefaultInstance()
+
+            realm1.beginTransaction()
+            realm1.where(Session::class.java)
+                    .equalTo("id", sessionId)
+                    .findFirst()?.isBookmarked = bookmark
+
+            StrategyRegistry.instance
+                    .eventBusStrategy!!
+                    .postEventOnUIThread(BookmarkChangedEvent())
```
iamareebjamal: SYIF

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
sriramr98: Because I could not set a default getter implementation because the getter referenced the eventBus and that turned into infinite recursion. So I had to create an explicit getEventBus method

iamareebjamal: Why would it recur? That doesn't make sense

sriramr98: For some reason, accessing the property from inside the getter calls the getter again.

iamareebjamal: Why are you accessing it inside getter, of course it will recur that way 

It's like saying

```Java
int getAge() {
  return getAge();
}
```

sriramr98: Yeah, android studio showed the error. But I found a solution using the 'field' naming.

iamareebjamal: Why are you using that?

sriramr98: Becuase we need to do this
if (field == null) {
                field = Bus()
            }
            return field

iamareebjamal: use lazy delegator

sriramr98: We don't need the lazy approach here. We can just use the field attribute.

iamareebjamal (approved):

PR #2407, opened 2018-04-22
added mobile data fix for download

### LINKED ISSUE #2364: app downloads data on cellular even though download only over wifi is turned on
downloading new feeds is the same even when download only over wifi is enabled 
it is giving a prompt that charges may apply in both the cases(download only over wifi is turned on or off)
Would you like to work on the issue?
yes

sriramr98:
Fixes #2364
Changes: Fixed the bug where app downloads data in mobile data even when chosen to download only over wifi.


sriramr98:
@iamareebjamal please review the PR.

iamareebjamal (approved):

PR #2409, opened 2018-04-29
fix: convert classes to kotlin

### LINKED ISSUE #2408: Convert some classes to kotlin
**Actual Behaviour**

There are some non android specific classes in java 

**Expected Behaviour**

Convert non android specific classes to kotlin

Yes

Parent issue #2357

nikit19:
Fixes #2408 

Changes: converted non android specific classes to kotlin


iamareebjamal:
These classes are to be removed by the end of this GSOC. We are moving away from event based arch. So no point of touching these

PR #2412, opened 2018-04-30
fix: app does not crash on selecting some featured speakers

### LINKED ISSUE #2411: App crashes on selecting some featured speakers
**Actual Behaviour**

App crashes on selecting some featured speakers

**Expected Behaviour**

App should not crash

**Steps to reproduce it**

Open the app. Select featured speakers

**LogCat for the issue**

 `java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter string`

Yes

nikit19:
Fixes #2411 

Changes: app does not crash on selecting some featured speakers

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
+++ b/android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
@@ -19,9 +19,6 @@ import org.fossasia.openevent.common.date.DateConverter
 import org.fossasia.openevent.config.StrategyRegistry
 import org.fossasia.openevent.data.Event
 
-import java.util.regex.Matcher
-import java.util.regex.Pattern
-
 object Utils {
 
         @JvmStatic
@@ -48,12 +45,10 @@ object Utils {
         }
 
         @JvmStatic
-        fun checkStringEmpty(string: String): String {
-            var finalString = ""
-            if (!isEmpty(string)) {
-                finalString = string
-            }
-            return finalString
+        fun nullToEmpty(string: String?): String? {
+            if (isEmpty(string))
+                return ""
+            return string
         }
 
         @JvmStatic
--- a/android/app/src/main/java/org/fossasia/openevent/core/discount/DiscountViewHolder.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/discount/DiscountViewHolder.java
@@ -49,10 +49,10 @@ public class DiscountViewHolder extends RecyclerView.ViewHolder {
     }
 
     public void bindDiscountCode(DiscountCode discountCode) {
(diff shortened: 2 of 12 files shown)


android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
```diff
@@ -48,10 +48,10 @@ object Utils {
         }
 
         @JvmStatic
-        fun checkStringEmpty(string: String): String {
+        fun checkStringEmpty(string: String?): String {
             var finalString = ""
             if (!isEmpty(string)) {
-                finalString = string
+                finalString = string!!
```
iamareebjamal: It's better to just replace this line with `throw new NullPointerException()`

nikit19: Okay, I will update the PR

android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
```diff
@@ -48,10 +45,10 @@ object Utils {
         }
 
         @JvmStatic
-        fun checkStringEmpty(string: String): String {
+        fun checkStringEmpty(string: String?): String {
             var finalString = ""
             if (!isEmpty(string)) {
-                finalString = string
+                finalString = string ?: throw  NullPointerException()
```
iamareebjamal: This was sarcasm. You obviously don't want to do this. !! does exactly this

nikit19: You could have used an emoji to make it look more like a sarcastic comment :laughing: 
sometimes it is difficult to figure out
but I guess I got your point now  :P

android/app/src/main/java/org/fossasia/openevent/common/utils/Utils.kt
```diff
@@ -48,8 +45,8 @@ object Utils {
         }
 
         @JvmStatic
-        fun checkStringEmpty(string: String): String {
-            var finalString = ""
+        fun checkStringEmpty(string: String?): String? {
+            var finalString: String? = ""
```
iamareebjamal: Make this function like this

```
if (isEmpty(string))
    return ""
return string
```

iamareebjamal: Also change the function name to `nullToEmpty`

nikit19:
Done with the requested changes. Please review

iamareebjamal (approved):

PR #2414, opened 2018-05-01
fix: type label error fixed

### LINKED ISSUE #2413: Type label is visible while the value returned is null
**Actual Behaviour**

The "Type" label is visible when the value returned is null

**Expected Behaviour**

It shouldn't be visible

**Screenshots of the issue**

mtp://[usb:001,008]/Internal%20storage/Pictures/Screenshots/2018_05_01_23.37.39.png

Yes

dreadpool2:
Fixes #2413

Changes: Type label is now not visible when value returned is null.

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
@@ -225,6 +225,7 @@ public class SessionDetailActivity extends BaseActivity implements AppBarLayout.
         SessionType type = session.getSessionType();
 
         if (type != null) {
+            typeLabel.setVisibility(View.VISIBLE);
             sessionType.setText(type.getName());
             typelayout.setVisibility(View.VISIBLE);
         }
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
```diff
@@ -225,6 +225,7 @@ private void updateSession() {
         SessionType type = session.getSessionType();
 
         if (type != null) {
+            typeLabel.setVisibility(View.VISIBLE);
             sessionType.setText(type.getName());
             typelayout.setVisibility(View.VISIBLE);
```
iamareebjamal, line 230: When type layout is shown and hidden, why is there a need to show and hide textview, please make it only one

iamareebjamal (changes requested):

android/app/src/main/java/org/fossasia/openevent/core/track/session/SessionDetailActivity.java
dreadpool2: @iamareebjamal Actually the `typelabel` is not a part of the `typelayout`, thats why I had to handle its visibility separately

iamareebjamal (approved):

PR #2447, opened 2018-05-12
fix : Unnecessary Space error fixed

### LINKED ISSUE #2436: Unnecessary extra space in Track item
**Actual Behaviour**

There is an unnecessary extra space in track item whenever the Type label is not present

**Expected Behaviour**

This should not happen

Yes

dreadpool2:
Fixes #2436

### DIFF
--- a/android/app/src/main/res/layout/tracksactvity_item.xml
+++ b/android/app/src/main/res/layout/tracksactvity_item.xml
@@ -75,6 +75,7 @@
                     style="@style/Base.TextAppearance.AppCompat.Subhead"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
+                    android:visibility="gone"
                     android:textColor="@color/primary_text"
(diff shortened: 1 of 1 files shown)


android/app/src/main/res/layout/tracksactvity_item.xml
```diff
@@ -75,6 +75,7 @@
                     style="@style/Base.TextAppearance.AppCompat.Subhead"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
+                    android:visibility="gone"
```
iamareebjamal, line 78: So when are we showing this?

dreadpool2: @iamareebjamal It was already managed at the other end. The visibility check is already present in the SessionViewHolder file

iamareebjamal (approved):

PR #2467, opened 2018-05-16
fix : Share 'Call for speakers' message directly from app

### LINKED ISSUE #2425: Direct Explicit Intent of Mail not present while calling for speakers
**Actual Behaviour**

We can only copy the email id while calling for speakers

**Expected Behaviour**

The Call for speakers should directly lead us to our preferred mail system

Yes

dreadpool2:
Fixes #2425

### DIFF
--- a/android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragment.java
+++ b/android/app/src/main/java/org/fossasia/openevent/core/about/AboutFragment.java
@@ -358,22 +358,25 @@ public class AboutFragment extends BaseFragment implements OnBookmarkSelectedLis
         TextView toDateOfEvent = dialogView.findViewById(R.id.to_date_textview);
 
         if (event.isValid() && event.getSpeakersCall().isValid()) {
-            SpeakersCall speakersCall = event.getSpeakersCall();
+            final SpeakersCall speakersCall = event.getSpeakersCall();
             holder.setText(event.getEventCopyright().getHolder());
             String announcementString = Html.fromHtml(speakersCall.getAnnouncement()).toString();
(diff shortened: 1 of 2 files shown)


android/app/src/main/assets/event
```diff
@@ -78,7 +78,8 @@
         "starts-at": "2016-11-01T16:00:00+08:00",
         "privacy": "public",
         "ends-at": "2017-01-20T17:00:00+08:00",
-        "hash": ""
+        "hash": "",
+        "email-subject": "Be a Speaker at the FOSSASIA Summit"
```
nikit19: Please don't change this file. This is a copy of the JSON response so that the app can work offline. These changes would only work when the app uses this JSON file and not when it requests it from the server.

nikit19 (changes requested):
Please see my comment

android/app/src/main/assets/event
dreadpool2: @nikit19 I will revert this :+1:

dreadpool2:
@iamareebjamal Please review

iamareebjamal (changes requested):
This will only work for FOSSASIA, it should work for any app

dreadpool2:
@iamareebjamal Pardon me, but I didn't get you

iamareebjamal:
This app is a template app, app generator generates eventyay apps for any event, not just FOSSASIA

dreadpool2:
@iamareebjamal Should I just change the subject then ?

iamareebjamal:
Change it in such way that it works generally

dreadpool2:
@iamareebjamal PR updated

iamareebjamal (approved):

nikit19 (approved):

PR #2485, opened 2019-01-06
fix: Updated UIs for Fragments with no items

### LINKED ISSUE #2466: Improve UI of various fragments when there are no items
**Actual Behaviour**

Currently only a text message is shown when there are no discount codes,FAQs,speakers etc in their respective fragments

**Expected Behaviour**

If an image is shown along with the text it would be more appealing to the user.
Similar to what has been done with the live feed

**Steps to reproduce it**
Open fragments with no items

**Screenshots of the issue**
screenshot_20180516-174504

Yes

raystatic:
Fixes #2466

Changes: Images added in the fragments when no items are there in it

Screenshots for the change: 

whatsapp image 2019-01-07 at 00 57 20
whatsapp image 2019-01-07 at 00 57 20 1
whatsapp image 2019-01-07 at 00 57 20 2


iamareebjamal:
This app is deprecated.

